PR #16, opened 2018-09-18
macOS fixes

jrprice:
With these changes clvk builds on macOS (though I also have hack the OpenCL headers slightly).

### DIFF
--- a/src/device.hpp
+++ b/src/device.hpp
@@ -66,8 +66,8 @@ typedef struct _cl_device_id {
         return std::min(max_alloc_size() * 4, actual_memory_size());
     }
 
-    size_t max_alloc_size() const {
-        size_t max_buffer_size = m_properties.limits.maxStorageBufferRange;
+    uint64_t max_alloc_size() const {
+        uint64_t max_buffer_size = m_properties.limits.maxStorageBufferRange;
         return std::min(max_buffer_size, actual_memory_size());
     }
 
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -19,6 +19,10 @@
 #include <utility>
 #include <vector>
 
+#ifdef __APPLE__
+#include <unistd.h>
+#endif
+
 #include <vulkan/vulkan.h>
 
 #include "spirv/1.0/spirv.hpp"


src/device.hpp
```diff
@@ -63,12 +63,12 @@ typedef struct _cl_device_id {
     }
 
     uint64_t memory_size() const {
-        return std::min(max_alloc_size() * 4, actual_memory_size());
+        return std::min<uint64_t>(max_alloc_size() * 4, actual_memory_size());
```
kpet: I think it's better to change the return type of `max_alloc_size()` to `uint64_t`. The API reports a `cl_ulong` anyway which, contrary to `size_t` is guaranteed to be 64-bit.

src/device.hpp
```diff
@@ -63,12 +63,12 @@ typedef struct _cl_device_id {
...
     }
 
     size_t max_alloc_size() const {
         size_t max_buffer_size = m_properties.limits.maxStorageBufferRange;
-        return std::min(max_buffer_size, actual_memory_size());
+        return std::min<uint64_t>(max_buffer_size, actual_memory_size());
```
kpet: You can get rid of this one by changing the type of `max_buffer_size` to `uint64_t`; `size_t` was a kind of arbitrary and not very helpful choice on my part.

src/program.cpp
```diff
@@ -16,6 +16,7 @@
 #include <cstring>
 #include <iostream>
 #include <map>
+#include <unistd.h>
```
kpet: Could you wrap that in a `#ifdef mac` (not sure what the best macro to use is)?

jrprice:
Updated to address comments, thanks for the feedback!

kpet:
>(though I also have hack the OpenCL headers slightly).

Anything worth upstreaming there?

jrprice:
> Anything worth upstreaming there?

Potentially, I'm considering opening a PR for OpenCL-Headers so we can discuss the changes (which might be considered controversial. 

> BTW, is the simple_test working for you?

Yes, although I haven't tried with MoltenVK yet (I have my own Vulkan implementation/emulator).

kpet:
>Yes, although I haven't tried with MoltenVK yet

That's great to hear :).

> (I have my own Vulkan implementation/emulator).

Is that public? If not, do you intend to release it?

jrprice:
> > (I have my own Vulkan implementation/emulator).
> 
> Is that public? If not, do you intend to release it?

Yes, it's called Talvos, and is available here:
https://talvos.github.io/
https://github.com/talvos/talvos

It's not a "real" Vulkan implementation, it's intended to be a framework for creating dynamic analysis tools for Vulkan/SPIR-V (in a Valgrind-like manner). As such, it runs very slowly, so watch out for that if you try it.

There's one or two more changes that need to be made (to both clvk and Talvos) to get them working together, so will try to get them upstreamed soon.

kpet:
Interesting. This sounds really nice.

>There's one or two more changes that need to be made (to both clvk and Talvos) to get them working together, so will try to get them upstreamed soon.

Great! Looking forward to the patches.

PR #18, opened 2018-09-18
Only enable VK_EXT_debug_report if it is supported

jrprice:
This checks if the instance actually supports the `VK_EXT_debug_report` extension, instead of unconditionally enabling it. A message is now logged if it was not enabled.

With this change, clvk runs on Talvos with no modifications to either (on Linux). For `simple_test` to work however, the device type needs to be changed from `GPU` to `ALL`.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -26,6 +26,7 @@ VkInstance gVkInstance;
 cvk_platform *gPlatform;
 loglevel gLoggingLevel = loglevel::fatal;
 bool gLoggingColour = true;
+bool gDebugReportEnabled = false;
 
 std::string gCLSPVPath = DEFAULT_CLSPV_BINARY_PATH;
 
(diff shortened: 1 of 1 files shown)


src/init.cpp
```diff
@@ -140,16 +144,24 @@ static void init_vulkan()
...
-    res = fnvkCreateDebugReportCallbackEXT(gVkInstance, &callbackInfo, nullptr, &gVkDebugCallback);
-    CVK_VK_CHECK_FATAL(res, "Can't setup debug callback");
+    if (gDebugReportEnabled) {
+        CVK_VK_GET_INSTANCE_PROC(vkCreateDebugReportCallbackEXT);
+
+        res = fnvkCreateDebugReportCallbackEXT(gVkInstance, &callbackInfo, nullptr, &gVkDebugCallback);
+        CVK_VK_CHECK_FATAL(res, "Can't setup debug callback");
+    }
+    else {
+        cvk_info("VK_EXT_debug_report not enabled");
```
kpet: I'm half tempted to suggest this should be a warning.

jrprice: Changed to warning.

kpet: Thanks (wasn't really requesting it TBH, merely thinking out loud)

PR #34, opened 2019-03-14
Use clspv api through Cmake option

alan-baker:
Adds Cmake option to select compile style:
 * default remains separate process
 * new option uses Clspv API

Changes:
* Changes to support using Clspv API
* update UBO test to meet std140 rules

I've tested the simple tests using both compiler options.

### DIFF
--- a/tests/simple_ubo/simple.cpp
+++ b/tests/simple_ubo/simple.cpp
@@ -28,10 +28,10 @@
     } while (0)
 
 const char *program_source = R"(
-kernel void test_simple(global int* out, constant int* c_data)
+kernel void test_simple(global int4* out, constant int4* c_data)
 {
     size_t gid = get_global_id(0);
-    out[gid] = gid + c_data[gid];
+    out[gid] = (int4)(gid, gid, gid, gid) + c_data[gid];
 }
 )";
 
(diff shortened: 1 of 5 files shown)


tests/simple_ubo/simple.cpp
```diff
@@ -129,7 +129,8 @@ int main(int argc, char* argv[])
     bool success = true;
     auto buffer_data = static_cast<cl_int*>(ptr);
     for (cl_uint i = 0; i < BUFFER_SIZE/sizeof(cl_int); ++i) {
-        if (buffer_data[i] != static_cast<cl_int>(i + 1)) {
+        int i_div_4 = i / 4;
+        if (buffer_data[i] != static_cast<cl_int>(i_div_4 + 1)) {
             printf("Failed comparison at buffer_data[%d]: expected %d but got %d\n",
                    i, i + 1, buffer_data[i]);
```
kpet: Didn't spot you forgot to update the message, I'll fix that as part of folding this test into the framework.

tests/simple_ubo/simple.cpp
```diff
@@ -71,7 +71,7 @@ int main(int argc, char* argv[])
     CHECK_CL_ERRCODE(err);
 
     // Build program
-    err = clBuildProgram(program, 1, &device, " -constant-args-ubo -inline-entry-points ", nullptr, nullptr);
+    err = clBuildProgram(program, 1, &device, " -constant-args-ubo -inline-entry-points -relaxed-ubo-layout ", nullptr, nullptr);
```
kpet: `-relaxed-ubo-layout` isn't really needed, is it?

alan-baker: No that was in there from an earlier debugging effort. I forgot to remove it.

kpet: Thanks for confirming.

PR #118, opened 2019-11-19
Add missing <array> include

jrprice:
PR #117 introduced uses of `std::array` without adding this include.


kpet:
Thanks! What build configuration did you catch that with?

BTW, just so you know, I've got most of the runtime image support done (see assigned issues) and I'm working on pushing this.

PR #124, opened 2019-11-25
Build fixes

jrprice:
One of these would have been caught earlier if there was a Clang configuration in Travis - @kpet is this something we can add or were you hoping to get Clang coverage via a macOS configuration?

### DIFF
--- a/tests/api/testcl.hpp
+++ b/tests/api/testcl.hpp
@@ -107,7 +107,7 @@ struct holder {
             deleter();
         }
     }
-    void deleter() = delete;
+    void deleter() { assert(false); }
     operator T() {
         return m_obj;
     }
--- a/src/memory.hpp
+++ b/src/memory.hpp
@@ -14,6 +14,8 @@
 
 #pragma once
 
+#include <array>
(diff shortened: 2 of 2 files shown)


tests/api/testcl.hpp
```diff
@@ -107,7 +107,7 @@ struct holder {
             deleter();
         }
     }
-    void deleter() = delete;
```
kpet: What was the issue with `delete`?

jrprice: Clang does not seem to like calling a deleted method in a template class even when there is a specialized (non-deleted) definition of that method for all instantiations of the template. 

There has been similar issues in the past that have now been fixed, e.g.:
https://bugs.llvm.org/show_bug.cgi?id=17537
But I guess they haven't hit the case where the deleted method is a member of templated class before.

kpet: Right, I had a feeling this was a case of broken toolchain. Thanks for the details (and sorry I didn't spot the commit message that described the issue before I asked).

PR #126, opened 2019-11-26
Fix build error for CMAKE_BUILD_TYPE=Debug

jrprice:
I'd recommend making all CI configs use a Debug build, both to catch this sort of thing and also so that the tests run with all assertions enabled.


kpet:
I'm not sure about switching all of them as release builds do catch other issues. I'll change the online compiler configuration to use a debug build for now. We can revise this if we find it's not enough.

PR #127, opened 2019-11-27
Switch the online compiler Travis configuration to use a Debug build

kpet:


kpet:
Now covered in Azure, closing.

PR #144, opened 2020-01-10
Remove deleted move special member functions

jrprice:
Fixes a build error occurring with vectors of `refcounted_holder` objects. The `push_back`/`emplace_back` methods require the element type to be [MoveInsertable](https://en.cppreference.com/w/cpp/named_req/MoveInsertable), and will try to use the move constructor if it is declared. This produces an error if the move constructor is deleted, but if it is left undeclared the compiler can use the copy constructor instead.

The build error was introduced with the recent change to support literal samplers, which added a vector of `cvk_sampler_holder` objects.

Full disclosure: my knowledge of this corner of C++ is hairy at best, and I couldn't work out the motivation for marking these as deleted in the first place.

### DIFF
--- a/src/objects.hpp
+++ b/src/objects.hpp
@@ -76,8 +76,6 @@ template <typename T> struct refcounted_holder {
         }
     }
 
-    refcounted_holder(const refcounted_holder&& other) = delete;
-
     ~refcounted_holder() {
         if (m_refcounted != nullptr) {
             m_refcounted->release();
@@ -88,7 +86,6 @@ template <typename T> struct refcounted_holder {
 
     operator T*() const { return m_refcounted; }
 
-    refcounted_holder& operator=(const refcounted_holder&) = delete;
     refcounted_holder& operator=(const refcounted_holder&&) = delete;
 
     void reset(T* refc) {


kpet:
I'm no expert either but I think this is a bug in your compiler. I couldn't find any source stating that deleting the move constructor prevents the copy constructor from being used to satisfy MoveInsertable.

What we should be doing anyway is to add a move constructor to avoid refcounting temporary objects up/down. I've done that in [this commit](https://github.com/kpet/clvk/commit/3c664e4adce1183028e35deacccb03bbad7d1b29) and I'm hopeful this will fix your issue as well.

jrprice:
> I couldn't find any source stating that deleting the move constructor prevents the copy constructor from being used to satisfy MoveInsertable.

Deleting the move constructor does not prevent the copy constructor from satisfying MoveInsertable. My understanding is that the move constructor still participates in overload resolution (despite being deleted) and takes precedence over the copy constructor in this case, thus producing an error when the compiler tries to use the deleted function.

In any case, your change fixes my issue, so this PR can be closed.

kpet:
Hmm, I couldn't find a clear description of this behaviour. Do you have a pointer?

Recent compilers (GCC9 and Clang9) both seem to not fail on that. Can you say what was the failing compiler?

Glad to hear the issue is solved!

jrprice:
>  I couldn't find a clear description of this behaviour. Do you have a pointer?

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2007/n2346.htm#delete

The first sentence is the relevant one:
"The definition form =delete; indicates that the function may not be used. However, all lookup and overload resolution occurs before the deleted definition is noted. That is, it is the definition that is deleted, not the symbol; overloads that resolve to that definition are ill-formed."

That said, there appears to be an exception to this for move constructors:
https://en.cppreference.com/w/cpp/language/move_constructor
"The deleted implicitly-declared move constructor is ignored by overload resolution (otherwise it would prevent copy-initialization from rvalue)."

So could still be a compiler bug.


> Can you say what was the failing compiler?

I can only reproduce this error using Clang trunk and libc++:
https://godbolt.org/z/X7jxr8
Since it doesn't happen with libstdc++, I assume it's something new in libc++ in trunk.

kpet:
Thanks for the links.

>"The deleted implicitly-declared move constructor is ignored by overload resolution (otherwise it would prevent copy-initialization from rvalue)."

I think that's a smoking gun here. You may want to report this.

PR #146, opened 2020-01-14
Add support for literal samplers with online Clspv

jrprice:


src/program.cpp
```diff
@@ -415,6 +415,54 @@ bool spir_binary::load_descriptor_map(
...
+                break;
+            case clspv::version0::CLK_ADDRESS_MIRRORED_REPEAT:
+                desc.addressing_mode = CL_ADDRESS_MIRRORED_REPEAT;
+                break;
+            case clspv::version0::CLK_ADDRESS_REPEAT:
+                desc.addressing_mode = CL_ADDRESS_REPEAT;
+                break;
+            default:
+                cvk_error("Invalid sampler addressing mode: %d",
+                          entry.sampler_data.mask & 0xe);
```
kpet: You could have introduced a local variable to avoid repeating the mask and magic number.

src/program.cpp
```diff
@@ -415,6 +415,54 @@ bool spir_binary::load_descriptor_map(
...
         }
+
+        if (entry.kind == clspv::version0::DescriptorMapEntry::Kind::Sampler) {
+            sampler_desc desc;
+            desc.descriptorSet = entry.descriptor_set;
+            desc.binding = entry.binding;
+
+            desc.normalized_coords = entry.sampler_data.mask & 0x1;
+
+            switch (entry.sampler_data.mask & 0xe) {
```
kpet: It would be nice to introduce symbolic names for the masks, probably in clspv's `DescriptorMap.h`. They are used in clspv as well.

src/program.cpp
```diff
@@ -415,6 +415,54 @@ bool spir_binary::load_descriptor_map(
...
+            case clspv::version0::CLK_ADDRESS_MIRRORED_REPEAT:
+                desc.addressing_mode = CL_ADDRESS_MIRRORED_REPEAT;
+                break;
+            case clspv::version0::CLK_ADDRESS_REPEAT:
+                desc.addressing_mode = CL_ADDRESS_REPEAT;
+                break;
+            default:
+                cvk_error("Invalid sampler addressing mode: %d",
+                          entry.sampler_data.mask & 0xe);
+                break;
```
kpet: `return false`? Loading the descriptor map should fail in that case.

src/program.cpp
```diff
@@ -415,6 +415,54 @@ bool spir_binary::load_descriptor_map(
...
+            case clspv::version0::CLK_FILTER_NEAREST:
+                desc.addressing_mode = CL_FILTER_NEAREST;
+                break;
+            case clspv::version0::CLK_FILTER_LINEAR:
+                desc.addressing_mode = CL_FILTER_LINEAR;
+                break;
+            default:
+                cvk_error("Invalid sampler filter mode: %d",
+                          entry.sampler_data.mask & 0x30);
+                break;
```
kpet: Same comment here. Loading the descriptor map should fail in that case.

jrprice:
Addressed comments, PTAL.

jrprice:
>  you're going to have to retrigger the CI jobs a couple of times for the build to complete

Done; bots are green now.

kpet:
Thanks!

PR #152, opened 2020-01-24
Add ICD support

jrprice:
All OpenCL API objects now inherit from an `icd_object` base class which contains the dispatch table. Downcasts have been introduced as necessary to convert to the clvk subclasses.

Linking to clvk directly is still supported. The dispatch table function pointers are obtained by calling `dlsym()` on the clvk library, which circumvents having to rename the API definitions to avoid clashing with the definitions provided by the ICD loader.

This is perhaps a slightly strange approach, so let me know if you'd rather I just rename the APIs and set the function pointers directly (e.g. `m_dispatch->clFoo() = clvk::clFoo()`, which then requires an another definition of `clFoo()` which calls `clvk::clFoo()` to make direct linking work). 

The many `static_cast` calls are a little ugly; I couldn't see an easy way to make them nicer. We could shorten them a bit by adding a conversion method to `icd_object`, which would lead to (e.g.) `context.as<cvk_context*>()`, but this only saves a few characters. Let me know what you think.

Tested this with OpenCL CTS -  no regressions.

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -256,13 +256,12 @@ cl_int clGetDeviceIDs(cl_platform_id platform, cl_device_type device_type,
     return CL_SUCCESS;
 }
 
-cl_int clGetDeviceInfo(cl_device_id device, cl_device_info param_name,
+cl_int clGetDeviceInfo(cl_device_id dev, cl_device_info param_name,
                        size_t param_value_size, void* param_value,
                        size_t* param_value_size_ret) {
     LOG_API_CALL(
         "device = %p, param_name = %d, size = %zu, value = %p, size_ret = %p",
-        device, param_name, param_value_size, param_value,
-        param_value_size_ret);
+        dev, param_name, param_value_size, param_value, param_value_size_ret);
     cl_int ret = CL_SUCCESS;
 
     size_t size_ret = 0;
@@ -284,11 +283,11 @@ cl_int clGetDeviceInfo(cl_device_id device, cl_device_info param_name,
     cl_platform_id val_platform;
     cl_device_id val_deviceid;
 
+    auto device = icd_downcast(dev);
+
     if (!is_valid_device(device)) {
         return CL_INVALID_DEVICE;
     }
-
-    auto dev = icd_downcast(device);
 
     switch (param_name) {
     case CL_DEVICE_PLATFORM:
@@ -297,12 +296,12 @@ cl_int clGetDeviceInfo(cl_device_id device, cl_device_info param_name,
         size_ret = sizeof(val_platform);
         break;
     case CL_DEVICE_TYPE:
-        val_devicetype = dev->type();
+        val_devicetype = device->type();
         copy_ptr = &val_devicetype;
         size_ret = sizeof(val_devicetype);
         break;
     case CL_DEVICE_NAME:
-        val_string = dev->name();
+        val_string = device->name();
         copy_ptr = val_string.c_str();
         size_ret = val_string.size_with_null();
         break;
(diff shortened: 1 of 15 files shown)


src/kernel.hpp
```diff
@@ -192,10 +193,13 @@ struct cvk_kernel_argument_values {
             if (size != sizeof(void*)) {
                 return CL_INVALID_ARG_SIZE;
             }
-
-            auto refc = *reinterpret_cast<refcounted* const*>(value);
-
-            m_kernel_resources[arg.binding].reset(refc);
+            if (arg.kind == kernel_argument_kind::sampler) {
+                m_kernel_resources[arg.binding].reset(static_cast<cvk_sampler*>(
+                    *reinterpret_cast<const cl_sampler*>(value)));
```
kpet: I suggest doing this in two/three steps (value to api sampler, api sampler to sampler, use) for readability.

jrprice:
Thanks for the review. I've pushed updates which I believe address your comments.

- `-Bsymbolic` - I thought there would probably be a flag to do this, just couldn't find it! I've switched to this approach with a statically defined dispatch table, using a default initializer in the `icd_object` class so that no other code has to worry about doing this. I've also tested on macOS where it seems to work without any additional linker flags.

- I've added the `icd_downcast` routine you suggested. Note that there are still several cases where a `static_cast` is required (e.g. for downcasting `cl_mem` to buffer/image classes depending on the use).

- I've added a commit that renames function parameters. It didn't make the patch /that/ much smaller, partly due to the fact that renaming the function parameters changes the way the code is formatted. The original reason I avoided renaming the function parameters was to keep them consistent with the names used in the spec/headers. I'll let you decide whether to keep this commit or not :-)

- Added CI coverage for the ICD interface (on Linux only). I /think/ it's correct (tested on a Debian-derived workstation), but it's hard to tell if it's actually running with the ICD interface in CI.

src/api.cpp
```diff
@@ -89,26 +91,7 @@ struct api_query_string : public std::string {
...
-    if (platforms != nullptr) {
-        platforms[0] = gPlatform;
-    }
-
-    if (num_platforms != nullptr) {
-        *num_platforms = 1;
-    }
-
-    return CL_SUCCESS;
+    return clIcdGetPlatformIDsKHR(num_entries, platforms, num_platforms);
```
kpet: This will result in incorrect logging. I suggest introducing a `cvk_get_platform_ids` function called used by both `clGetPlatformIDs` and `clIcdGetPlatformIDsKHR`. The logging will have to be duplicated.

src/device.hpp
```diff
@@ -210,9 +209,12 @@ typedef struct _cl_device_id {
...
     std::string m_extensions;
+};
 
-} cvk_device;
+static inline cvk_device* icd_downcast(cl_device_id device) {
+    return static_cast<cvk_device*>(device);
+}
 
-typedef struct _cl_platform_id {
+typedef struct _cl_platform_id : public clvk::icd_object {
```
kpet: Any specific reason `cvk_platform` is declared differently from all other objects?

src/memory.hpp
```diff
@@ -298,7 +306,7 @@ struct cvk_image : public cvk_mem {
     size_t height() const { return m_desc.image_height; }
     size_t depth() const { return m_desc.image_depth; }
     size_t array_size() const { return m_desc.image_array_size; }
-    cvk_mem* buffer() const { return m_desc.buffer; }
+    cvk_mem* buffer() const { return static_cast<cvk_mem*>(m_desc.buffer); }
```
kpet: Shouldn't this one be an `icd_downcast`?

tests/travis/step-script.sh
```diff
@@ -46,3 +46,10 @@ fi
 ./simple_test
 ./simple_test_static
 ./api_tests
+
+if [ "${TRAVIS_OS_NAME}" == "linux" ]; then
+    echo $PWD/libOpenCL.so >clvk.icd
+    export OCL_ICD_VENDORS=$PWD
```
kpet: We could probably use `OCL_ICD_FILENAMES` to avoid the file. It doesn't matter really.

jrprice: This environment variable is not provided by `ocl-icd`, which I believe is the ICD loader packaged by Debian (and thus what Travis CI is using).

jrprice:
Addressed comments, PTAL.

tests/travis/step-script.sh
kpet: Right. We can test with the Khronos loader later (I did use that in my local testing and it worked fine so I'm not overly worried). Most distributions use `ocl-icd` at the moment anyway.

kpet:
Seems you went back on the CMake in the last change. The macOS bots are failing. I think it would be more practical to avoid force-pushing in the future, I would likely have spotted that in review (you can't use the diff since last review feature when force-pushing).

jrprice:
Fixed now, sorry about that.

I force pushed because I had rebased on top of the latest changes on master, but that wasn't really necessary for this PR. Will avoid doing this in the future.

PR #158, opened 2020-03-05
Change how kernel args are retained and released

### LINKED ISSUE #157: Retaining objects when call clSetKernelArg
clSetKernelArg currently retains arguments by virtue of maintaining a refcounted_holder for each arg. This seems to go against the guidance in the API documentation:
>A kernel object does not update the reference count for objects such as memory, sampler objects specified as argument values by clSetKernelArg. Users may not rely on a kernel object to retain objects specified as argument values to the kernel.

This leads to a situation where the kernel argument must be reset or the kernel must be released in order to full release the memory for buffers and samplers. This recently resulted in a device crash for some tests I was running (due to OOM).

It seems it would be better to retain the memory objects and samplers attached as arguments when the kernel is enqueued and released when that kernel instance completes. To do this, `kernel_argument_values` should only maintain a pointer to the refcounted object instead of a refcounted_holder.

@kpet, any thoughts about the suggested approach?

alan-baker:
Fixes #157

* Changed cvk_kernel_argument_values to hold raw refcounted pointers
* Add new methods to cvk_kernel to retain and release arguments
  * retain args prior to setting up descriptor sets and enqueueing
  * release upon completion of the kernel

This is the simplest change to address #157 and I'd like some feedback on it if possible. I think I've made the changes suggested in the issue. I also have the suggested test with the destructor callback, but am not sure it belongs in this repo. 

I have tested this against the app that previously ran OOM due to not free'ing buffers and it passes now.

### DIFF
--- a/src/kernel.cpp
+++ b/src/kernel.cpp
@@ -322,6 +322,7 @@ bool cvk_kernel::setup_descriptor_sets(
 
     // Transfer ownership of the argument values to the command
     arg_values = std::move(m_argument_values);
+    arg_values->retain_resources();
 
     // Create a new set, copy the argument values
     m_argument_values = cvk_kernel_argument_values::create(*arg_values.get());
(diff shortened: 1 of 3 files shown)


alan-baker:
Good point about calling release on the potentially incorrect resources. Since the resources get transferred to the cvk_command_kernel, that destructor could call release_resources() if you want to avoid the bool management. The retain_resources() (but move to cvk_kernel_argument_values) seems like it could stay how it is since they are both referring to the same resources at that point. That might make it slightly simpler.

kpet:
Unfortunately, that won't work, as you wouldn't be able to safely destroy a `cvk_command_kernel` until the `cvk_kernel_argument_values` was assigned and `retain_resources` was called.

>The retain_resources() (but move to cvk_kernel_argument_values) seems like it could stay how it is since they are both referring to the same resources at that point.**

Yes but since argument setting is delegated to `cvk_kernel_argument_values` by `cvk_kernel`, I think it's cleaner to do it in `cvk_kernel_argument_values`.

Issue #172, opened 2020-03-31
Respect VkPhysicalDeviceLimits::maxMemoryAllocationCount


jrprice:
When using UBOs/SSBOs for POD arguments, each kernel enqueue performs a new memory allocation, which is freed when the kernel instance completes. We don't currently track how many kernel instances are currently in flight, and so it is possible to exceed `VkPhysicalDeviceLimits::maxMemoryAllocationCount` allocations (the min max value for this is 4096). When this happens kernel enqueue and kernel creation commands start failing with `CL_OUT_OF_RESOURCES`.

To deal with this (without just failing), clvk would need to track the maximum number of allocations coming from buffers/images, kernels, and kernel instances, and stop enqueueing new kernels when this limit is hit. To avoid blocking the host thread at the point of enqueue I suspect we need to move some kernel enqueue logic (e.g. the part that causes POD buffer allocation) to the background submission thread. This may also intersect with #167.

It'd be nice to also deal with the `cvk_kernel::MAX_INSTANCES` limit (which I think we currently just ignore when enqueuing kernels?) with the same solution, although this is a per-kernel limit whereas the above is a global limit.

dneto0:
Alternately, use a single big buffer and dole out sub-allocations for individual launches.
This transforms it into an internal memory allocation/deallocation problem.

kpet:
I'm guessing you're hitting this in a practical application, aren't you? The move to push constants is likely to make this go away in many practical cases.

In theory, failing with `CL_OUT_OF_RESOURCES` is perfectly acceptable behaviour but I doubt many applications can make sense of this or have logic for retying.

On POD buffers, sub-allocating feels much more attractive than deferring the whole command building. Small buffers and images could also be allocated from a pool to make it less likely to hit the limit.

How urgently do you need a solution?

jrprice:
I'm hitting this with the MACE benchmarks.

I'm happy to wait for push constants, but just wanted to flag this as a potential issue when not using them, if we keep that option alive.

Agree that a large buffer with suballocations is a good solution.

PR #174, opened 2020-03-31
Implement a program scope pipeline cache

jrprice:
Introduces a `cvk_entry_point` class which now owns all pipeline state (e.g. descriptor set layouts and information about POD arguments). One instance of this class is shared between all kernel objects created from the same program that target the same kernel function.

Using specialization constants as a key, a `cvk_entry_object` object holds a map of pipelines which can be reused for kernel launches that have the same work-group size and local memory argument sizes. These pipelines are destroyed when the `cvk_entry_point` object is destroyed, which occurs when the parent program object is destroyed.

There isn't a great deal of new code here, it's mostly just refactoring existing code into a new class and changing the lifetime of the Vulkan pipeline structures.

### DIFF
--- a/src/kernel.cpp
+++ b/src/kernel.cpp
@@ -17,120 +17,10 @@
 #include "kernel.hpp"
 #include "memory.hpp"
 
-bool cvk_kernel::build_descriptor_set_layout(
-    VkDevice vkdev, const std::vector<VkDescriptorSetLayoutBinding>& bindings) {
-    VkDescriptorSetLayoutCreateInfo createInfo = {
-        VK_STRUCTURE_TYPE_DESCRIPTOR_SET_LAYOUT_CREATE_INFO, nullptr,
-        0,                                      // flags
-        static_cast<uint32_t>(bindings.size()), // bindingCount
-        bindings.data()                         // pBindings
-    };
-
-    VkResult res;
-    if (bindings.size() > 0) {
-        VkDescriptorSetLayout setLayout;
-        res = vkCreateDescriptorSetLayout(vkdev, &createInfo, 0, &setLayout);
-        if (res != VK_SUCCESS) {
-            cvk_error("Could not create descriptor set layout");
-            return false;
-        }
-        m_descriptor_set_layouts.push_back(setLayout);
-    }
-
-    return true;
-}
-
-bool cvk_kernel::build_descriptor_sets_layout_bindings_for_arguments(
-    VkDevice vkdev, binding_stat_map& smap, uint32_t& num_resources) {
-    bool pod_found = false;
-
-    std::vector<VkDescriptorSetLayoutBinding> layoutBindings;
-    for (auto& arg : m_args) {
-        VkDescriptorType dt = VK_DESCRIPTOR_TYPE_MAX_ENUM;
-
-        switch (arg.kind) {
-        case kernel_argument_kind::buffer:
-            dt = VK_DESCRIPTOR_TYPE_STORAGE_BUFFER;
-            break;
-        case kernel_argument_kind::buffer_ubo:
-            dt = VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER;
-            break;
-        case kernel_argument_kind::ro_image:
-            dt = VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE;
-            break;
-        case kernel_argument_kind::wo_image:
-            dt = VK_DESCRIPTOR_TYPE_STORAGE_IMAGE;
-            break;
-        case kernel_argument_kind::sampler:
-            dt = VK_DESCRIPTOR_TYPE_SAMPLER;
-            break;
-        case kernel_argument_kind::local:
-            continue;
-        case kernel_argument_kind::pod:
-        case kernel_argument_kind::pod_ubo:
-            if (!pod_found) {
-                if (arg.kind == kernel_argument_kind::pod) {
-                    dt = VK_DESCRIPTOR_TYPE_STORAGE_BUFFER;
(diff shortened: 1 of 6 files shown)


src/kernel.cpp
```diff
@@ -17,120 +17,10 @@
...
-
-    return true;
-}
-
 std::unique_ptr<cvk_buffer> cvk_kernel::allocate_pod_buffer() {
     cl_int err;
-    auto buffer =
-        cvk_buffer::create(m_context, 0, m_pod_buffer_size, nullptr, &err);
+    auto buffer = cvk_buffer::create(
+        m_context, 0, m_entry_point->pod_buffer_size(), nullptr, &err);
```
kpet, line 23: I'm wondering whether this shouldn't be delegated to `cvk_entry_point` since it owns the descriptors.

src/program.hpp
```diff
@@ -340,6 +462,7 @@ struct cvk_program : public _cl_program, api_object {
     spir_binary m_binary{SPV_ENV_VULKAN_1_0};
     std::vector<cvk_sampler_holder> m_literal_samplers;
     std::vector<VkPushConstantRange> m_push_constant_ranges;
+    std::unordered_map<std::string, cvk_entry_point*> m_entry_points;
```
kpet, line 465: I'm wondering whether we should use `std::unique_ptr` here. I don't think matters too much (beyond shortening the destructor's code).

src/program.hpp
```diff
@@ -163,6 +164,121 @@ enum class build_operation
...
+    }
+
+    CHECK_RETURN cl_int init();
+
+    CHECK_RETURN VkPipeline
+    create_pipeline(const cvk_spec_constant_map& spec_constants);
+
+    const std::vector<kernel_argument>& args() const { return m_args; }
+
+    VkDescriptorPool descriptor_pool() const { return m_descriptor_pool; }
```
kpet, line 207: The descriptor pool should not be externally accessible (it is externally-synchronised). This makes it much harder to respect the Vulkan requirements.

src/kernel.cpp
```diff
@@ -303,12 +83,13 @@ bool cvk_kernel::setup_descriptor_sets(
     std::lock_guard<std::mutex> lock(m_lock);
 
     // Allocate descriptor sets
+    auto& descriptor_set_layouts = m_entry_point->descriptor_set_layouts();
     VkDescriptorSetAllocateInfo descriptorSetAllocateInfo = {
         VK_STRUCTURE_TYPE_DESCRIPTOR_SET_ALLOCATE_INFO, nullptr,
-        m_descriptor_pool,
+        m_entry_point->descriptor_pool(),
```
kpet, line 89: The allocation code below would violate Vulkan threading requirements on the descriptor pool when multiple kernels using the same entry point allocate descriptors concurrently. Whatever object owns the descriptor pool should be the one responsible for descriptor allocation if we don't want the kernel to have to request the entry point to lock its descriptor pool.

Having the kernel own the descriptor pool enables applications to create several kernels to deal with the `MAX_INSTANCES` limit or reduce contention on the descriptor pool in multithreaded submission scenarios but this would use more memory. I can't think of a killer argument in favour of one solution or the other so I suggest we do whatever is the simplest.

src/kernel.cpp
```diff
@@ -152,42 +42,14 @@ cl_ulong cvk_kernel::local_mem_size() const {
...
-            vkdev, bindingTypes)) {
-        return CL_INVALID_VALUE;
-    }
-
-    uint32_t num_resources;
-    if (!build_descriptor_sets_layout_bindings_for_arguments(
-            vkdev, bindingTypes, num_resources)) {
-        return CL_INVALID_VALUE;
+    cl_int errcode;
+    m_entry_point = m_program->get_entry_point(m_name, &errcode);
```
kpet, line 46: I _think_ this is the first non-CL function that returns the value and sets an error code via a pointer passed as a parameter instead of doing the opposite. I don't think I feel too strongly about that though.

src/kernel.cpp
jrprice: I agree. I was going to explore moving this to `cvk_entry_point` as part of trying to fix issue #172, by having `cvk_entry_point` dish out suballocations of a single large buffer. I can do the simple step of delegating the current approach to `cvk_entry_point` as a first step in the PR though.

src/program.hpp
jrprice: Sure, will do.

src/kernel.cpp
jrprice: Good point.

Given that `cvk_entry_point` owns the descriptor layouts, it makes sense to me to have it own the descriptor pool and do the allocations/frees. This does require it to mutex around these routines as you point out, but not sure whether that's a performance issue in practice. I'd also be reluctant to tie the lifetime of a descriptor pool to a `cl_kernel` object since some workloads (e.g. MACE) seem to create new kernels for every enqueue, so this would be expensive.

I'll push a change to move allocation/free to `cvk_entry_point` (with locks).

src/kernel.cpp
jrprice: `cvk_buffer::create()` does this. I can't actually see any non-CL functions that return an error code and create something as an output parameter (though there are functions that do not create stuff like `cvk_kernel::set_arg`). I also don't feel strongly about this though :-)

kpet: Yeah, that one is almost a CL function :). I'm sure there are but let's not spend time on this.

src/kernel.cpp
kpet: I don't think owning the layout is a good argument in favour: there is only one layout for each entry point but we have the choice for descriptor pools. That being said, I'm not worried about performance either and what you're proposing sounds good.

PR #175, opened 2020-03-31
Fix the clvk-side Windows build errors

fodinabor:
This _fixes_ the build issues from:  #173 

There are two more issues in OpenCL-Headers, that are addressed separately (although the submodule needs updating then).

Also exporting all symbols through CMake is probably not the best idea and should be handled better..

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -92,7 +92,7 @@ else()
     DEFAULT_LLVMSPIRV_BINARY_PATH="$<TARGET_FILE:llvm-spirv>")
 endif()
 
-if(MSVC)
+if(WIN32)
 	target_compile_definitions(OpenCL-objects PRIVATE
 		NOMINMAX)
 endif()
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -16,6 +16,8 @@
 if(NOT MSVC)
     add_compile_options(-Wall -W -Wextra)
     add_compile_options(-Wno-ignored-attributes)
+else()
+	set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
 endif()
 
 if (UNIX AND NOT APPLE)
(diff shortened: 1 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -90,6 +92,11 @@ else()
     DEFAULT_LLVMSPIRV_BINARY_PATH="$<TARGET_FILE:llvm-spirv>")
 endif()
 
+if(MSVC)
+	target_compile_definitions(OpenCL-objects PRIVATE
+		NOMINMAX)
+endif()
```
fodinabor, line 98: In windows.h (which is transitively included by some Vulkan or OpenCL headers) `min` and `max` are defined as macros -> messing with `std::min` and `std::max`.

Also pretty often for windows projects `UNICODE` is defined... but I'm not sure, whether this would make any difference in clvk, as it is not directly using any Win32 apis.. some [docs](https://docs.microsoft.com/en-us/cpp/text/support-for-unicode?view=vs-2019).

src/CMakeLists.txt
```diff
@@ -16,6 +16,8 @@
 if(NOT MSVC)
     add_compile_options(-Wall -W -Wextra)
     add_compile_options(-Wno-ignored-attributes)
+else()
+	set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
```
fodinabor, line 20: When creating libraries on windows it is necessary to explicitly state, which symbols should be exported using `__declspec(dllexport)`. This is usually wrapped in a macro like `CLVK_EXPORT` and then added to all functions that should be exported by the dll.
The quick _fix_ is to export all symbols, this behavior is triggered by this CMake variable.

Though, I'm open to go through clvk and add the `CLVK_EXPORT` a bit more selectively.. just wanted to push the diff before going to bed..

src/CMakeLists.txt
kpet: Thanks for the explanation! Looking at the Vulkan headers I'm wondering whether this shouldn't be guarded by `WIN32` rather than `MSVC`.

src/CMakeLists.txt
kpet: I had a feeling this was the case :). Thanks!

>Though, I'm open to go through clvk and add the CLVK_EXPORT a bit more selectively.. just wanted to push the diff before going to bed..

That would be amazing 🥇 . Let me know if you're going for this or if you'd prefer getting this PR in first.

src/CMakeLists.txt
fodinabor: I agree :)

src/CMakeLists.txt
kpet: I'll get this in as is since it's immediately useful. I've created https://github.com/kpet/clvk/issues/177 to cover making exports a bit more selective.

PR #187, opened 2020-04-08
Fix use of vkGetPhysicalDeviceFeatures2 for Vulkan 1.0

jrprice:
The Vulkan loader may report a newer API version than the underlying Vulkan device, so we cannot rely on the version reported by `vkGetInstanceVersion` to determine whether or not to enable the extension. Instead, enable the extension if it is present, check the device's API version at the point we use the function, and get the KHR suffixed variant via `vkGetInstanceProcAddr` if we are on Vulkan 1.0.

This change allows clvk to work over MoltenVK. I have not tested any other Vulkan 1.0 devices, but my other Vulkan 1.1 devices still work.

### DIFF
--- a/src/device.cpp
+++ b/src/device.cpp
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "device.hpp"
+#include "init.hpp"
 #include "memory.hpp"
 #include "utils.hpp"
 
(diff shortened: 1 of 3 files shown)


tingxingdong:
is this related to issue #173 ?

jrprice:
> is this related to issue #173 ?

It's related to the [comment I posted in that issue](https://github.com/kpet/clvk/issues/173#issuecomment-607439770), but the logs you posted suggested that your problem was elsewhere. Might still be worth a try though.

PR #188, opened 2020-04-08
Infrastructure to support pod push constants

alan-baker:
* Adds a new pod buffer in host memory for pushconstant pod args
  * treated similarly to the pod buffer, but is not a VkBuffer
* adding updated descriptor map parsing

There are two commits. The first commit includes hacks to the compile to force the use of pod push constants (how I tested this). The second commit removes those options so that the new functionality is unused (as per https://github.com/google/clspv/issues/529). 

I only added infrastructure this time to reduce the size of the PR and because not all the clspv functionality is in place to do a complete switch yet. I'd appreciate feedback on whether you think this is a good direction.

### DIFF
--- a/src/kernel.hpp
+++ b/src/kernel.hpp
@@ -143,11 +143,8 @@ struct cvk_kernel_argument_values {
 
             m_pod_buffer = std::move(buffer);
         } else if (m_kernel->has_pod_arguments()) {
+            // TODO(#101): host out-of-memory errors are currently unhandled.
             auto buffer = m_kernel->allocate_pod_pushconstant_buffer();
-            if (buffer == nullptr) {
-                return false;
-            }
-
             m_pod_pushconstant_buffer = std::move(buffer);
         }
 
--- a/src/kernel.hpp
+++ b/src/kernel.hpp
@@ -22,7 +22,6 @@
 #include "program.hpp"
 
 struct cvk_kernel_argument_values;
-struct cvk_command_buffer;
 
 struct cvk_kernel : public _cl_kernel, api_object {
 
--- a/src/kernel.hpp
+++ b/src/kernel.hpp
@@ -22,6 +22,7 @@
 #include "program.hpp"
 
 struct cvk_kernel_argument_values;
+struct cvk_command_buffer;
 
 struct cvk_kernel : public _cl_kernel, api_object {
 
@@ -50,6 +51,11 @@ struct cvk_kernel : public _cl_kernel, api_object {
     bool has_pod_arguments() const {
         return m_entry_point->has_pod_arguments();
     }
+
+    bool has_pod_buffer_arguments() const {
+        return m_entry_point->has_pod_buffer_arguments();
+    }
+
     const std::string& name() const { return m_name; }
     uint32_t num_args() const { return m_args.size(); }
     uint32_t num_set_layouts() const {
@@ -60,6 +66,8 @@ struct cvk_kernel : public _cl_kernel, api_object {
     }
     cvk_program* program() const { return m_program; }
 
+    const std::vector<kernel_argument>& arguments() const { return m_args; }
+
     kernel_argument_kind arg_kind(int index) const {
         return m_args[index].kind;
     }
@@ -68,6 +76,7 @@ struct cvk_kernel : public _cl_kernel, api_object {
 
 private:
     std::unique_ptr<cvk_buffer> allocate_pod_buffer();
+    std::unique_ptr<std::vector<uint8_t>> allocate_pod_pushconstant_buffer();
     friend cvk_kernel_argument_values;
 
     std::mutex m_lock;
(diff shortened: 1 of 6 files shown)


src/program.cpp
```diff
@@ -353,6 +383,10 @@ bool spir_binary::parse_kernel(const std::vector<std::string>& tokens,
         if (!parse_local_arg(arg, tokens, toknum)) {
             return false;
         }
+    } else if (tokens[toknum] == "offset") {
```
kpet, line 386: Another example of a needed change to the descriptor map. We should probably rework the format so that `argKind` comes before `offset` and select what non-descriptor entry is going to be parsed next. Not saying this has to be done now though.

src/queue.cpp
```diff
@@ -399,6 +399,20 @@ cl_int cvk_command_kernel::build() {
                            &m_global_offsets);
     }
 
+    if (m_kernel->has_pod_arguments() &&
+        !m_kernel->has_pod_buffer_arguments()) {
+        for (auto& arg : m_kernel->arguments()) {
+            if (arg.kind == kernel_argument_kind::pod_pushconstant) {
+                CVK_ASSERT(arg.offset + arg.size <=
+                           m_argument_values->pod_pushconstant_buffer().size());
+                vkCmdPushConstants(
```
kpet, line 408: The corresponding ranges have not be declared at pipeline layout creation time (see `m_push_constant_ranges`). This probably won't pass validation.

src/kernel.hpp
```diff
@@ -127,37 +135,55 @@ struct cvk_kernel_argument_values {
...
     bool init() {
-        if (m_kernel->has_pod_arguments()) {
+        if (m_kernel->has_pod_buffer_arguments()) {
             auto buffer = m_kernel->allocate_pod_buffer();
             if (buffer == nullptr) {
                 return false;
             }
 
             m_pod_buffer = std::move(buffer);
+        } else if (m_kernel->has_pod_arguments()) {
```
kpet, line 145: We will need a unit test to exercise the case where POD arguments don't fit in the push constant space passed to clspv. It should pass with default options, expect a failure (probably during kernel creation so we don't fail if none of the problematic kernels are created) when push constants are used and always pass later on when both are used automatically. Further down the line, we will need to extend this to cover cases where multiple buffers are used (SSBO and UBO).

src/program.cpp
```diff
@@ -1484,3 +1529,10 @@ std::unique_ptr<cvk_buffer> cvk_entry_point::allocate_pod_buffer() {
 
     return buffer;
 }
+
+std::unique_ptr<std::vector<uint8_t>>
+cvk_entry_point::allocate_pod_pushconstant_buffer() {
+    std::unique_ptr<std::vector<uint8_t>> buffer(
```
kpet, line 1535: return std::make_unique...

src/kernel.hpp
```diff
@@ -127,37 +135,55 @@ struct cvk_kernel_argument_values {
...
+            auto buffer = m_kernel->allocate_pod_pushconstant_buffer();
+            if (buffer == nullptr) {
+                return false;
+            }
```
kpet, line 149: An exception will have be thrown, this error handling code will never run.

src/kernel.hpp
```diff
@@ -187,6 +213,10 @@ struct cvk_kernel_argument_values {
 
     VkBuffer pod_vulkan_buffer() const { return m_pod_buffer->vulkan_buffer(); }
 
+    std::vector<uint8_t>& pod_pushconstant_buffer() const {
```
kpet, line 216: Either the vector should be `const` or the function should not.

src/program.cpp
alan-baker: Yes, I agree. I have further descriptor map changes coming in clspv though so probably should wait till it stabilizes some more.

src/kernel.hpp
alan-baker: I'm not sure I follow.  Like the VkBuffer case this feeds back to return CL_OUT_OF_RESOURCES when initializing the kernel.

src/kernel.hpp
alan-baker: A unit test makes sense, but currently clvk will never enable pod push constants. Manually adding the option will cause clvk to fail (because -pod-ubo is also specified). This should become testable after clspv starts automating the generation of pod args.

src/queue.cpp
alan-baker: Thanks, I've added this in the latest commit.

src/kernel.hpp
kpet: You could pass `-pod-ubo=0` but I agree it might not be worth spending time on testing intermediate states if the entire move to push constants spans a relatively small period of time.

src/kernel.hpp
kpet: The only thing that `allocate_pod_pushconstant_buffer` does is allocate host memory. If the allocation fails, a `std::bad_alloc` will have been thrown (and either caught somewhere or it will have triggered a call to `std::terminate`). This function can never return a `nullptr`. Maybe renaming the function would make this clearer.

Host memory allocation failures are currently left unhandled in clvk (see #101).

alan-baker: Right, I'd forgotten about that. I will change this to return true for now then. I can add a comment to refer to that issue as well.

src/kernel.hpp
alan-baker: That option can only appear once per compile, so clspv would generate an error unless I change how the option is specified.

kpet: Right, nothing directly testable at this stage then.

PR #189, opened 2020-04-09
Rework logging

### LINKED ISSUE #99: Add support for logging to a file
Controlled via an environment variable set to the path of the output file. When the variable is specified, all logging goes to a file. Any issues with accessing the logfile should be treated as a fatal error and the application terminated.

kpet:
- Move logging code to dedicated source files
- Disable colour output by default when the output is not a terminal
- Add support for logging to a file
- Make it possible to select the output destination (stdout, stderr or file for now)
- Remove a few globals
- Fix a global init race condition (init_logging vs. gLoggingLevel)

Fixes #99


kpet:
Re-run passed: https://dev.azure.com/kpet/clvk/_build/results?buildId=156. Merging.

Issue #162, opened 2020-03-16
TSAN: "unlock of an unlocked mutex (or by a wrong thread)"


jrprice:
TSAN produces this warning:

> WARNING: ThreadSanitizer: unlock of an unlocked mutex (or by a wrong thread)

This is because the `m_lock` mutex in `cvk_program` is locked in `cvk_program::build()`, and then unlocked by the `do_build()` background thread. One fix would be to move the rest of the `build()` method to the background thread as well, which would involve moving a few `cvk_program` members to some sort state attached to that thread (e.g. `m_input_programs`).

kpet:
Nice catch! This is nasty. I broke this when introducing (quick and (too) dirty) support for asynchronous builds, before the first public release of clvk.

Looking at this now, I think I really should have introduced a `cvk_program_build` (or something) object to package all the input/output state required (operation, callback, options, status, etc). I think we probably want the program to own one of these at any given time and to delegate some of the queries (build options, build status, etc) to that new object. I'm sure there are some interesting interactions with separate compilation and linking as well. How does that sound? Do you want to work on a solution?

jrprice:
> I think I really should have introduced a cvk_program_build (or something) object [...]

Yes, this is what I was envisioning as the solution as well.

> Do you want to work on a solution?

Sure, I can take this, will try and get to it soon(ish).

kpet:
Great, thanks!

PR #192, opened 2020-04-17
Use build status to avoid concurrent builds

jrprice:
This is a much simpler fix than we previously discussed in #162. Instead of using the mutex to signal when the build has completed, just rely on the fact that we already have this information in the build status. The mutex is now just used to avoid concurrent calls to clBuildProgram.

The check is slightly conservative; this doesn't allow the same program to be built for two different devices concurrently (which would be handled by the larger refactor into a cvk_program_build that we briefly discussed). I think this is fine for now however, given that the rest of the program build logic assumes single device anyway.


kpet:
You actually did in one of the commit messages, I hadn't spotted that. Nothing to discuss then.

PR #195, opened 2020-04-20
Parse module-scope specialization constants

alan-baker:
* Parses module scope specialization constants in the descriptor map
* Currently supports workgroup size
  * updated how workgroup size spec ids are generated
* Update clspv submodule

### DIFF
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -445,6 +445,30 @@ bool spir_binary::parse_pushconstant(const std::vector<std::string>& tokens,
     return true;
 }
 
+bool spir_binary::parse_specconstant(const std::vector<std::string>& tokens,
+                                     int toknum) {
+    auto name = tokens[toknum++];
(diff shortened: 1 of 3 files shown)


alan-baker:
Builds are failing because the SPIRV-LLVM-Translator isn't up-to-date with the LLVM used by clspv.

src/program.cpp
```diff
@@ -445,6 +445,30 @@ bool spir_binary::parse_pushconstant(const std::vector<std::string>& tokens,
     return true;
 }
 
+bool spir_binary::parse_specconstant(const std::vector<std::string>& tokens,
+                                     int toknum) {
+    auto name = tokens[toknum++];
+    spec_constant constant;
+    if (name == "workgroup_size_x") {
```
kpet, line 452: We may want to turn this into a map lookup in the future to make things more compact.

PR #197, opened 2020-04-21
Support work dimensions as a spec constant

alan-baker:
* Clspv default for -work-dim is true so stop passing that option
* Parse work dim spec constant in descriptor map
* Populate work dim spec constant value if necessary
* Add test for work dims

### DIFF
--- a/tests/api/enqueue.cpp
+++ b/tests/api/enqueue.cpp
@@ -88,3 +88,52 @@ TEST_F(WithCommandQueue, KernelNoArguments) {
     clReleaseEvent(event);
     ASSERT_EQ(status, CL_COMPLETE);
 }
+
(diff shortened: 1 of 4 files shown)


tests/api/enqueue.cpp
```diff
@@ -88,3 +88,52 @@ TEST_F(WithCommandQueue, KernelNoArguments) {
     clReleaseEvent(event);
     ASSERT_EQ(status, CL_COMPLETE);
 }
+
+TEST_F(WithCommandQueue, WorkDim) {
```
kpet, line 92: This is covered by a conformance test but it also covers a bunch of other functions as well. Good to have as a unit test/(base for a )benchmark.

PR #198, opened 2020-04-23
Fix TSAN issues for async builds

jrprice:
TSAN flagged a couple of issues in the `async_build` CTS test.

The first is around how the device status is read and written. I've made it atomic; the accesses use sequential consistency by default which I think is fine here and keeps the code simple, though I can switch to acquire/release if you prefer. I also had to make the background thread call the program completion callback *before* setting the device status. This is what the `async_build` test appears to assume will happen, but I couldn't see any spec language that requires this.

There's also some some cleanup to fix a thread leak.


src/program.hpp
```diff
@@ -395,15 +396,15 @@ struct cvk_program : public _cl_program, api_object {
...
         m_thread->join();
-        delete m_thread;
     }
 
     void complete_operation(cvk_device* device, cl_build_status status) {
-        m_dev_status[device] = status;
         if (m_operation_callback != nullptr) {
             m_operation_callback(this, m_operation_callback_data);
         }
+        m_dev_status[device] = status;
```
kpet, line 407: This doesn't make sense to me. It essentially means that the callback can't observe the status of the build which feels wrong. In general, the spirit of the spec wrt. callbacks is that they are called at some undefined point after a given event. I think the test should probably loop until it confirms the callback has been called, probably with a max number of attempts so test doesn't deadlock on failure. I think we should probably clarify the spec as well and add two tests where he callback is required to observe a successful and failed build.

jrprice: Yeah I agree this did feel weird. I opted for this approach to avoid making changes to CTS that might break existing implementations, but I'm also fine with clarifying the spec and modifying CTS.

jrprice:
I've reverted setting the device status to before firing the callback as per recent spec/CTS updates. PTAL.

PR #224, opened 2020-04-29
Remove unnecessary compile options

alan-baker:
* Remove unnecessary options
  * -work-dim is on by default
  * -cluster-pod-kernel-args is on by default
  * -pod-ubo removal allows clspv to choose implementation
* Added some tests to confirm both ubo and push constant pod args work
* Update clspv

### DIFF
--- a/tests/api/enqueue.cpp
+++ b/tests/api/enqueue.cpp
@@ -141,7 +141,7 @@ TEST_F(WithCommandQueue, WorkDim) {
 TEST_F(WithCommandQueue, PodUBO) {
     static const char* program_source =
         "kernel void test(global int* out, int a, int4 b, int c) { *out = a + "
-        "b.x + c; }";
+        "b.y + c; }";
 
     auto kernel = CreateKernel(program_source, " -pod-ubo ", "test");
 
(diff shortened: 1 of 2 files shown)


tests/api/enqueue.cpp
```diff
@@ -137,3 +137,81 @@ TEST_F(WithCommandQueue, WorkDim) {
     EnqueueUnmapMemObject(buffer, data);
     Finish();
 }
+
+TEST_F(WithCommandQueue, PodUBO) {
+    static const char* program_source =
+        "kernel void test(global int* out, int a, int4 b, int c) { *out = a + "
+        "b.x + c; }";
```
kpet, line 144: Maybe using another component than `x` would make for a slightly better test?

tests/api/enqueue.cpp
```diff
@@ -137,3 +137,81 @@ TEST_F(WithCommandQueue, WorkDim) {
...
+
+    // Unmap the buffer
+    EnqueueUnmapMemObject(buffer, data);
+    Finish();
+}
+
+TEST_F(WithCommandQueue, PodPushConstant) {
+    static const char* program_source =
+        "kernel void test(global int* out, int a, int4 b, int c) { *out = a + "
+        "b.x + c; }";
```
kpet, line 183: Same comment here.

tests/api/enqueue.cpp
alan-baker: I'll use different components in each test.

alan-baker:
The mac build had a problem getting a resource. Seems unrelated to the change.

kpet:
Yes, I was looking at that. It seems indeed unrelated, I've just retriggered the job (it's the only debug build at the moment).

kpet:
Passed on retry, thanks for the changes!

PR #228, opened 2020-05-06
Enable {8,16}-bit storage extensions and features

jrprice:
Will need more work to enable the right Clspv options once https://github.com/google/clspv/pull/580 lands.

### DIFF
--- a/src/device.cpp
+++ b/src/device.cpp
@@ -125,11 +125,11 @@ void cvk_device::init_features(VkInstance instance) {
     m_features_float16_int8.sType =
         VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_FLOAT16_INT8_FEATURES_KHR;
     m_features_variable_pointer.sType =
-        VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VARIABLE_POINTERS_FEATURES;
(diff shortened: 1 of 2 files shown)


src/device.cpp
```diff
@@ -125,6 +126,10 @@ void cvk_device::init_features(VkInstance instance) {
         VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_FLOAT16_INT8_FEATURES_KHR;
     m_features_variable_pointer.sType =
         VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VARIABLE_POINTERS_FEATURES;
+    m_features_8bit_storage.sType =
+        VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_8BIT_STORAGE_FEATURES;
```
alan-baker, line 130: Should this be VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_8BIT_STORAGE_FEATURES_KHR (KHR suffix) for older implementations? Same for 16-bit.

jrprice: Yes, will fix.

jrprice: Done, also did the variable pointers one for consistency.

PR #236, opened 2020-05-13
Add memory barrier between host writes and kernels

jrprice:
Without this I am getting some CTS failures on some devices.


jrprice:
You're right. I've dug some more, and it wasn't the host->device consistency that was the core problem. I'll open a different PR with an annoying workaround for what does seem to be a driver bug.

PR #238, opened 2020-05-29
Batch kernel dispatches in the same command buffer

jrprice:
Summary:
* `cvk_command_queue` maintains a `cvk_command_kernel_group`
  - Add new commands to it as long as they are batchable
  - End the group when a non-batchable command is enqueued
* `cvk_command_kernel_group` has a `cvk_command_buffer`
  - Used for all of its kernel commands
* `cvk_command_kernel::build()` now takes a `cvk_command_buffer` parameter
  - Called by `cvk_command_kernel_group::add_kernel()`
* Non-batchable kernel commands are in their own `cvk_command_kernel_group`
  - Avoids splitting batchable/non-batchable logic in `cvk_command_kernel`
* Added `CLVK_MAX_BATCH_SIZE` environment variable
  - Some apps might need to limit this to avoid device timeout
  - Defaults to 10000

Having `cvk_command_kernel::do_action()` be a never-called method is a little awkward. The simplest alternative is to have two different codepaths for batched/non-batched kernels (i.e. two different `::build()` methods, one that receives a command buffer and another that creates one), but this also felt a little unpleasant to me.

This is mostly either a performance win or neutral. However, this does give some performance regressions for certain benchmarks on certain devices. This seems to be related to push constant updates for POD args. I'm still investigating and will try and address in a follow-up soon (though I also don't mind if you want to hold off on this PR until it's resolved).

### DIFF
--- a/src/queue.cpp
+++ b/src/queue.cpp
@@ -775,6 +775,11 @@ cl_int cvk_command_kernel_group::do_action() {
     for (auto& kcmd : m_kernel_commands) {
         auto ev = kcmd->event();
         if (profiling) {
+            auto ts_group_submit =
+                m_event->get_profiling_info(CL_PROFILING_COMMAND_SUBMIT);
+            ev->set_profiling_info(CL_PROFILING_COMMAND_SUBMIT,
+                                   ts_group_submit);
+
             if (gQueueProfilingUsesTimestampQueries) {
                 auto perr = kcmd->set_profiling_info_from_query_results();
                 // Report the first error if no errors were present
--- a/src/queue.cpp
+++ b/src/queue.cpp
@@ -126,8 +126,8 @@ cl_int cvk_command_queue::enqueue_command(cvk_command* cmd, _cl_event** event) {
             m_kernel_group = nullptr;
         }
 
-        // Create a standalone batch for a non-batchable kernel command
         if (is_kernel_command) {
+            // Create a standalone batch for a non-batchable kernel command
             auto group = new cvk_command_kernel_group(this);
             group->set_dependencies(cmd->dependencies());
 
@@ -139,9 +139,9 @@ cl_int cvk_command_queue::enqueue_command(cvk_command* cmd, _cl_event** event) {
                 return CL_OUT_OF_RESOURCES;
             }
             m_groups.back()->commands.push_back(group);
+        } else {
+            m_groups.back()->commands.push_back(cmd);
         }
-
-        m_groups.back()->commands.push_back(cmd);
     }
 
     cvk_debug_fn("enqueued command %p, event %p", cmd, cmd->event());
--- a/src/queue.cpp
+++ b/src/queue.cpp
@@ -124,21 +124,14 @@ cl_int cvk_command_queue::enqueue_command(cvk_command* cmd, _cl_event** event) {
         }
 
         if (is_kernel_command(cmd)) {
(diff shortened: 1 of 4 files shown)


kpet:
To break the radio silence: I've had a look at the change and it looks overall nice and simple but I probably won't be able to dive into it before the week-end. On POD arguments, were you thinking of something similar to what I alluded to in https://github.com/kpet/clvk/commit/be04d05427bf1c64e5d3523d8c67fc12020b5bc8?

jrprice:
> On POD arguments, were you thinking of something similar to what I alluded to in [be04d05](https://github.com/kpet/clvk/commit/be04d05427bf1c64e5d3523d8c67fc12020b5bc8)?

This doesn't seem to be the cause of the regressions. I hacked a change to do a single push constant update for POD args and it didn't help. But I do think that's a good idea in any case.

src/queue.cpp
```diff
@@ -51,11 +58,91 @@ cvk_command_queue::~cvk_command_queue() {
...
+            auto group = new cvk_command_kernel_group(this);
+            group->set_dependencies(cmd->dependencies());
+
+            err = group->add_kernel(static_cast<cvk_command_kernel*>(cmd));
+            if (err != CL_SUCCESS) {
+                return err;
+            }
+            if (!group->end()) {
+                return CL_OUT_OF_RESOURCES;
+            }
```
kpet, line 140: Maybe this could be an init function to `cvk_command_kernel_group` that would prepare the group from a single command? Back to your question from the commit message:

>Having cvk_command_kernel::do_action() be a never-called method is a little awkward. [...]

Yeah and this forces us to create a group for every non-batchable kernel command but it's not the end of the world. I'm wondering whether the alternative wouldn't look a bit better though (not sure at all TBH).

jrprice: I've gone back to having a separate `cvk_command_kernel::build()` for non-batched kernels that creates its own command buffer, rather than having to create a standalone `cvk_command_kernel_group`. It's actually not as awkward as it was last time I tried it.

jrprice:
Addressed review comments, PTAL.

Side note: Why is `cvk_command_queue::m_groups` a `deqeue`? AFAICT (with the way the code is currently written) it will never have more than one element.

kpet:
I've run a few tests locally using an NVidia device and it seems this change leads to a crash in `test_multiples` as well as a hang in `test_profiling` (`read_image_float` is the first hanging test). Have you observed similar issues?

>Side note: Why is cvk_command_queue::m_groups a deqeue

I wanted to have the ability to have multiple flush groups in flight but I agree with you that with the current code, it will never contain more than one. I have no concrete plans to make use of this and I'm fine with removing it if you think the associated complexity is a burden.

jrprice:
> I've run a few tests locally using an NVidia device and it seems this change leads to a crash in `test_multiples` as well as a hang in `test_profiling` (`read_image_float` is the first hanging test). Have you observed similar issues?

I've fixed the hang - `cvk_command_kernel_group` was leaving the command pool locked in between adding kernels, which caused a deadlock when enqueueing a non-kernel command that needs to use the command pool. The fix I've pushed is simple but not entirely pretty.

I couldn't reproduce the `test_multiples` crash though, do you know which subtest is crashing?

kpet:
I can no longer reproduce the crash on `test_multiples`, either your latest patch fixed it or it was something in my environment. All tests are now passing at my end. I'll review the latest changes ASAP.

PR #242, opened 2020-06-30
Create command pool with RESET_COMMAND_BUFFER_BIT

jrprice:
Avoids creeping memory growth (and subsequent crashes) with some drivers, and gives performance wins in some cases.

For example, the Arm Mali developer guide [says](https://developer.arm.com/documentation/101897/0200/cpu-overheads/command-pools-for-vulkan?lang=en):

> Command pools do not automatically recycle memory from deleted command buffers unless created with the RESET_COMMAND_BUFFER_BIT flag. Pools without this flag do not recycle their memory until the application resets the pool.

It later recommends *against* using `RESET_COMMAND_BUFFER_BIT` (due to potential increase in CPU overheads), but the alternative is using `vkResetCommandPool`, which is more difficult to fit into clvk right now. I didn't see any significant performance drops when using `RESET_COMMAND_BUFFER_BIT` in my own benchmarking, and there were several significant performance gains.

### DIFF
--- a/src/device.cpp
+++ b/src/device.cpp
@@ -43,9 +43,8 @@ void cvk_device::init_properties(VkInstance instance) {
     m_driver_properties.pNext = nullptr;
     if (m_properties.apiVersion < VK_MAKE_VERSION(1, 1, 0)) {
         // Use the extension on Vulkan 1.0 platforms
-        auto func = reinterpret_cast<PFN_vkGetPhysicalDeviceProperties2KHR>(
-            vkGetInstanceProcAddr(instance,
-                                  "vkGetPhysicalDeviceProperties2KHR"));
+        auto func =
+            GET_INSTANCE_PROC(instance, vkGetPhysicalDeviceProperties2KHR);
         if (!func) {
             cvk_fatal(
                 "Failed to get pointer to vkGetPhysicalDeviceProperties2KHR()");
--- a/src/device.cpp
+++ b/src/device.cpp
@@ -33,24 +33,19 @@ void cvk_device::init_driver_behaviors(VkInstance instance) {
 
     cvk_info("Initialising driver behaviors");
 
-    if (!(m_properties.apiVersion >= VK_MAKE_VERSION(1, 2, 0) ||
-          is_vulkan_extension_enabled(
-              VK_KHR_DRIVER_PROPERTIES_EXTENSION_NAME))) {
-        cvk_warn("VK_KHR_driver_properties not supported");
-        cvk_warn("Using default Vulkan driver behaviors.");
-        return;
-    }
+    // Disable all driver-specific behaviors by default
+    m_driver_behaviors = 0;
 
-    // Get physical device properties
-    VkPhysicalDeviceProperties2KHR properties;
-    properties.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PROPERTIES_2_KHR;
-    properties.pNext = &m_driver_properties;
-    m_driver_properties.sType =
-        VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DRIVER_PROPERTIES_KHR;
-    m_driver_properties.driverID = static_cast<VkDriverIdKHR>(0);
-    m_driver_properties.pNext = nullptr;
-    if (m_properties.apiVersion < VK_MAKE_VERSION(1, 1, 0)) {
-        // Use the extension on Vulkan 1.0 platforms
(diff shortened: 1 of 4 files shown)


jrprice:
> The first approach that comes to mind would be to introduce a small framework to manage driver-specific behaviour.

Yes, we've been considering doing something like this to conditionally enable some of Clspv's workarounds for driver bugs too. I've added a simple implementation of this and used it to conditionally enable the `RESET_COMMAND_BUFFER_BIT` flag.

src/device.cpp
```diff
@@ -29,6 +29,56 @@ cvk_device* cvk_device::create(cvk_platform* platform, VkInstance instance,
...
+    properties.pNext = &m_driver_properties;
+    m_driver_properties.sType =
+        VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DRIVER_PROPERTIES_KHR;
+    m_driver_properties.driverID = static_cast<VkDriverIdKHR>(0);
+    m_driver_properties.pNext = nullptr;
+    if (m_properties.apiVersion < VK_MAKE_VERSION(1, 1, 0)) {
+        // Use the extension on Vulkan 1.0 platforms
+        auto func = reinterpret_cast<PFN_vkGetPhysicalDeviceProperties2KHR>(
+            vkGetInstanceProcAddr(instance,
+                                  "vkGetPhysicalDeviceProperties2KHR"));
```
kpet, line 48: There is a `GET_INSTANCE_PROC` macro that does just that.

src/device.cpp
```diff
@@ -29,6 +29,56 @@ cvk_device* cvk_device::create(cvk_platform* platform, VkInstance instance,
...
 }
 
+void cvk_device::init_properties(VkInstance instance) {
+
+    cvk_info("Initialising properties");
+
+    // Get physical device properties
+    VkPhysicalDeviceProperties2KHR properties;
+    properties.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PROPERTIES_2_KHR;
+    properties.pNext = &m_driver_properties;
```
kpet, line 39: Doing this unconditionally assumes that the extension is available or the version at least 1.2. This has been a problem for some people in the past in feature initialisation (see `cvk_device::init_features`). Given that device properties are necessary for extension initialisation, I think we'll have to query them twice (I can't think of a better way right now at least).

src/device.cpp
```diff
@@ -29,6 +29,56 @@ cvk_device* cvk_device::create(cvk_platform* platform, VkInstance instance,
...
+                 m_driver_properties.conformanceVersion.minor,
+                 m_driver_properties.conformanceVersion.subminor,
+                 m_driver_properties.conformanceVersion.patch);
+    }
+
+    // Alter behavior based on the target Vulkan device and driver version
+    if (m_driver_properties.driverID == VK_DRIVER_ID_ARM_PROPRIETARY_KHR) {
+        // Workaround for resource management bug on Mali GPUs.
+        // TODO: Make conditional on driver version when fixed in the driver.
+        m_use_reset_command_buffer_bit = true;
```
kpet, line 78: FYI, I was thinking of introducing a global enum for all driver-specific bits of behaviour so we can easily get an overview. I'm perfectly fine with what you've done here but thought I'd start getting the idea out there.

src/device.cpp
jrprice: A driver that doesn't support the extension is required to just skip the structure. From the [Vulkan spec](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#fundamentals-validusage-pNext):
> Any component of the implementation (the loader, any enabled layers, and drivers) must skip over, without processing (other than reading the sType and pNext members) any extending structures in the chain not defined by core versions or extensions supported by that component.

Are you saying that some Vulkan implementations don't actually do this? I tried to make the rest of the function robust to the extension not being supported (e.g. the `if (m_driver_properties.driverID != 0)` check). I'm happy to add a query for the extension if this is a real problem though.

kpet: True. Yeah, it seems some implementations don't do this correctly. Let's try to keep it simple first as what you've done ought to work on a correct Vulkan implementation.

jrprice: Actually I've decided to do the extension check after all. It's useful to report when the extension is not present so we know when these behaviors are not being set for a device. Querying properties should be fast anyway.

src/device.cpp
jrprice: I was thinking of doing this too, so I've added that to this PR as well. LMK if this isn't quite what you had in mind.

kpet: Exactly what I had in mind, thanks!

PR #244, opened 2020-07-08
Batch descriptor set updates

jrprice:
Instead of calling `vkUpdateDescriptorSets` for each kernel argument separately, collect all the `VkWriteDescriptor` structures and call it just once.

This yields significant improvements to kernel enqueue latency on some Vulkan implementations.

### DIFF
--- a/src/kernel.cpp
+++ b/src/kernel.cpp
@@ -97,7 +97,7 @@ bool cvk_kernel::setup_descriptor_sets(
 
     // Make enough space to store all descriptor write structures
     size_t max_descriptor_writes =
-        m_args.size() + program()->literal_sampler_descs().size() + 1;
+        m_args.size() + program()->literal_sampler_descs().size();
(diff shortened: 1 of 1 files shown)


src/kernel.cpp
```diff
@@ -95,13 +95,24 @@ bool cvk_kernel::setup_descriptor_sets(
         return false;
     }
 
+    // Make enough space to store all descriptor write structures
+    size_t max_descriptor_writes =
+        m_args.size() + program()->literal_sampler_descs().size() + 1;
```
kpet, line 100: Why +1?

jrprice: Didn't occur to me that POD args would be included in m_args as well :-)

jrprice:
FYI I'm toying with some follow ups to improve things further. Specifically, reusing the descriptors for subsequent launches if the arguments haven't changed saves a little more time here. Could take things even further and maintain a cache of descriptor sets (where the args are the key), but not sure this would add much value right now.

PR #245, opened 2020-07-16
Reuse descriptor sets between kernel launches

jrprice:
Summary:

* Make descriptor sets and POD buffer part of the `cvk_kernel_kernel_argument_values` structure

* Reuse existing `cvk_kernel_argument_values` instance when enqueueing
  kernel with same args multiple times

* Create new `cvk_kernel_argument_values` instance when setting argument
  for a kernel object that has already been enqueued

* Handle shared ownership of `cvk_kernel_argument_values` via `std::shared_ptr`

Notes:

This ended up being more intrusive than I originally imagined it would be. Some of the changes around how `cvk_kernel_argument_values` ownership and resource retention is handled are not directly related to descriptor set reuse, but I think are a little cleaner than the previous implementation (feel free to disagree!). If this is going too far, the second commit here can be removed, and we can have descriptor set reuse via a second dedicated structure (`cvk_kernel_descriptors`) instead via the first commit on its own.

### DIFF
--- a/src/kernel.hpp
+++ b/src/kernel.hpp
@@ -103,7 +103,13 @@ struct cvk_kernel_argument_values {
           m_kernel_resources(other.m_kernel_resources),
           m_local_args_size(other.m_local_args_size) {}
 
-    ~cvk_kernel_argument_values() {}
+    ~cvk_kernel_argument_values() {
+        for (auto ds : m_descriptor_sets) {
+            if (ds != VK_NULL_HANDLE) {
+                m_entry_point->free_descriptor_set(ds);
+            }
+        }
+    }
 
     static std::shared_ptr<cvk_kernel_argument_values>
     create(cvk_entry_point* entry_point) {
--- a/src/kernel.hpp
+++ b/src/kernel.hpp
@@ -22,66 +22,25 @@
 #include "program.hpp"
 
 struct cvk_kernel_argument_values;
-struct cvk_kernel_descriptors;
-
-struct cvk_kernel_descriptors : refcounted {
-    cvk_kernel_descriptors(cvk_entry_point* entry_point)
-        : m_entry_point(entry_point), m_descriptor_sets{VK_NULL_HANDLE} {}
-
-    virtual ~cvk_kernel_descriptors() {
-        for (auto ds : m_descriptor_sets) {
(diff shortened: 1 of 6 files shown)


src/kernel.hpp
```diff
@@ -95,22 +91,23 @@ using cvk_kernel_holder = refcounted_holder<cvk_kernel>;
...
     cvk_kernel_argument_values(const cvk_kernel_argument_values& other)
-        : m_kernel(other.m_kernel), m_pod_buffer(nullptr),
-          m_owns_resources(false), m_kernel_resources(other.m_kernel_resources),
+        : m_entry_point(other.m_entry_point), m_is_enqueued(false),
+          m_args(m_entry_point->args()), m_pod_arg(nullptr),
+          m_kernel_resources(other.m_kernel_resources),
           m_local_args_size(other.m_local_args_size) {}
 
-    ~cvk_kernel_argument_values() { release_resources(); }
+    ~cvk_kernel_argument_values() {}
```
kpet, line 106: The destructor should probably free the descriptor sets.

jrprice: Whoops, don't know how I missed that :-/

Fixed.

jrprice:
> I'm wondering whether we could tie the lifetime of argument resources to argument value objects instead of individual kernel commands. I don't think it makes a big practical difference but the current scheme seems to duplicate what the shared pointer is already doing.

The issue here is that the kernel object also holds a reference to the argument value object, and you don't want the kernel object to retain the argument resources (this was issue #157). The lifetime of the argument resources is inherently linked to the lifetime of the kernel commands that use them, so I think it makes sense as it currently is, even if there is some resource management duplication.

kpet:
Oh yes, I remember now. That change was not pretty :). It's good you managed to get rid of most of the ugliness I suggested to Alan.

src/kernel.hpp
kpet: Thanks. I almost missed it as well. Neither the validation layers nor valgrind caught it.

kpet:
Hmm, it seems the tests are now crashing.

jrprice:
> Hmm, it seems the tests are now crashing.

Investigating now.

jrprice:
Fixed - was missing initialization.

PR #250, opened 2020-08-05
Switch to use embedded reflection

alan-baker:
Change clvk to use clspv's embedded reflection in the SPIR-V binary.

* Offline and online compilation now share parsing of reflection information
* "binary" returned by clvk no longer contains the descriptor map
  * left the magic number to minimize overall number of changes
* if the Vulkan implementation does not support VK_KHR_non_semantic_info then the reflection information is stripped after parsing
  * the stripped binary is stored separately so that the binary with embedded reflection may be returned for clGetProgramInfo
* update submodules

### DIFF
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -750,9 +750,6 @@ cl_build_status cvk_program::compile_source(const cvk_device* device) {
     if (build_from_il) {
         cmd += " -x ir ";
     }
-    cmd += " -descriptormap=";
-    cmd += descriptor_map_file;
-    cmd += " ";
     cmd += clspv_input_file;
     cmd += " ";
     cmd += options;
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -28,8 +28,8 @@
 #endif
 #include "spirv-tools/linker.hpp"
 #include "spirv-tools/optimizer.hpp"
-#include "spirv/unified1/spirv.hpp"
 #include "spirv/unified1/NonSemanticClspvReflection.h"
+#include "spirv/unified1/spirv.hpp"
 
 #include "init.hpp"
 #include "log.hpp"
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -279,7 +279,7 @@ spv_result_t parse_reflection(void* user_data,
                 auto mask = helper->constants[inst->words[7]];
                 uint32_t coords = mask & clspv::kSamplerNormalizedCoordsMask;
                 bool normalized_coords =
-                    coords != clspv::CLK_NORMALIZED_COORDS_TRUE;
+                    coords == clspv::CLK_NORMALIZED_COORDS_TRUE;
                 cl_addressing_mode addressing;
                 switch (mask & clspv::kSamplerAddressMask) {
                 case clspv::CLK_ADDRESS_NONE:
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -22,12 +22,14 @@
 
 #include <vulkan/vulkan.h>
 
+#include "clspv/Sampler.h"
 #ifdef CLSPV_ONLINE_COMPILER
 #include "clspv/Compiler.h"
 #endif
 #include "spirv-tools/linker.hpp"
 #include "spirv-tools/optimizer.hpp"
-#include "spirv/1.0/spirv.hpp"
+#include "spirv/unified1/spirv.hpp"
+#include "spirv/unified1/NonSemanticClspvReflection.h"
 
 #include "init.hpp"
 #include "log.hpp"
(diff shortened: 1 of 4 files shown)


src/program.cpp
```diff
@@ -47,14 +49,292 @@ struct membuf : public std::streambuf {
...
+    spir_binary* binary;
+};
+
+spv_result_t parse_reflection(void* user_data,
+                              const spv_parsed_instruction_t* inst) {
+    // Helper function to map instruction to argument type.
+    auto inst_to_arg_kind = [](uint32_t inst) {
+        switch (static_cast<NonSemanticClspvReflectionInstructions>(inst)) {
+        case NonSemanticClspvReflectionArgumentStorageBuffer:
+            return kernel_argument_kind::buffer;
```
kpet, line 65: We probably should rename  values for `kernel_argument_kind` ultimately.

src/program.cpp
```diff
@@ -47,14 +49,292 @@ struct membuf : public std::streambuf {
...
+                helper->binary->add_spec_constant(spec_constant::work_dim,
+                                                  dim_id);
+                break;
+            }
+            case NonSemanticClspvReflectionPushConstantGlobalOffset:
+            case NonSemanticClspvReflectionPushConstantEnqueuedLocalSize:
+            case NonSemanticClspvReflectionPushConstantGlobalSize:
+            case NonSemanticClspvReflectionPushConstantRegionOffset:
+            case NonSemanticClspvReflectionPushConstantNumWorkgroups:
+            case NonSemanticClspvReflectionPushConstantRegionGroupOffset: {
```
kpet, line 266: I'm wondering whether a single `NonSemanticClspvReflectionPushConstant` instruction with an enum argument to specify the push constant wouldn't have been better. All code that needs to handle push constants generically will have to maintain an ever-growing list of instructions with the current scheme.

src/program.cpp
```diff
@@ -189,559 +458,56 @@ bool spir_binary::validate() const {
...
-                arg.size = entry.kernel_arg_data.pod_arg_size;
             }
-        }
+#undef msgtpl
+        };
 
-        m_dmaps[entry.kernel_arg_data.kernel_name].push_back(arg);
+    spvtools::Optimizer opt(m_target_env);
+    opt.SetMessageConsumer(consumer);
+    opt.RegisterPass(spvtools::CreateStripReflectInfoPass());
```
kpet, line 488: Maybe the pass factory should have had "clspv" in the name to be consistent with the instruction set name.

src/program.cpp
```diff
@@ -47,14 +49,292 @@ struct membuf : public std::streambuf {
     }
 };
 
+struct reflection_helper {
```
kpet, line 52: Not that it matters too much but: `reflection_parse_data`?

src/program.cpp
```diff
@@ -1213,15 +975,32 @@ void cvk_program::do_build() {
...
     }
 
+    // Strip the reflection information if non-semantic info is not supported
+    // by the Vulkan implementation. This stripped binary is stored separately
+    // from |m_binary| because clvk needs to be able to provide the binary with
+    // reflection information for clGetProgramInfo.
+    const uint32_t* spir_data = m_binary.spir_data();
+    size_t spir_size = m_binary.spir_size();
+    if (!device->is_vulkan_extension_enabled(
+            "VK_KHR_shader_non_semantic_info")) {
```
kpet, line 985: Could use the extension name macro to guarantee that the name is correct.

src/program.cpp
alan-baker: I considered that when designing the instruction set, but it only really simplifies this switch statement. You still need the conversion switch statement in the other method. I felt this is easier to validate than also validating specific enum values. That said, this could change in future versions of the non-semantic instruction set.

src/program.cpp
alan-baker: The reflection stripping pass existed before my instruction set and handles non-semantic instructions generally and also the HLSL reflection instructions and decorations.

alan-baker:
I'll take a look at the CTS failures. 

>We'll probably want to do more renaming/refactoring but that can happen after this change.

Agreed, but I wanted to keep this change as minimal as reasonably possible.

src/program.cpp
```diff
@@ -47,14 +49,292 @@ struct membuf : public std::streambuf {
...
+            case NonSemanticClspvReflectionLiteralSampler: {
+                // Track descriptor set and binding. Decode the sampler mask.
+                auto descriptor_set = helper->constants[inst->words[5]];
+                if (descriptor_set >= spir_binary::MAX_DESCRIPTOR_SETS)
+                    return SPV_ERROR_INVALID_DATA;
+                auto binding = helper->constants[inst->words[6]];
+                auto mask = helper->constants[inst->words[7]];
+                uint32_t coords = mask & clspv::kSamplerNormalizedCoordsMask;
+                bool normalized_coords =
+                    coords != clspv::CLK_NORMALIZED_COORDS_TRUE;
```
alan-baker, line 282: This is backwards and the cause of the write image test failures.

alan-baker:
I believe the SPIRV-Tools linker/optimizer is having problems with the extended instructions. Need to investigate how to fix that.

alan-baker:
> I believe the SPIRV-Tools linker/optimizer is having problems with the extended instructions. Need to investigate how to fix that.

SPIRV-Tools optimizer representation is reordering the module and producing an invalid binary.

alan-baker:
The last push fixes the compiler conformance test crash. All those tests match the checked in results. With that and the sampler fix, CTS should be good now.

src/program.cpp
kpet: True. Enum validation could be done with just a range check but I'm not overly worried about this (we're unlikely to add new push constants every week) and, as you say, it's easy enough to evolve it.

src/program.cpp
kpet: Ok, that's a bit counter-intuitive. I'm wondering whether generic stripping of all non-semantic instructions shouldn't be a separate pass. I'm not familiar enough with the HLSL bits to have an informed opinion though.

kpet:
All the usual CTS tests are now passing. All looking good, thanks!

PR #254, opened 2020-08-17
Remove redundant get()

alan-baker:
* Addresses clang-tidy warning


dneto0:
The rule is described at http://clang.llvm.org/extra/clang-tidy/checks/readability-redundant-smartptr-get.html

PR #264, opened 2020-09-09
Added further details when exceeding workgroup size limits

judicaelclair:
Signed-off-by: Judicaël CLAIR <judicael.clair@arm.com>


kpet (approved):
Thanks!

Issue #265, opened 2020-09-09
Add simple local work size selection heuristics


kpet:
clvk currently always picks {1,1,1} when the application doesn't provide a local work size. This leads to poor performance and a large number of workgroups with some workloads that some Vulkan implementations can't accommodate.

PR #275, opened 2020-10-15
Add a simple work-group size selection heuristic

jrprice:
Start at (1,1,1) and alternately double the X and Y dimensions while it is still valid to do so, capped to a total size of 64.

It's pretty crude, but this yields huge gains on a workload that I'm looking at that uses a `NULL` work-group size. I've left TODOs for some of the many potential areas for improvement in the comments.

Contributes to #265.

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -3481,7 +3481,7 @@ cl_int CLVK_API_CALL clEnqueueNDRangeKernel(
 
     std::array<uint32_t, 3> goff = {0, 0, 0};
     std::array<uint32_t, 3> gws = {1, 1, 1};
-    std::array<uint32_t, 3> lws = {1, 1, 1}; // FIXME pick a sensible default
+    std::array<uint32_t, 3> lws = {1, 1, 1};
 
     for (cl_uint i = 0; i < work_dim; i++) {
         gws[i] = global_work_size[i];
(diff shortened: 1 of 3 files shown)


src/api.cpp
```diff
@@ -3493,6 +3493,55 @@ cl_int CLVK_API_CALL clEnqueueNDRangeKernel(
...
+            // TODO: Allow non-uniform sizes if supported?
+            uint32_t newY = lws[1] * 2;
+            if (gws[1] % newY == 0 &&
+                newY <= limits.maxComputeWorkGroupCount[1] &&
+                lws[0] * newY <= maxSize) {
+                lws[1] = newY;
+                changed = true;
+            }
+
+            // TODO: Consider increasing the Z dimension as well?
```
kpet, line 3541: I suggest writing this as a loop over dimensions (probably excluding Z for now). This would shorten the code and make it more generic.

src/api.cpp
```diff
@@ -3493,6 +3493,55 @@ cl_int CLVK_API_CALL clEnqueueNDRangeKernel(
         }
     }
 
+    // Try to pick a sensible work-group size if the user didn't specify one.
+    if (local_work_size == nullptr) {
+        lws[0] = lws[1] = lws[2] = 1;
```
kpet, line 3498: The initialiser above will have done that.

kpet (changes requested):
Looks like a reasonable start. Doing anything will be better than {1,1,1} :). I suggest you push the logic into `cvk_device`.

src/api.cpp
jrprice: Done.

src/api.cpp
jrprice: Agreed. I've left this in now that I've moved the code into `cvk_device` though, to avoid assuming that the caller has done that initialization.

jrprice:
> I suggest you push the logic into cvk_device.

Done, PTAL.

src/api.cpp
kpet: Makes sense.

kpet (approved):
Thanks!

PR #276, opened 2020-11-06
Support creating an image with a host pointer

### LINKED ISSUE #108: Correctly perform image layout transitions
Things work on nVidia's driver without them but this is definitely incorrect.

jrprice:
This adds a deferred image initialization step that is executed
immediately prior to the first device command that uses the image. The
contents of the host pointer are copied to a staging buffer during
image creation, and then copied to the image on the device queue that
is receiving the first command.

This initialization step also performs an image layout transitions,
which fixes #108.


jrprice:
The image initialization step doesn't necessarily need to be deferred - the commands could be submitted during creation with a fence that is checked just before the image is needed. This involved a bit more refactoring though, since you need a device queue and command pool during image creation to do this. I'm happy to do it like this if you'd rather though - some of that refactoring may help with some other things that we'll want to do later anyway.

jrprice:
This is passing all of the `kernel_read_write` image tests for me on NVIDIA when using `CL_FILTER_NEAREST`, for all image types and with `use_pitches`. There's some failures when linear filtering is used, but I think they are unrelated (will look into them separately). There's some failures with arrayed images on other platforms, but I also suspect that they are unrelated.

I think some more work will be needed to get `CL_MEM_USE_HOST_PTR` fully conformant, but it seems to now work for all of the tests that I've run so far (except the fact that mapping the image returns a different pointer).

jrprice:
Looking into the SwiftShader test failures on CI now...

jrprice:
The CI failures are due to a bug in SwiftShader that has [already been fixed](https://github.com/google/swiftshader/commit/e8dd233c7a85f3c689caf06c226a7f8405a480d3#diff-1640eba7bc008e754403a96daaa5094027d130e3d7adbd3fc272bf5ad9441d9b).

kpet:
The CI is using pre-builts for Swiftshader. I've run the pipeline to update them but it seems they're not getting picked up when I retrigger your jobs. You may have to rebase or, worst case, open another PR.

On linear filtering, be aware that the spec doesn't define a required tolerance on results for GPUs. This is currently hard-coded in the CTS. Yet another thing we need to fix.

jrprice:
Thanks, I've rebased and CI is now passing.

callumfare:
Hi @kpet  @jrprice, what's the status of this PR? We (Codeplay) are interested in support for creating images from host pointers and I want to check that these changes are a good starting off point, and that's it's ok for us to pick up work on this. I'm assuming the PR isn't being worked on any more.

Incidentally, could the staging buffer design here be a more robust (although less performant) solution for CL_MEM_OBJECT_IMAGE1D_BUFFER? I noticed there were some comments on #339 that using the vkMemory from the buffer directly might not work depending on the image layout.

kpet:
HI @callumfare This PR is not a good direct starting point for a performant approach to creating images with a host pointer (and correctly performing the initial image layout transition). Those operations need to be handled asynchronously whereas the approach in this PR is fully synchronous and will result in the submission latency being incurred potentially multiple times per individual command. That's the reason the PR has been lingering. I have gone reasonably far with prototyping an implementation that doesn't suffer from this limitation but haven't made those changes public yet. Let's have an email/etc discussion to figure out the best path forward.

On 1D Buffer images, we should avoid using a staging buffer for performance reasons (see more notes on https://github.com/kpet/clvk/pull/389).

kpet:
@jrprice I've opened https://github.com/kpet/clvk/pull/397 that supersedes this PR. I've nicked your tests and a few other bits and pieces and added you as co-author for #397. Closing this PR.

PR #277, opened 2020-11-25
Check if arguments have been set at kernel enqueue

jrprice:
Track arguments that have not yet been set in a hash set, and remove
them once they have been set. If there is anything left in the set
when clEnqueueNDRangeKernel is called, return CL_INVALID_KERNEL_ARGS.

An alternative would be to track which arguments have been set in a `std::vector<bool>` and then use `std::all_of()`. The `std::unordered_set` approach leaves the door open to provide better diagnostics via the context callback or clvk's log (e.g. we can specify exactly which arguments were unset).

I didn't do anything for the "... or if a kernel argument declared to be a pointer to a type does not point to a named address space" part. I'm not actually sure what the spec means here - this sounds to me like something that would be a program build failure, not something that should be caught during kernel enqueue.


kpet:
I'm a bit concerned with the cost of calling `std::unordered_set::erase` for each argument set. I'm not sure I understand why a set allows for better diagnostics compared to a vector. Couldn't you use the argument position to index into the vector?

>I didn't do anything for the [...] this sounds to me like something that would be a program build failure

Agree this should be a compilation error. I think we should fix the spec.

jrprice:
Yeah I'm not sure what I was thinking there. `vector<bool>` works fine, I've pushed an update to do that instead.

I'll open an OpenCL-Docs PR to fix the spec w.r.t. that other issue.

kpet:
Spec fix: https://github.com/KhronosGroup/OpenCL-Docs/pull/542

src/kernel.cpp
```diff
@@ -84,7 +87,14 @@ cl_int cvk_kernel::set_arg(cl_uint index, size_t size, const void* value) {
 
     auto const& arg = m_args[index];
 
-    return m_argument_values->set_arg(arg, size, value);
+    cl_int ret = m_argument_values->set_arg(arg, size, value);
+
+    if (ret == CL_SUCCESS) {
+        // Mark argument as set
+        m_args_set[index] = true;
+    }
```
kpet, line 95: We probably want to set the entry to false in the case of a failed argument setting operation to handle the following scenario:
- all arguments set and kernel enqueued successfully
- one argument changed but failed to be set
- subsequent kernel enqueues should fail but they would succeed with the current code.

Probably worth extending the test with that case as well.

kpet (changes requested):
Thanks! `vector` will reduce the cost across argument setting and kernel enqueue.

src/kernel.cpp
jrprice: This is an interesting scenario. I can't see anywhere in the spec that explicitly describes what should happen here. I've just written the test case (passing `0` to the `arg_size` argument of `clSetKernelArg()` and then trying to enqueue the kernel again), but so far all of the native OpenCL implementations I've tried (Mali, Adreno, NVIDIA) still allow the kernel to be enqueued, despite the fact that they all return the correct error from `clSetKernelArg`.

If you think this is the wrong behavior, maybe this needs a spec change before we land this in clvk?

kpet: Agree we should clarify the spec. I've created https://github.com/KhronosGroup/OpenCL-Docs/issues/548. Would you prefer to merge this change as is and follow up later or wait for the spec issue to be resolved? (I'm fine either way.)

jrprice: There is no urgency to land this PR from my side, happy to wait for resolution on the spec issue.

kpet: @jrprice We're yet to update the spec but I think the agreement we've reached with the WG means this PR could now get in with a small tweak.

jrprice: @kpet I added a test for enqueuing after a failed call to `clSetKernelArg`, which I believe we concluded will be guaranteed to succeed for the type of failure I used. Is this the tweak you meant?

kpet: @jrprice Apologies I missed your reply. I _think_ I meant that we should mark the argument as unset when argument setting failed with a code indicating that state may have been modified. `cvk_kernel::set_arg` will have modified state when returning `CL_OUT_OF_RESOURCES` for example. However, looking at this a bit more there are more fundamental issues with error handling and calling `clSetKernelArg` on a kernel for which a previous `clSetKernelArg` call failed with `CL_OUT_OF_RESOURCES` would result in a crash. Since the issue pre-dates this PR and this PR is a net improvement, I'll merge it as is (it's been a while!). Thanks!

kpet (approved):

PR #278, opened 2020-12-01
Add support for pipeline cache serialization

jrprice:
If the user sets the `CLVK_CACHE_DIR` environment variable to a non-empty value, clvk will write Vulkan pipeline data to a per-device file in that directory during teardown. On the next run, clvk will load pipeline cache data from this directory and use it to seed the initial pipeline cache.

This also changes the Vulkan pipeline cache to be device-scope, rather than program scope. We could keep using per-program pipeline caches if we serialized them all separately (during program teardown), but we'd need to hash the SPIR-V to uniquely identify them. I didn't want to go this route just yet since it would probably add an additional dependency for the hash function (unless we rolled our own...).

This provides significant performance improvements (1.5-1.8X) for part of a workload that I'm looking at on both Adreno and Mali GPUs. 

I left the environment variable name generic as we may also want to cache the Clspv output at some point as well to improve OpenCL-C compilation speed. The particular workload I'm looking manually caches program binaries and uses `clCreateProgramFromBinary`, which is why the Vulkan pipeline creation cost is showing up as an issue.

We haven't talked about any of this before so I'm happy to discuss here or elsewhere before review if you want to.

### DIFF
--- a/src/device.cpp
+++ b/src/device.cpp
@@ -528,15 +528,16 @@ get_pipeline_cache_filename(VkPhysicalDeviceProperties properties) {
     }
 
     // The pipeline cache file path is:
-    // ${CLVK_CACHE_DIR}/<UUID>.clvk-pipeline-cache.bin
+    // ${CLVK_CACHE_DIR}/clvk-pipeline-cache.<UUID>.bin
     std::stringstream uuid_ss;
     for (int i = 0; i < VK_UUID_SIZE; i++) {
         uuid_ss << std::hex << (int)properties.pipelineCacheUUID[i];
     }
     std::string cache_path = cache_dir;
     cache_path += "/";
+    cache_path += "clvk-pipeline-cache.";
     cache_path += uuid_ss.str();
-    cache_path += ".clvk-pipeline-cache.bin";
+    cache_path += ".bin";
     return cache_path;
 }
 
--- a/src/device.cpp
+++ b/src/device.cpp
@@ -520,7 +520,8 @@ bool cvk_device::init_time_management(VkInstance instance) {
 
 // Returns the pipeline cache file path for a given Vulkan implementation.
 // If pipeline cache serialization is not enabled, an empty string is returned.
-std::string get_pipeline_cache_filename(VkPhysicalDeviceProperties properties) {
+static std::string
+get_pipeline_cache_filename(VkPhysicalDeviceProperties properties) {
     const char* cache_dir = getenv("CLVK_CACHE_DIR");
     if (cache_dir == nullptr) {
         return "";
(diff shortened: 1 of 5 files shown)


src/device.cpp
```diff
@@ -515,6 +518,110 @@ bool cvk_device::init_time_management(VkInstance instance) {
     return true;
 }
 
+// Returns the pipeline cache file path for a given Vulkan implementation.
+// If pipeline cache serialization is not enabled, an empty string is returned.
+std::string get_pipeline_cache_filename(VkPhysicalDeviceProperties properties) {
```
kpet, line 523: Could be `static`

src/device.cpp
```diff
@@ -515,6 +518,110 @@ bool cvk_device::init_time_management(VkInstance instance) {
...
+    std::string cache_path = get_pipeline_cache_filename(m_properties);
+    if (cache_path.empty()) {
+        return;
+    }
+
+    // Retrieve the pipeline cache data from the Vulkan implementation
+    size_t size;
+    res = vkGetPipelineCacheData(m_dev, m_pipeline_cache, &size, nullptr);
+    if (res != VK_SUCCESS) {
+        cvk_warn("Failed to retrieve pipeline cache size");
```
kpet, line 599: This should probably be an error message (or the error messages below should be warnings).

src/device.cpp
```diff
@@ -515,6 +518,110 @@ bool cvk_device::init_time_management(VkInstance instance) {
...
+        return;
+    }
+    std::vector<char> cache_data(size);
+    res = vkGetPipelineCacheData(m_dev, m_pipeline_cache, &size,
+                                 cache_data.data());
+    if (res != VK_SUCCESS) {
+        cvk_warn("failed to retrieve pipeline cache data");
```
kpet, line 606: Same here

src/device.cpp
```diff
@@ -515,6 +518,110 @@ bool cvk_device::init_time_management(VkInstance instance) {
...
+// Returns the pipeline cache file path for a given Vulkan implementation.
+// If pipeline cache serialization is not enabled, an empty string is returned.
+std::string get_pipeline_cache_filename(VkPhysicalDeviceProperties properties) {
+    const char* cache_dir = getenv("CLVK_CACHE_DIR");
+    if (cache_dir == nullptr) {
+        return "";
+    }
+
+    // The pipeline cache file path is:
+    // ${CLVK_CACHE_DIR}/<UUID>.clvk-pipeline-cache.bin
```
kpet, line 530: Minor detail: maybe the fixed part could come before the UUID so all pipeline cache files are grouped together when listing the content of the cache directory (i.e. `clvk-pipeline-cache.<UUID>.bin`), unless you were thinking of reusing the UUID to prefix cached SPIR-V files as well. I don't really mind either way.

kpet (changes requested):
Looks great! I like the approach, the move to per-device pipeline caches and the ability to cache SPIR-V files in the future. Just a couple of minor touches please.

jrprice:
Thanks for the review. Feedback addressed, PTAL.

kpet (approved):
Thanks!

PR #280, opened 2020-12-09
Enable cl_khr_fp16 if the device can support it

jrprice:
This requires 16-bit storage and shaderFloat16.

### DIFF
--- a/src/device.cpp
+++ b/src/device.cpp
@@ -374,6 +374,14 @@ void cvk_device::build_extension_ils_list() {
             MAKE_NAME_VERSION(1, 0, 0, "cl_khr_device_uuid"));
     }
 
+    // Enable cl_khr_fp16 if we have 16-bit storage and shaderFloat16
+    if (is_vulkan_extension_enabled(VK_KHR_16BIT_STORAGE_EXTENSION_NAME) &&
+        is_vulkan_extension_enabled(
+            VK_KHR_SHADER_FLOAT16_INT8_EXTENSION_NAME) &&
+        m_features_float16_int8.shaderFloat16) {
+        m_extensions.push_back(MAKE_NAME_VERSION(1, 0, 0, "cl_khr_fp16"));
+    }
+
     // Build extension string
     for (auto& ext : m_extensions) {
         m_extension_string += ext.name;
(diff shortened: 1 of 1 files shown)


src/device.cpp
```diff
@@ -374,6 +374,14 @@ void cvk_device::build_extension_ils_list() {
             MAKE_NAME_VERSION(1, 0, 0, "cl_khr_device_uuid"));
     }
 
+    // Enable cl_khr_fp16 if we have 16-bit storage and shaderFloat16
+    if (is_vulkan_extension_enabled(VK_KHR_16BIT_STORAGE_EXTENSION_NAME) &&
+        is_vulkan_extension_enabled(
+            VK_KHR_SHADER_FLOAT16_INT8_EXTENSION_NAME) &&
```
kpet, line 380: These extensions have become core in 1.2. We need to handle that case as well.

src/device.cpp
```diff
@@ -374,6 +374,14 @@ void cvk_device::build_extension_ils_list() {
...
+        m_features_float16_int8.shaderFloat16) {
```
kpet, line 381: I think we should check for at least `m_features_16bit_storage.storageBuffer16BitAccess` and probably `m_features_16bit_storage.uniformAndStorageBuffer16BitAccess` and/or  `m_features_16bit_storage.storagePushConstant16` for POD argument buffers and POD arguments in push constants.

kpet (changes requested):

src/device.cpp
jrprice: I've added a check for `storageBuffer16BitAccess`.

I don't think checking for uniforms and push constants is necessary. clvk already tells Clspv whether or not these things are supported, and Clspv should automatically fallback to storage buffers for things like POD arguments if necessary.

src/device.cpp
jrprice: Can you clarify what needs to be done here? My impression (maybe incorrect?) was that these extensions would still be advertised even on newer Vulkan versions where they are core. I think the existing code for enabling extension during device initialization also makes this assumption; we only enable 16-bit storage features if the extension is present for example.

kpet: I could not find a definitive source but I'm pretty sure advertising the extensions is not a requirement. As you say, there are other cases so we might as well deal with all of them with one later change.

src/device.cpp
kpet: Yes, with Alan's recent-ish changes, uniforms and push constants will be selected only when supported. I agree that checking for storage buffer support is enough.

kpet (approved):
LGTM, thanks!

PR #282, opened 2020-12-22
Use per-module Vulkan pipeline caches

jrprice:
Compute the SHA-1 hash of the SPIR-V module and use that as a key for both in-memory and serialized caches of Vulkan pipelines caches.

Perform these cache lookups during program build. If there was a cache hit, we know that the SPIR-V has previously been validated, and so we can skip validation this time round. This significantly improves `clBuildProgram()` performance when creating programs from binaries.

I implemented the SHA-1 hashing from scratch, which avoids adding extra dependencies or the potential license complexities of incorporating third-party code into clvk directly. I've tested it against some of the test vectors available from NIST, but as I'm writing this I realize that I should probably add some unit tests for it here too.

The `cvk_device::get_pipeline_cache()` interface feels a little clunky; the `bool` result is used by the caller to avoid SPIR-V validation. The `bool` could be made an output parameter instead, but this didn't feel much better to me.


jrprice:
I've added a bunch of tests for the SHA-1 implementation now, taken from a set of tests provided by NIST. Good thing too - there was a few bugs in my implementation that are now fixed.

kpet (changes requested):
I haven't reviewed the details of the SHA1 implementation (and probably will trust you and the testing). Overall looks good but now needs rebasing.

jrprice:
Addressed review comment and merged with master, PTAL.

src/sha1.cpp
```diff
@@ -199,15 +201,15 @@ cvk_sha1_hash cvk_sha1(const uint8_t* data, uint32_t length) {
     // Process data in 512-bit blocks.
     while (length >= 64) {
         // Copy block to local buffer and process it.
-        memcpy(block, data, 64);
+        memcpy(block, data_i8, 64);
         sha1_process_block(H, block);
 
         length -= 64;
-        data += 64;
+        data_i8 += 64;
```
kpet, line 208: I hadn't spotted the pointer arithmetic here. Probably still better to have the cast in the implementation.

kpet (approved):
LGTM!

PR #283, opened 2020-12-29
Handle reqd_work_group_size attribute

### LINKED ISSUE #50: Pass api:kernel_required_group_size conformance test
Testing CL_KERNEL_WORK_GROUP_SIZE query.

Can be done in clvk by inspecting the SPIR-V module using SPIR-V tools (prototyped this) but probably better done in clspv with a suitable interface to clvk.

jrprice:
* Parse the PropertyRequiredWorkgroupSize extended instruction and
  store the required work-group size

* Implement the CL_KERNEL_COMPILE_WORK_GROUP_SIZE query

* Check the requested work-group size against the required work-group
  size during kernel enqueue

Closes #50.

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -2689,8 +2689,11 @@ cl_int CLVK_API_CALL clGetKernelWorkGroupInfo(
         copy_ptr = &val_ulong;
         ret_size = sizeof(val_ulong);
         break;
+    case CL_KERNEL_COMPILE_WORK_GROUP_SIZE:
+        copy_ptr = kernel->required_work_group_size().data();
+        ret_size = sizeof(size_t[3]);
+        break;
     case CL_KERNEL_GLOBAL_WORK_SIZE:        // TODO
-    case CL_KERNEL_COMPILE_WORK_GROUP_SIZE: // TODO
     case CL_KERNEL_PRIVATE_MEM_SIZE:        // TODO
     default:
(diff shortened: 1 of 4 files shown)


src/api.cpp
```diff
@@ -2689,8 +2689,11 @@ cl_int CLVK_API_CALL clGetKernelWorkGroupInfo(
         copy_ptr = &val_ulong;
         ret_size = sizeof(val_ulong);
         break;
+    case CL_KERNEL_COMPILE_WORK_GROUP_SIZE:
+        copy_ptr = kernel->required_work_group_size().data();
+        ret_size = sizeof(size_t[3]);
```
kpet, line 2694: In most (all?) other places, size calculations use `size()` on the container * sizeof(type). I'm not saying one is better than the other, just pointing out the very minor style divergence.

kpet (approved):
LGTM, thanks! I believe this will work for all cases. clspv now also supports the case where there are kernels with and without the attribute in the same program (`3b6097783`) and the generation of the non-semantic instructions is tested (`test/Reflection/mixed_reqd_workgroup_sizes.cl`). This just leaves the [documentation](https://github.com/google/clspv/blob/master/docs/OpenCLCOnVulkan.md#work-group-size) to update. I'll open a PR.

kpet:
clspv doc update: https://github.com/google/clspv/pull/693

PR #284, opened 2020-12-29
Fix clEnqueueWriteImage with offsets

jrprice:
Fix clEnqueueWriteImage with offsets

* Do not apply `origin` as an offset when copying from the host
  pointer to the staging buffer.
    
* Use the region size instead of the image size when calculating
  default pitches.
    
Added a test for this.

### DIFF
--- a/tests/api/images.cpp
+++ b/tests/api/images.cpp
@@ -302,3 +302,105 @@ TEST_F(WithCommandQueue,
     EnqueueUnmapMemObject(image, data);
     Finish();
 }
+
+TEST_F(WithCommandQueue, ImageWriteOffset) {
+    // Create a 2D image.
+    const size_t IMAGE_WIDTH = 16;
+    const size_t IMAGE_HEIGHT = 16;
+
+    // This is the full image (used for initial fill and final read back).
+    size_t full_origin[3] = {0, 0, 0};
+    size_t full_region[3] = {IMAGE_WIDTH, IMAGE_HEIGHT, 1};
+
+    // We are going to write to this region.
+    size_t write_origin[3] = {IMAGE_WIDTH / 4, IMAGE_HEIGHT / 4, 0};
+    size_t write_region[3] = {IMAGE_WIDTH / 2, IMAGE_HEIGHT / 2, 1};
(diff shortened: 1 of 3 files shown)


jrprice:
Just noticed that this failed CI.

I've reproduced this locally, and the crash seems to be an existing bug unrelated to this change (the new test still crashes without this change). It's crashing somewhere around internal memory mapping/unmapping routines, but only seems to happen with SwiftShader. I'll investigate next week when I'm back.

jrprice:
I've believe I've fixed this now. There was another issue with pitches when writing to a subset of the image that meant we were writing beyond the bounds of the staging buffer. I've made the test better.

CI is still failing, but that looks related to the build cache and maybe a SPIRV-LLVM-Translator issue? The failures look the same as the ones on master anyhow.

tests/api/images.cpp
```diff
@@ -302,3 +302,105 @@ TEST_F(WithCommandQueue,
...
+    // This is the full image (used for initial fill and final read back).
+    size_t full_origin[3] = {0, 0, 0};
+    size_t full_region[3] = {IMAGE_WIDTH, IMAGE_HEIGHT, 1};
+
+    // We are going to write to this region.
+    size_t write_origin[3] = {IMAGE_WIDTH / 4, IMAGE_HEIGHT / 4, 0};
+    size_t write_region[3] = {IMAGE_WIDTH / 2, IMAGE_HEIGHT / 2, 1};
+
+    cl_uchar fill_value = 0xFF;
+    std::vector<cl_uchar> write_data(IMAGE_WIDTH * IMAGE_HEIGHT, 0);
```
kpet, line 320: Suggest reducing the size of `write_data` to the size of `write_region` (and adapting the init loop nest below) to make it clear that the other data is never accessed.

kpet (changes requested):
The fix looks good, thanks! One comment on the test.

tests/api/images.cpp
jrprice: Good idea, done.

kpet (approved):

PR #285, opened 2020-12-29
Initialize memory allocation pointer

jrprice:
Prevents invalid calls to vkFreeMemory upon destruction for memory
allocations that failed.

This was causing tests in allocations CTS suite to crash on some platforms.


kpet (approved):
Oh, yes. LGTM, thanks!

PR #289, opened 2021-01-13
Fix float->int conversion UB in clEnqueueFillImage

jrprice:
The behavior of float->int conversion when the value is out-of-range or NaN is undefined, so manually handle NaN and clamp the value first.

With this change (and #276), the full `clFillImage` CTS suite passes on NVIDIA for me.

### DIFF
--- a/src/memory.cpp
+++ b/src/memory.cpp
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <cmath>
+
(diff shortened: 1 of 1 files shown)


kpet (approved):
This matches the behaviour described in 8.3.1.2 and 6.4.3.3 except for the absence of guarantee on the rounding mode used. LGTM as is since it fixes obvious issues and is enough for the conformance tests. Thanks!

PR #290, opened 2021-02-10
Add missing icd.hpp include in objects.hpp

jrprice:
Without this, `_cl_context` is not a complete type, which causes build failures since it is used as a base class in this file. The `icd.hpp` include was previously happening indirectly via `device.hpp`, but that include was removed in 3961358.


kpet (approved):
LGTM, thanks!

PR #300, opened 2021-03-26
Control SPIR-V validation through env variable

mantognini:
Let the user skip validation or ignore errors using the CLVK_SPIRV_VALIDATION environment variable.


kpet (approved):
LGTM. Thanks for this!

PR #333, opened 2021-11-30
Fix two bugs from cvk_ndrange commit

jrprice:
There were two small bugs:
* Single task enqueue was launching (1, 0, 0) instead of (1, 1, 1)
* The 3rd dimension of CL_KERNEL_COMPILE_WORK_GROUP_SIZE was incorrect


kpet (approved):
No such thing as a trivial change :). Thanks for the fixes!

PR #336, opened 2022-01-07
fix for local buffer argument in kernel prototype

rjodinchr:
In OpenCL CTS 'vstore_local', we have this pattern:
```
for (args - last_arg) {
  clSetKernelArg(arg)
}
for (various alignmentOffset){
  clSetKernelArg(last_arg, alignmentOffset);
  clEnqueueNDRangeKernel();
}
```
When we clSetKernelArg the second time. As the kernel has been enqueue
once, we will create a new cvk_kernel_argument_values in
cvk_kernel::set_arg. And we will lose information about the local
buffer argument, which led to a local buffer with the wrong size.

Let's copy the needed information in order to ensure that the local
buffer will always have the proper size.

### DIFF
--- a/tests/api/local_buffer.cpp
+++ b/tests/api/local_buffer.cpp
@@ -1,0 +1,65 @@
+// Copyright 2022 The clvk authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "testcl.hpp"
+
+static bool check(cl_int* buf, size_t size, size_t limit) {
+    for (size_t i = 0; i < size; i++) {
+        if ((i >= limit && buf[i] != 0xffffffff) ||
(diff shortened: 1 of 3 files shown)


kpet (approved):
Good catch! I'm wondering whether we should add a unit test to `api_tests` or whether we can rely on the CTS. It's unlikely to change but, if it does, we could lose coverage of that case since that's not what the CTS test was designed to exercise.

rjodinchr:
I will do that

tests/api/local_buffer.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+    for (size_t limit : {64, 96, 128}) {
+        SetKernelArg(kernel, 2, sizeof(cl_int), &limit);
+        cl_int dst_buf[nb_wis];
+        memset(dst_buf, 0xff, buffer_size);
+
+        EnqueueWriteBuffer(dst, true, 0, buffer_size, dst_buf);
+
+        EnqueueNDRangeKernel(kernel, 1, nullptr, &nb_wis, &nb_wis);
+        Finish();
```
kpet, line 59: The call to `clFinish` is not required since the following buffer read is blocking.

tests/api/local_buffer.cpp
```diff
@@ -0,0 +1,65 @@
...
+    for (size_t i = 0; i < size; i++) {
+        if ((i >= limit && buf[i] != 0xffffffff) ||
+            (i < limit && (buf[i] != (limit - 1 - i)))) {
+            return false;
+        }
+    }
+    return true;
+}
+
+TEST_F(WithCommandQueue, LocalBuffer) {
```
kpet, line 27: A name a bit more descriptive than `LocalBuffer` would have been nice but one can always use the git history to get the info.

kpet (approved):
Thanks!

PR #339, opened 2022-02-04
add support for CL_MEM_OBJECT_IMAGE1D_BUFFER

rjodinchr:
cl_image use the vkMemory from the cl_buffer.
it only works if the device host visible memory is also device local
memory so that the memory can be bind by vulkan to the vkImage.

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -1837,7 +1837,7 @@ cl_int CLVK_API_CALL clGetMemObjectInfo(cl_mem mem, cl_mem_info param_name,
         break;
     case CL_MEM_ASSOCIATED_MEMOBJECT:
         if (memobj->is_image_type()) {
-            auto img = static_cast<cvk_image*>(mem);
+            auto img = static_cast<cvk_image*>(memobj);
             val_memobj = img->buffer();
         } else {
             val_memobj = memobj->parent();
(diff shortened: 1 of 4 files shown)


src/memory.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include <cmath>
 
+#include "CL/cl.h"
```
kpet, line 17: This is already pulled via `memory.hpp`. The OpenCL headers are included via `cl_headers.hpp`.

src/api.cpp
```diff
@@ -1836,7 +1836,12 @@ cl_int CLVK_API_CALL clGetMemObjectInfo(cl_mem mem, cl_mem_info param_name,
         ret_size = sizeof(val_uint);
         break;
     case CL_MEM_ASSOCIATED_MEMOBJECT:
-        val_memobj = memobj->parent();
+        if (memobj->is_image_type()) {
+            auto img = static_cast<cvk_image*>(mem);
```
kpet, line 1840: Might be better to cast from `memobj` for readability (and to leave dealing with external types to `icd_downcast`).

kpet (changes requested):
Thanks! I've just done a first pass on this. The behaviour of images with a `Buffer` dimension is poorly specified. I'll need to spend more time reading the specs (or getting it clarified) to check that this is correct. Going by memory aliasing rules alone, this is not guaranteed to work as clvk always creates optimally-tiled images that are not linear resources. I can see how it's very likely to work for 1D images but I can't find anything in the specs that allows this.

(I might accept the PR anyway and revisit when looking at https://github.com/kpet/clvk/issues/43)

src/memory.cpp
rjodinchr: Sorry, I think it has been added automatically by clangd and I forgot to remove it

kpet (approved):
This will need a bit more effort to guarantee it always works (spec and/or implementation). Merging in the interest of forward progress.

PR #340, opened 2022-02-05
Support CL_INTENSITY image format

matze:
Without this, `clCreateImage()` returns `CL_OUT_OF_RESOURCES`. The difference between `CL_LUMINANCE` and `CL_INTENSITY` lies in the fourth component, when for example reading into a `float4`. Unfortunately, I cannot find the place where this is handled (if at all?).

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -4243,6 +4243,20 @@ std::unordered_map<cl_image_format, VkFormat, ClFormatMapHash, ClFormatMapEqual>
         {{CL_LUMINANCE, CL_SIGNED_INT32}, VK_FORMAT_R32_SINT},
         {{CL_LUMINANCE, CL_FLOAT}, VK_FORMAT_R32_SFLOAT},
 
+        // INTENSITY formats
+        {{CL_INTENSITY, CL_UNORM_INT8}, VK_FORMAT_R8_UNORM},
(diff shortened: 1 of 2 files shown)


src/api.cpp
```diff
@@ -4304,7 +4318,8 @@ bool cl_image_format_to_vulkan_format(const cl_image_format& clformat,
         success = true;
     }
 
-    if (clformat.image_channel_order == CL_LUMINANCE) {
+    if (clformat.image_channel_order == CL_LUMINANCE ||
+        clformat.image_channel_order == CL_INTENSITY) {
         *components = {VK_COMPONENT_SWIZZLE_R, VK_COMPONENT_SWIZZLE_R,
                        VK_COMPONENT_SWIZZLE_R, VK_COMPONENT_SWIZZLE_A};
```
kpet, line 4324: This is where you should handle the component ordering. You'll want to use 4 times `VK_COMPONENT_SWIZZLE_R` for `CL_INTENSITY` so the R component of the format you've mapped it on is replicated.

kpet (changes requested):
Thanks for the contribution! Just needs one small fix to the component selection logic (see other comment).

kpet (approved):
Thanks, all passing the CTS.

Issue #45, opened 2019-04-12
Support cl_khr_il_program


kpet:
SPIR-V via the SPIRV-LLVM translator.

- [x] Support using offline tools
- [ ] Support in online compilation mode

PR #341, opened 2022-02-07
Add test that loads program from IL Binary

RSenApps:
Tests using clCreateProgramWithILKHR extension when
CLVK_CLSPV_ONLINE_COMPILER is disabled. Added simple-from-il-binary
test is based off of the existing simple-from-binary test.

Contributes to #45.

@DCastagna @jrprice

### DIFF
--- a/tests/azure/steps-run-tests.yml
+++ b/tests/azure/steps-run-tests.yml
@@ -10,6 +10,8 @@ parameters:
     type: string
   - name: compilerAvailable
     type: boolean
+  - name: onlineCompiler
+    type: boolean
 
 steps:
   - download: swiftshader
@@ -43,7 +45,8 @@ steps:
         export CLVK_LOG=3
         ./build/simple_test_from_il_binary
         ./build/simple_test_from_il_binary_static
-      displayName: Offline IR compilation simple tests
+      displayName: Offline IL compilation simple tests
+      condition: and(${{ parameters.compilerAvailable }}, not(${{ parameters.onlineCompiler }}))
   - ${{ if eq(parameters.os, 'windows') }}:
     - script: |
         cp build/src/Release/OpenCL.dll build/Release
--- a/tests/azure/steps-run-tests.yml
+++ b/tests/azure/steps-run-tests.yml
@@ -37,7 +37,7 @@ steps:
       displayName: Offline compilation simple tests
     - script: |
         set -ex
-        clang -c -target spir -O0 -emit-llvm -o simple-cl.bc ./tests/simple-from-il-binary/simple.cl
+        clang  -Xclang -finclude-default-header -c -target spir -O0 -emit-llvm -o simple-cl.bc ./tests/simple-from-il-binary/simple.cl
         ./build/llvm-spirv simple-cl.bc -o simple-cl.spv
         export VK_ICD_FILENAMES=${{ parameters.icd_json_unix }}
         export CLVK_LOG=3
--- a/tests/azure/steps-run-tests.yml
+++ b/tests/azure/steps-run-tests.yml
@@ -39,7 +39,7 @@ steps:
       displayName: Offline compilation simple tests
     - script: |
         set -ex
-        clang  -Xclang -finclude-default-header -c -target spir -O0 -emit-llvm -o simple-cl.bc ./tests/simple-from-il-binary/simple.cl
+        ./build/external/clspv/third_party/llvm/bin/clang  -Xclang -finclude-default-header -c -target spir -O0 -emit-llvm -o simple-cl.bc ./tests/simple-from-il-binary/simple.cl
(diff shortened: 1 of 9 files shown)


tests/simple-from-il-binary/simple.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+    // Create program
+    size_t program_size = size;
+    const unsigned char* program_code =
+        reinterpret_cast<const unsigned char*>(code.data());
+
+    clCreateProgramWithILKHR_fn clCreateProgramWithILKHR = NULL;
+    clCreateProgramWithILKHR =
+        (clCreateProgramWithILKHR_fn)clGetExtensionFunctionAddressForPlatform(
+            platform, "clCreateProgramWithILKHR");
```
DCastagna, line 97: Does this test the extension is present in the extension string?  Please do keep in mind the name of the extension is "cl_khr_il_program".

RSenApps: Thanks, I added a check for this.

tests/azure/steps-run-tests.yml
```diff
@@ -35,6 +37,16 @@ steps:
         ./build/simple_test_from_binary
         ./build/simple_test_from_binary_static
       displayName: Offline compilation simple tests
+    - script: |
+        set -ex
+        clang  -Xclang -finclude-default-header -c -target spir -O0 -emit-llvm -o simple-cl.bc ./tests/simple-from-il-binary/simple.cl
```
kpet, line 42: I'm wondering whether this should build and use the clang coming from the LLVM tree that clspv uses to avoid spurious CI failures because of incompatibilities between whatever version of clang is available on the CI runners and the ToT version of llvm-spirv. Unclear to me how much of a practical problem this could be.

tests/simple-from-il-binary/simple.cpp
```diff
@@ -0,0 +1,200 @@
...
+    CHECK_CL_ERRCODE(err);
+
+    printf("Device: %s\n", device_name);
+
+    size_t extensions_length;
+    err = clGetDeviceInfo(device, CL_DEVICE_EXTENSIONS, 0, nullptr,
+                          &extensions_length);
+    CHECK_CL_ERRCODE(err);
+
+    char* extensions = (char*)calloc(extensions_length + 1, sizeof(char));
```
kpet, line 67: Could use `std::vector` or `std::string` maybe.

tests/simple-from-il-binary/simple.cpp
```diff
@@ -0,0 +1,200 @@
...
+    err =
+        clGetDeviceInfo(device, CL_DEVICE_EXTENSIONS,
+                        sizeof(char) * extensions_length, extensions, nullptr);
+    CHECK_CL_ERRCODE(err);
+
+    if (extensions[extensions_length] != '\0') {
```
kpet, line 73: I don't think a simple example should (have to) check this but I understand why you might have done it. The spec doesn't say anything about the string being NUL-terminated but, IIRC, this is guaranteed by the CTS. We should probably get that clarified in the specification.

tests/simple-from-il-binary/simple.cpp
```diff
@@ -0,0 +1,200 @@
...
+    size_t program_size = size;
+    const unsigned char* program_code =
+        reinterpret_cast<const unsigned char*>(code.data());
+
+    clCreateProgramWithILKHR_fn clCreateProgramWithILKHR = NULL;
+    clCreateProgramWithILKHR =
+        (clCreateProgramWithILKHR_fn)clGetExtensionFunctionAddressForPlatform(
+            platform, "clCreateProgramWithILKHR");
+    if (clCreateProgramWithILKHR == NULL) {
+        fprintf(stderr, "Failed to find extension clCreateProgramWithILKHR\n");
```
kpet, line 122: nit: s/extension/extension function/

tests/simple-from-il-binary/simple.cpp
```diff
@@ -0,0 +1,200 @@
...
+    ifile.read(reinterpret_cast<char*>(code.data()), size);
+    if (!ifile.good()) {
+        fprintf(stderr, "Failed to read SPIR-V module from %s\n", fname);
+        return EXIT_FAILURE;
+    }
+    printf("Loaded SPIR-V data from %s, size = %u\n", fname, size);
+
+    // Create program
+    size_t program_size = size;
+    const unsigned char* program_code =
```
kpet, line 114: Could use `auto` to avoid spelling out the type twice.

kpet (commented):
Thanks for the contribution! One real question around the version of clang we use in the new CI tests; other comments are minor.

RSenApps:
Thank you for the feedback, I agreed with all your points and added the changes to the PR.

kpet (approved):
All looking good, thanks!

PR #342, opened 2022-02-09
lock call to clspv to be thread-safe

rjodinchr:
clspv is based on LLVM.

LLVM options parsing is done using global variable that are not thread safe.

Thus, we need to lock call to clspv in order to ensure a thread safe execution.


kpet (approved):
LGTM, thanks!

PR #343, opened 2022-02-11
Support IL binaries in online compilation mode

### LINKED ISSUE #45: Support cl_khr_il_program

RSenApps:
Adds support for the cl_khr_il_program extension in online
compilation mode by using the readSpirv api from
SPIRV-LLVM-Translator and clspv::Compile.

Closes #45.

@DCastagna @jrprice

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -86,7 +86,10 @@ endif()
 target_include_directories(OpenCL-objects PRIVATE
   "${CLSPV_SOURCE_DIR}/include")
 if (CLVK_CLSPV_ONLINE_COMPILER)
-  set(OpenCL-dependencies ${OpenCL-dependencies} clspv_core)
+  set(OpenCL-dependencies ${OpenCL-dependencies} clspv_core LLVMSPIRVLib)
+  add_dependencies(OpenCL-objects intrinsics_gen)
+  target_include_directories(OpenCL-objects PRIVATE
+    "${LLVM_SPIRV_SOURCE}/include")
 else()
   if (CLVK_COMPILER_AVAILABLE)
     add_dependencies(OpenCL-objects clspv)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -87,7 +87,6 @@ target_include_directories(OpenCL-objects PRIVATE
   "${CLSPV_SOURCE_DIR}/include")
 if (CLVK_CLSPV_ONLINE_COMPILER)
   set(OpenCL-dependencies ${OpenCL-dependencies} clspv_core LLVMSPIRVLib)
-  add_dependencies(OpenCL-objects intrinsics_gen)
   target_include_directories(OpenCL-objects PRIVATE
     "${LLVM_SPIRV_SOURCE}/include")
 else()
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -93,14 +93,19 @@ if (CLVK_CLSPV_ONLINE_COMPILER)
 else()
   if (CLVK_COMPILER_AVAILABLE)
     add_dependencies(OpenCL-objects clspv)
-    add_dependencies(OpenCL-objects llvm-spirv)
-    set_target_properties(llvm-spirv PROPERTIES RUNTIME_OUTPUT_DIRECTORY
-      ${CMAKE_BINARY_DIR})
     target_compile_definitions(OpenCL-objects PRIVATE
-      DEFAULT_LLVMSPIRV_BINARY_PATH="$<TARGET_FILE:llvm-spirv>")
-    target_compile_definitions(OpenCL-objects PRIVATE
-      DEFAULT_CLSPV_BINARY_PATH="$<TARGET_FILE:clspv>")
+    DEFAULT_CLSPV_BINARY_PATH="$<TARGET_FILE:clspv>")
   endif()
+endif()
+
+# llvm-spirv needed to test simple_test_from_il_binary in CI even when
+# not used in online compilation mode
+if (CLVK_COMPILER_AVAILABLE)
+  add_dependencies(OpenCL-objects llvm-spirv)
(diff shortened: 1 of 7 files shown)


src/CMakeLists.txt
```diff
@@ -86,7 +86,10 @@ endif()
 target_include_directories(OpenCL-objects PRIVATE
   "${CLSPV_SOURCE_DIR}/include")
 if (CLVK_CLSPV_ONLINE_COMPILER)
-  set(OpenCL-dependencies ${OpenCL-dependencies} clspv_core)
+  set(OpenCL-dependencies ${OpenCL-dependencies} clspv_core LLVMSPIRVLib)
+  add_dependencies(OpenCL-objects intrinsics_gen)
```
DCastagna, line 90: What is this for?

src/program.cpp
```diff
@@ -561,6 +569,7 @@ bool save_string_to_file(const std::string& fname, const std::string& text) {
     return ofile.good();
 }
 
+#ifndef CLSPV_ONLINE_COMPILER
```
DCastagna, line 572: Why do you care about stripping out a function if it's not used? Did you get a warning?

src/program.cpp
```diff
@@ -805,22 +820,32 @@ cl_build_status cvk_program::compile_source(const cvk_device* device) {
...
+        membuf m_il_buf(m_il_start, m_il_start + m_il.size());
+        std::istream m_il_stream(&m_il_buf);
+        // llvm-spirv is based on LLVM. LLVM options parsing is done using
+        // global variable that are not thread safe. Thus, we need to lock call
+        // to llvm-spirv in order to ensure a thread safe execution.
+        static std::mutex llvmspirv_compile_lock;
+        llvmspirv_compile_lock.lock();
+        if (!llvm::readSpirv(Context, m_il_stream, M, Err)) {
+            cvk_error_fn("Fails to load SPIR-V as LLVM Module: %s", Err);
+            return CL_BUILD_ERROR;
```
DCastagna, line 840: What happens if we hit this to the std::mutex?  Is it released in the dtor?

src/program.cpp
```diff
@@ -805,22 +820,32 @@ cl_build_status cvk_program::compile_source(const cvk_device* device) {
...
-    // clspv is based on LLVM. LLVM options parsing is done using global
-    // variable that are not thread safe. Thus, we need to lock call to clspv in
-    // order to ensure a thread safe execution.
-    static std::mutex clspv_compile_lock;
-    clspv_compile_lock.lock();
-    auto result = clspv::CompileFromSourceString(
-        m_source, "", build_options, m_binary.raw_binary(), &m_build_log);
-    clspv_compile_lock.unlock();
+    if (build_from_il) {
+        llvm::LLVMContext Context;
```
DCastagna, line 826: Looks like you are using llvm naming convention a bit around here, can we use clvk naming convention instead?

src/program.cpp
```diff
@@ -842,41 +867,75 @@ cl_build_status cvk_program::compile_source(const cvk_device* device) {
             return CL_BUILD_ERROR;
         }
     }
+#endif // CLSPV_ONLINE_COMPILER
 
-    // Compose clspv command-line
-    std::string cmd{gCLSPVPath};
+    bool binaryLoaded = false;
```
DCastagna, line 873: Same, naming convention

src/program.cpp
```diff
@@ -842,41 +867,75 @@ cl_build_status cvk_program::compile_source(const cvk_device* device) {
...
+    // clspv is based on LLVM. LLVM options parsing is done using global
+    // variable that are not thread safe. Thus, we need to lock call to clspv in
+    // order to ensure a thread safe execution.
+    static std::mutex clspv_compile_lock;
+    clspv_compile_lock.lock();
     if (build_from_il) {
-        cmd += " -x ir ";
+        llvm::BumpPtrAllocator A;
+        llvm::StringSaver Saver(A);
+        llvm::SmallVector<const char*, 20> argv;
```
DCastagna, line 894: Can you pass a plain old "const char *const argv" to clspv::Compile?

src/CMakeLists.txt
RSenApps: I was previously getting errors similar to https://reviews.llvm.org/D64357, but it looks like after rebasing on top of https://github.com/kpet/clvk/pull/344 which updates to LLVM 15.0.0git this is no longer needed so I will remove this change.

src/program.cpp
RSenApps: Yes I was getting a warning and also I was attempting to match the similar approach used for the save_string_to_file method above.

src/program.cpp
RSenApps: Thanks I will fix this to be consistent

src/program.cpp
RSenApps: Ah good catch, I forgot to fix the lock when adding the error check. I'll wrap both uses of mutex with std::lock_guard to prevent future bugs from being introduced as well...

src/program.cpp
RSenApps: clspv::Compile expects an already tokenized const char *const argv[]. https://github.com/google/clspv/blob/14378ce9e67d6727ce24578ef98811b6395fc5df/lib/Compiler.cpp#L942

I wanted to try to be consistent with how the command line options are built and tokenized to try to ensure a uniform behavior when running in different configurations so I followed the same approach used by clspv::CompileFromSourceString https://github.com/google/clspv/blob/14378ce9e67d6727ce24578ef98811b6395fc5df/lib/Compiler.cpp#L1075

DCastagna:
LGTM. Kevin, what do you think?

kpet (approved):
Sorry for the delay, I was away for a few days. Thanks! This looks like a good step forward although ultimately we will want to remove the need to use temporary files at all.

PR #346, opened 2022-02-22
add test in CI for standalone build and run of clvk tests

rjodinchr:
clvk tests can now be build independently from clvk.
This new CI test tests this usecase.

### DIFF
--- a/tests/azure/job-precommit.yml
+++ b/tests/azure/job-precommit.yml
@@ -54,4 +54,10 @@ jobs:
       os: ${{ parameters.os }}
       compilerAvailable: ${{ parameters.compilerAvailable }}
       onlineCompiler: ${{ parameters.onlineCompiler }}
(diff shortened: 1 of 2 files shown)


rjodinchr:
Is there a tool to check the yaml syntax for azure?
I modified a few things, but it is hard to know what is right and what is wrong here.

kpet:
The way I do it is typically to log into the Azure web interface and submit the pipeline manually, it gives explicit error messages with line numbers etc then. As I hinted in the review, I'm not sure it's worth you spending too much time on this given that a) it's rather unlikely this would break b) the CI will (have to) move to Github Actions. Unless you're particularly keen to learn Azure pipelines and drive this to completion, I'm fine with you dropping the PR and creating an issue instead.

rjodinchr:
#347

PR #356, opened 2022-03-01
Compile IL binaries without temp files in online mode

RSenApps:
Updates clspv to include google/clspv#823, which provides support
for input languages other than OpenCL. This allows for using the
CompileFromSourceString api for IL binaries rather than needing
to use temporary files. This resolves the final review comment
in #343.

@dcastagna @jrprice


kpet (approved):
LGTM, thanks for following up!

PR #360, opened 2022-03-07
fix vec3 arg size check in clSetKernelArg

rjodinchr:
If a argument is a vec3, OpenCL requires the size passed to
clSetKernelArg to be 4 times the element size. Accept it even so clspv
arg size is 3 times the element size.

### DIFF
--- a/src/kernel.hpp
+++ b/src/kernel.hpp
@@ -212,11 +212,17 @@ struct cvk_kernel_argument_values {
     cl_int set_arg(const kernel_argument& arg, size_t size, const void* value) {
 
         if (arg.is_pod()) {
-            if (size != arg.size) {
+            // If the argument is a vec3, OpenCL requires to call clSetKernelArg
+            // with a size of 4 times the element size. But clspv arg size is
+            // only 3 times the element size. When size and arg.size do not
+            // match, make sure that we are not in this unusual case.
+            if (size != arg.size &&
+                !(arg.info.type_name[arg.info.type_name.length() - 1] == '3' &&
+                  (size == arg.size * 4 / 3))) {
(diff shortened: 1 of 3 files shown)


rjodinchr:
Compiling with `-cl-kernel-arg-info` enabled by default can be an issue.
Maybe just assuming that if size is exactly 4/3 of arg.size can be enough... as long as we take arg.size after that.
It would mean that we would accept wrong size in clSetKernel but it would not produced overflow or anything. Might be reasonable.

src/kernel.hpp
```diff
@@ -212,11 +212,17 @@ struct cvk_kernel_argument_values {
     cl_int set_arg(const kernel_argument& arg, size_t size, const void* value) {
 
         if (arg.is_pod()) {
-            if (size != arg.size) {
+            // If the argument is a vec3, OpenCL requires to call clSetKernelArg
+            // with a size of 4 times the element size. But clspv arg size is
+            // only 3 times the element size. When size and arg.size do not
+            // match, make sure that we are not in this unusual case.
+            if (size != arg.size &&
+                !(arg.info.type_name[arg.info.type_name.length() - 1] == '3' &&
```
kpet, line 220: I suggest you introduce `kernel_argument::is_vec3()` to help with readability (don't forget to assert the validity of `kernel_argument_info::extended_valid` there) and separation of concerns.

kpet (changes requested):
I'm fine with this as a pragmatic (and hopefully temporary) solution though, as you say, requiring `-cl-kernel-arg-info` is not ideal. I looked at what clspv does layout-wise and currently it doesn't pack arguments so the space is wasted anyway there.

kpet (approved):
Thanks!

PR #363, opened 2022-03-24
fix memory managment of image with packed normalized data type

rjodinchr:


kpet (approved):
Good catch, thanks! I'm guessing this was found by the CTS, wasn't it?

rjodinchr:
Yes. I think for many GPUs tests on packed normalized data types don't do much as they are just not supported, but on Intel chromebooks I saw them failed.

PR #366, opened 2022-03-28
Fix istream passed to readSpirv

DCastagna:
llvm::readSpirv assumes that the input stream containig SPIR-V is seekable.
In particular, when the input SPIR-V contains an OpType* that gets parsed
with SPIRVDecoder::getContinuedInstructions, tellg() and seekg() are called
on the istream.

This CL adds seeking support to the stream we pass as input of readSpirv.

### DIFF
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -52,6 +52,32 @@ struct membuf : public std::streambuf {
         auto send = reinterpret_cast<char*>(end);
         setp(sbegin, send);
     }
+    virtual pos_type seekoff(off_type off, std::ios_base::seekdir dir,
+                             std::ios_base::openmode which) {
+        if (off != 0)
+            return -1;
+        char* whence = eback();
+        if (dir == std::ios_base::cur) {
+            whence = gptr();
+        } else if (dir == std::ios_base::end) {
+            whence = egptr();
+        }
+        char* to = whence + off;
+        if (to >= eback() && to <= egptr()) {
+            setg(eback(), to, egptr());
+            return gptr() - eback();
+        }
+        return -1;
+    }
+
+    virtual pos_type seekpos(pos_type pos, std::ios_base::openmode which) {
+        char* to = eback() + pos;
+        if (to >= eback() && to <= egptr()) {
+            setg(eback(), to, egptr());
+            return gptr() - eback();
+        }
+        return -1;
+    }
 };
 
 struct reflection_parse_data {
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -52,10 +52,8 @@ struct membuf : public std::streambuf {
         auto send = reinterpret_cast<char*>(end);
         setp(sbegin, send);
     }
-    virtual pos_type seekoff(off_type off, std::ios_base::seekdir dir,
-                             std::ios_base::openmode which) {
-        if (off != 0)
-            return -1;
+    pos_type seekoff(off_type off, std::ios_base::seekdir dir,
+                     std::ios_base::openmode which) override {
         char* whence = eback();
         if (dir == std::ios_base::cur) {
             whence = gptr();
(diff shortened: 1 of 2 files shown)


src/program.cpp
```diff
@@ -52,6 +52,32 @@ struct membuf : public std::streambuf {
         auto send = reinterpret_cast<char*>(end);
         setp(sbegin, send);
     }
+    virtual pos_type seekoff(off_type off, std::ios_base::seekdir dir,
```
kpet, line 55: Suggest removing `virtual` (in keeping with the rest of the code base) but adding `override` since that guarantees the function is an override (which is the intended behaviour here).

src/program.cpp
```diff
@@ -52,6 +52,32 @@ struct membuf : public std::streambuf {
...
+        }
+        char* to = whence + off;
+        if (to >= eback() && to <= egptr()) {
+            setg(eback(), to, egptr());
+            return gptr() - eback();
+        }
+        return -1;
+    }
+
+    virtual pos_type seekpos(pos_type pos, std::ios_base::openmode which) {
```
kpet, line 73: same here

src/program.cpp
```diff
@@ -52,6 +52,32 @@ struct membuf : public std::streambuf {
         auto send = reinterpret_cast<char*>(end);
         setp(sbegin, send);
     }
+    virtual pos_type seekoff(off_type off, std::ios_base::seekdir dir,
+                             std::ios_base::openmode which) {
+        if (off != 0)
+            return -1;
```
kpet, line 58: Not sure I understand, the implementation only supports an offset of 0? Looking at the code below, you might have intended the opposite. Also, why treat 0 as an error? Am I missing something?

src/program.cpp
```diff
@@ -52,6 +52,32 @@ struct membuf : public std::streambuf {
...
+        char* whence = eback();
+        if (dir == std::ios_base::cur) {
+            whence = gptr();
+        } else if (dir == std::ios_base::end) {
+            whence = egptr();
+        }
+        char* to = whence + off;
+        if (to >= eback() && to <= egptr()) {
```
kpet, line 66: Should this be `to < egptr()`? Seems `egptr` returns a pointer _past_ the end of the get area (https://en.cppreference.com/w/cpp/io/basic_streambuf/gptr).

src/program.cpp
```diff
@@ -52,6 +52,32 @@ struct membuf : public std::streambuf {
...
+            setg(eback(), to, egptr());
+            return gptr() - eback();
+        }
+        return -1;
+    }
+
+    virtual pos_type seekpos(pos_type pos, std::ios_base::openmode which) {
+        char* to = eback() + pos;
+        if (to >= eback() && to <= egptr()) {
```
kpet, line 75: Same here

tests/simple-from-il-binary/simple.cl
```diff
@@ -1,5 +1,10 @@
+struct ParamStruct {
+    char a;
+};
+
 kernel void test_simple(global uint* out)
 {
+    struct ParamStruct a;
```
kpet, line 7: Fine to leverage the simple test for this for now as a shortcut but if we find more of these SPIR-V related issues, we'll need to extend the unit test framework instead. I'll capture this in an issue once when I merge this PR.

kpet (changes requested):
Thanks!

src/program.cpp
DCastagna: Done.

src/program.cpp
DCastagna: Done

src/program.cpp
DCastagna: Oops, that was some code I had for local debugging, removed.

src/program.cpp
DCastagna: Do we not want to allow to seek to the end of the stream?
For example, do you think seekoff(0, ios_base::end...) should return -1?

kpet: Yes, we do and what you've done is correct. I was off by one on the first review :).

kpet (approved):
LGTM, thanks!

PR #367, opened 2022-03-28
Fix 65fef82 (split of region bigger than vklimits)

rjodinchr:
Looking for region_offset and region_group_offset to know if we can
split the region is not enough as a test using only one of them can
still be split. But it will not be accepted by clvk with the current
test.
Instead let's looking into the kernel options to choose whether we try
it or not.

Also round_up the size of the workgroup to the local size in order to
keep region uniform.

### DIFF
--- a/src/program.hpp
+++ b/src/program.hpp
@@ -27,9 +27,9 @@
 #include "spirv-tools/libspirv.h"
 #include "spirv/1.0/spirv.hpp"
 
+#include "init.hpp"
 #include "memory.hpp"
 #include "objects.hpp"
-#include "init.hpp"
 
 const int SPIR_WORD_SIZE = 4;
 
--- a/src/program.hpp
+++ b/src/program.hpp
@@ -565,8 +565,11 @@ struct cvk_program : public _cl_program, api_object<object_magic::program> {
     }
 
     bool can_split_region() {
-        return options_allow_split_region(m_build_options) ||
-               options_allow_split_region(gCLSPVOptions);
+        return options_allow_split_region(m_build_options)
+#if COMPILER_AVAILABLE
+               || options_allow_split_region(gCLSPVOptions)
+#endif
+            ;
     }
 
 private:
--- a/src/program.hpp
+++ b/src/program.hpp
@@ -29,6 +29,7 @@
 
 #include "memory.hpp"
 #include "objects.hpp"
+#include "init.hpp"
 
 const int SPIR_WORD_SIZE = 4;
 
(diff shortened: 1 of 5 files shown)


src/program.hpp
```diff
@@ -556,9 +557,19 @@ struct cvk_program : public _cl_program, api_object<object_magic::program> {
...
     }
 
-    bool has_region_offset() const {
-        return push_constant(pushconstant::region_offset) &&
-               push_constant(pushconstant::region_group_offset);
+    bool options_allow_split_region(std::string options) {
+        return options.find("-cl-std=CL2.0") != std::string::npos ||
+               options.find("-cl-std=CL3.0") != std::string::npos ||
+               options.find("cl-arm-non-uniform-work-group-size") !=
+                   std::string::npos;
```
kpet, line 564: This doesn't exactly mirror the set of conditions in clspv. It doesn't feel great to hardcode the same set of conditions here. Maybe there is a case to have clspv pass the information in the SPIR-V (i.e. this program was compiled with support for non-uniform NDRanges)? Maybe we should always enable support for non-uniform NDRanges instead of splitting opportunistically? We're definitely going to need more tests to cover all the cases as well :).

src/queue.cpp
```diff
@@ -549,8 +549,8 @@ cl_int cvk_command_kernel::dispatch_uniform_region_iterate(
 
     size_t num_splitted_regions =
         ceil_div(num_workgroups[dim], vklimits.maxComputeWorkGroupCount[dim]);
-    size_t splitted_region_gws =
-        ceil_div(region.gws[dim], num_splitted_regions);
+    size_t splitted_region_gws = round_up(
```
kpet, line 552: I haven't tried to verify this but it feels like gws calculations are getting more complex than they should be. Each region's gws will be a round multiple of the number of workgroups since we've already done non-uniform splitting so the gws calculation should be as simple as:
```
gws = N * lws
```

I suggest we refactor the logic to make this apparent as a separate PR.

kpet (commented):
Getting a bit late here but I think I got this right. Let's see if tomorrow me still agrees :).

src/program.hpp
rjodinchr: I feel like always enabling the non-uniform NRRanges is the right move here.
It will be like that anyway when the default version will be 3.0.

src/queue.cpp
rjodinchr: I think I got things too complicated because for some reason I wanted to have regions that are of the same size.
But it will definitely be easier to just make regions reach the limit, except for the last one.
```
size_t splitted_region_gws = vklimits.maxComputeWorkGroupCount[dim] * region.lws[dim];
```
We already have the `std::min` doing what we want to deal with the last region.

src/program.hpp
kpet: Yes, we'll have that on by default once we switch to 3.0 (we'll need to rework the clspv flags so the only way to do this is not a deprecated Arm extension). There are ways to optimise further by having multiple variants of the compiled programs but we can treat that as an optimisation for later. Let's always enable non-uniform NDRanges for now. I'll capture options for further optimisations as part of merging this PR.

src/queue.cpp
kpet: Yup, easier to follow now, thanks!

kpet (changes requested):
Let's get rid of all the complexity and always enable support for non-uniform NDRanges.

kpet (approved):
Thanks!

PR #369, opened 2022-04-01
Define OpenCL as an application framework on Apple OSes.

woachk:
This allows straightforward replacement and easy use of install_name_tool down the road to use this on existing application binaries.


kpet (approved):
LGTM, thanks! I've seen your related blog post, thanks for the chuckle :).

PR #370, opened 2022-04-04
fix ceil-div to avoid overflow

rjodinchr:
some platform have maxComputeWorkGroupCount equal to UINT_MAX.
It generates an overflow when computing num_splitted_regions in
cvk_command_kernel::dispatch_uniform_region_iterate

### DIFF
--- a/src/utils.hpp
+++ b/src/utils.hpp
@@ -17,7 +17,6 @@
 #include "log.hpp"
 
 #include <cassert>
-#include <cstdint>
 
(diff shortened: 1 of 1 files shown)


kpet (changes requested):
I think this can be done with a simpler expression that doesn't require any branches: see https://stackoverflow.com/questions/2745074/fast-ceiling-of-an-integer-division-in-c-c

PR #371, opened 2022-04-14
memory: fix free of cvk_memory_allocation when using image1dbuffer

rjodinchr:
Add a test which is segfaulting without this fix due to a double free.

### DIFF
--- a/src/memory.cpp
+++ b/src/memory.cpp
@@ -108,7 +108,7 @@ bool cvk_buffer::init() {
     }
 
     // Allocate memory
-    m_memory = std::make_unique<cvk_memory_allocation>(
+    m_memory = std::make_shared<cvk_memory_allocation>(
         vkdev, params.size, params.memory_type_index);
     res = m_memory->allocate();
 
@@ -341,7 +341,7 @@ bool cvk_image::init() {
 
     if (m_desc.image_type == CL_MEM_OBJECT_IMAGE1D_BUFFER) {
         auto buffer = static_cast<cvk_mem*>(m_desc.buffer);
-        m_memory = std::unique_ptr<cvk_memory_allocation>(buffer->memory());
+        m_memory = buffer->memory();
         buffer->retain();
     } else {
         // Select memory type
--- a/src/memory.hpp
+++ b/src/memory.hpp
@@ -128,9 +128,9 @@ struct cvk_mem : public _cl_mem, api_object<object_magic::memory_object> {
         return m_properties;
     }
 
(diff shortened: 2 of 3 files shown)


src/memory.hpp
```diff
@@ -20,11 +20,15 @@
...
 #include "utils.hpp"
 
-struct cvk_memory_allocation {
+struct cvk_memory_allocation;
+using cvk_memory_allocation_holder = refcounted_holder<cvk_memory_allocation>;
 
-    cvk_memory_allocation(VkDevice dev, VkDeviceSize size, uint32_t type_index)
-        : m_device(dev), m_size(size), m_memory(VK_NULL_HANDLE),
-          m_memory_type_index(type_index) {}
+struct cvk_memory_allocation : public api_object<object_magic::memory_object> {
```
kpet, line 26: This doesn't look right. A `cvk_memory_allocation` is not an OpenCL API object.

kpet (commented):
I've finally got around to having a look at this. I can reproduce the crash with your test at my end. 1D Buffer image objects need to hold a non-owning reference to their memory. This will also be the case for other types of images created from buffers. I think I'd rather we find an RAII-friendly solution to this problem. A few ideas that come to mind (getting late in the day to make a decision):
- Use a shared pointer.
- Add a layer of indirection. I'm thinking something like a "memory view" that would hold a reference to the underlying memory and represent a window over it. There are other motivators for introducing such a construct (reasoning about overlaps, sub-buffers, etc).
- Refcount memory allocations using refcounted.

src/memory.cpp
```diff
@@ -341,7 +341,7 @@ bool cvk_image::init() {
 
     if (m_desc.image_type == CL_MEM_OBJECT_IMAGE1D_BUFFER) {
         auto buffer = static_cast<cvk_mem*>(m_desc.buffer);
-        m_memory = std::unique_ptr<cvk_memory_allocation>(buffer->memory());
+        m_memory = buffer->memory();
         buffer->retain();
     } else {
         // Select memory type
```
kpet, line 347: You could have used `make_shared` in this case too. The code as it is is well-defined so not a problem, it's just a bit inconsistent.

kpet (approved):
Thinking about this more, I don't like the fact that the use of `shared_ptr` means we have two parallel refcounts between the memory objects and their underlying memory allocation. Long-term, I'll probably implement something like a memory view/windows that can be owning or non-owning but this change solves the issue today and is clean enough. Thanks!

PR #374, opened 2022-04-22
clean api_tests output

rjodinchr:
api_tests output was containing the null character ('\0').
When you store the output of the test in a file, this character makes
some system (like the browser chrome) consider the file as binary
rather than text.
trim this null character in order to let chrome open the file as text.

### DIFF
--- a/tests/api/main.cpp
+++ b/tests/api/main.cpp
@@ -26,6 +26,12 @@
         }                                                                      \
     } while (0)
 
+static void trim_null_terminator(std::string& str) {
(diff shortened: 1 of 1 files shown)


kpet:
Right, that's because the NUL is included into the string. I've fixed this with https://github.com/kpet/clvk/commit/7a57a042ecaffdd8b292909f949451a5ba06cdc7. Let me know if that works for you.

rjodinchr:
It works for me, thanks

PR #385, opened 2022-07-14
Document BOINC supports clvk

truboxl:
Refer BOINC/boinc#4847

### DIFF
--- a/docs/supported-applications.md
+++ b/docs/supported-applications.md
@@ -16,7 +16,7 @@
 | [BabelStream](https://github.com/UoB-HPC/BabelStream) | Untested |
 | [Theano](https://github.com/Theano/Theano) | Untested |
 | [CLBlast](https://github.com/CNugteren/CLBlast) | Untested |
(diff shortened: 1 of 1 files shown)


kpet:
Thanks for testing and creating a PR. clvk was at fault here and I've fixed it with https://github.com/kpet/clvk/pull/386. Was that the only issue you faced? Should this PR be changed to say "supported" or are there other issues?

truboxl:
Thanks @kpet. I will verify with the latest clvk commit once I got back.
There's still other issues like how BOINC does not like max freq as 0MHz for FLOPS calculation and they just assign 100Giga FLOPS...

kpet:
Great, thanks! Feel free to open a PR to change the default frequency to some plausible (500 MHz?) hard-coded value. I can see applications tripping over that 0 if they attempt to use it in calculations (they shouldn't!!).

kpet (approved):
Thanks!

PR #387, opened 2022-07-21
add support for get image channel order and data type

rjodinchr:
Update clspv and SPIRV-Headers

### DIFF
--- a/src/kernel.cpp
+++ b/src/kernel.cpp
@@ -40,6 +40,10 @@ cl_int cvk_kernel::init() {
     // Store a copy of the arguments
     m_args = m_entry_point->args();
 
+    if (const auto* md = m_entry_point->image_metadata()) {
+        m_image_metadata = md;
+    }
+
     // Mark all arguments as unset
     m_args_set.resize(m_args.size(), false);
 
@@ -73,6 +77,32 @@ std::unique_ptr<cvk_kernel> cvk_kernel::clone(cl_int* errcode_ret) const {
     return kernel;
 }
 
+void cvk_kernel::set_image_metadata(cl_uint index, const void* image) {
+    if (!m_image_metadata) {
+        return;
+    }
+    auto md = m_image_metadata->find(index);
+    if (md != m_image_metadata->end()) {
+
+        auto mem = icd_downcast(*reinterpret_cast<const cl_mem*>(image));
+        assert(mem->is_image_type());
+        auto format = static_cast<cvk_image*>(mem)->format();
+
+        if (md->second.has_valid_order()) {
+            auto order_offset = md->second.order_offset;
+            auto order = format.image_channel_order;
+            m_argument_values->set_pod_data(order_offset, sizeof(order),
+                                            &order);
+        }
+        if (md->second.has_valid_data_type()) {
+            auto data_type_offset = md->second.data_type_offset;
+            auto data_type = format.image_channel_data_type;
+            m_argument_values->set_pod_data(data_type_offset, sizeof(data_type),
+                                            &data_type);
+        }
+    }
+}
+
 cl_int cvk_kernel::set_arg(cl_uint index, size_t size, const void* value) {
     std::lock_guard<std::mutex> lock(m_lock);
 
@@ -92,6 +122,13 @@ cl_int cvk_kernel::set_arg(cl_uint index, size_t size, const void* value) {
     if (ret == CL_SUCCESS) {
         // Mark argument as set
         m_args_set[index] = true;
+    }
+
+    // if the argument is an image, we need to set its metadata
+    // (channel_order/channel_data_type).
+    if (arg.kind == kernel_argument_kind::ro_image ||
+        arg.kind == kernel_argument_kind::wo_image) {
+        set_image_metadata(index, value);
     }
 
     return ret;
--- a/src/kernel.hpp
+++ b/src/kernel.hpp
@@ -29,7 +29,7 @@ struct cvk_kernel : public _cl_kernel, api_object<object_magic::kernel> {
 
     cvk_kernel(cvk_program* program, const char* name)
         : api_object(program->context()), m_program(program),
-          m_entry_point(nullptr), m_name(name) {}
+          m_entry_point(nullptr), m_name(name), m_image_metadata(nullptr) {}
 
     CHECK_RETURN cl_int init();
     std::unique_ptr<cvk_kernel> clone(cl_int* errcode_ret) const;
@@ -41,6 +41,12 @@ struct cvk_kernel : public _cl_kernel, api_object<object_magic::kernel> {
     std::shared_ptr<cvk_kernel_argument_values> argument_values() const {
         return m_argument_values;
     }
+
+    const kernel_image_metadata_map* get_image_metadata() const {
+        return m_image_metadata;
+    }
+
+    void set_image_metadata(cl_uint index, const void* image);
 
     CHECK_RETURN cl_int set_arg(cl_uint index, size_t size, const void* value);
     CHECK_RETURN VkPipeline
@@ -149,6 +155,7 @@ private:
     std::vector<kernel_argument> m_args;
     std::shared_ptr<cvk_kernel_argument_values> m_argument_values;
     std::vector<bool> m_args_set;
+    const kernel_image_metadata_map* m_image_metadata;
 };
 
 static inline cvk_kernel* icd_downcast(cl_kernel kernel) {
@@ -222,7 +229,10 @@ struct cvk_kernel_argument_values {
             if (m_pod_arg == nullptr) {
                 return CL_INVALID_PROGRAM;
             }
+        }
 
+        if (m_entry_point->has_pod_arguments() ||
+            m_entry_point->has_image_metadata()) {
             // TODO(#101): host out-of-memory errors are currently unhandled.
             auto buffer = std::make_unique<std::vector<uint8_t>>(
                 m_entry_point->pod_buffer_size());
@@ -241,6 +251,10 @@ struct cvk_kernel_argument_values {
         }
     }
 
+    void set_pod_data(uint32_t offset, size_t size, const void* value) {
+        memcpy(&pod_data()[offset], value, size);
+    }
+
     cl_int set_arg(const kernel_argument& arg, size_t size, const void* value) {
 
         if (arg.is_pod()) {
@@ -253,7 +267,7 @@ struct cvk_kernel_argument_values {
                 return CL_INVALID_ARG_SIZE;
             }
 
-            memcpy(&pod_data()[arg.offset], value, arg.size);
+            set_pod_data(arg.offset, arg.size, value);
         } else if (arg.kind == kernel_argument_kind::local) {
             CVK_ASSERT(value == nullptr);
             m_local_args_size[arg.pos] = size;
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -134,6 +134,10 @@ spv_result_t parse_reflection(void* user_data,
             return pushconstant::num_workgroups;
         case NonSemanticClspvReflectionPushConstantRegionGroupOffset:
             return pushconstant::region_group_offset;
+        case NonSemanticClspvReflectionImageArgumentInfoChannelOrderPushConstant:
+            return pushconstant::image_metadata;
+        case NonSemanticClspvReflectionImageArgumentInfoChannelDataTypePushConstant:
+            return pushconstant::image_metadata;
         default:
             cvk_error_fn("Unhandled reflection instruction for push constant");
             break;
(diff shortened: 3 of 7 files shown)


src/program.hpp
```diff
@@ -124,6 +128,11 @@ struct constant_data_buffer_info {
     std::vector<char> data;
 };
 
+using kernel_image_channel_metadata_map =
+    std::unordered_map<uint32_t, std::pair<uint32_t, uint32_t>>;
```
kpet, line 132: I think I'd rather we use introduce a struct for the metadata itself straight away. This would make the mechanism more extensible and improve readability (this results in code like  `second.first` and `second.second` in places already).

src/program.hpp
```diff
@@ -219,12 +231,33 @@ class spir_binary {
         m_constant_data_buffer.reset(new constant_data_buffer_info(info));
     }
 
+    void add_image_channel_order_metadata(const std::string& name,
+                                          uint32_t ordinal, uint32_t offset) {
+        if (m_image_channel_metadata[name].count(ordinal) > 0) {
+            m_image_channel_metadata[name][ordinal].first = offset;
+        } else {
+            m_image_channel_metadata[name][ordinal] =
+                std::make_pair(offset, IMAGE_METADATA_EMPTY);
```
kpet, line 240: If the metadata were a struct you could rely on the default constructor and all that special handling would not be necessary.

src/program.cpp
```diff
@@ -1521,6 +1552,26 @@ cl_int cvk_entry_point::init() {
...
+            image_md_is_max = true;
+        }
+        if (data_type_offset != IMAGE_METADATA_EMPTY &&
+            data_type_offset > m_pod_buffer_size) {
+            m_pod_buffer_size = data_type_offset;
+            image_md_is_max = true;
+        }
+    }
+    if (image_md_is_max) {
+        m_pod_buffer_size = round_up(m_pod_buffer_size + 4, 4);
```
kpet, line 1572: Why `+4`?

src/program.hpp
```diff
@@ -286,10 +319,18 @@ class cvk_entry_point {
 
     const std::vector<kernel_argument>& args() const { return m_args; }
 
+    const kernel_image_channel_metadata_map image_channel_metadata() const {
```
kpet, line 322: Guessing you intended the return type to be a reference, didn't you?

src/kernel.hpp
```diff
@@ -42,6 +42,12 @@ struct cvk_kernel : public _cl_kernel, api_object<object_magic::kernel> {
         return m_argument_values;
     }
 
+    kernel_image_channel_metadata_map& get_image_channel_metadata() {
```
kpet, line 45: This should probably return a const reference and be a const function.

tests/api/images.cpp
```diff
@@ -435,3 +435,42 @@ TEST_F(WithContext, Image1DBuffer) {
 
     auto image = CreateImage(CL_MEM_READ_WRITE, &format, &desc);
 }
+
+TEST_F(WithCommandQueue, ImageChannelGetter) {
+    const cl_channel_order order = CL_RGBA;
+    const cl_channel_type data_type = CL_SIGNED_INT32;
```
kpet, line 441: Might be worth testing with a few different formats.

tests/api/images.cpp
```diff
@@ -435,3 +435,42 @@ TEST_F(WithContext, Image1DBuffer) {
...
+    const cl_image_format format = {order, data_type};
+    const cl_image_desc desc = {
+        CL_MEM_OBJECT_IMAGE2D, 1, 1, 1, 1, 0, 0, 0, 0, nullptr};
+    auto image = CreateImage(CL_MEM_READ_ONLY, &format, &desc);
+    auto order_buffer =
+        CreateBuffer(CL_MEM_WRITE_ONLY, sizeof(cl_uint), nullptr);
+    auto data_type_buffer =
+        CreateBuffer(CL_MEM_WRITE_ONLY, sizeof(cl_uint), nullptr);
```
kpet, line 449: You could probably use a single buffer.

src/queue.cpp
```diff
@@ -393,6 +393,22 @@ void cvk_command_kernel::update_global_push_constants(
...
+                               sizeof(uint32_t),
+                               &m_argument_values->pod_data()[order_offset]);
+        }
+        if (data_type_offset != IMAGE_METADATA_EMPTY) {
+            vkCmdPushConstants(
+                command_buffer, m_kernel->pipeline_layout(),
+                VK_SHADER_STAGE_COMPUTE_BIT, data_type_offset, sizeof(uint32_t),
+                &m_argument_values->pod_data()[data_type_offset]);
+        }
+    }
```
kpet, line 411: Maybe this could be rewritten as a single push constant update for the entire range that contains image metadata? This would be in keeping with the pattern above, facilitating further shortening and likely faster.

src/program.cpp
```diff
@@ -1521,6 +1552,26 @@ cl_int cvk_entry_point::init() {
...
+            order_offset > m_pod_buffer_size) {
+            m_pod_buffer_size = order_offset;
+            image_md_is_max = true;
+        }
+        if (data_type_offset != IMAGE_METADATA_EMPTY &&
+            data_type_offset > m_pod_buffer_size) {
+            m_pod_buffer_size = data_type_offset;
+            image_md_is_max = true;
+        }
+    }
```
kpet, line 1570: Not sure I understand. Does this assume that the offset into the push constants provided by clspv can be used as an offset into the POD buffer? Surely this won't always hold. Maybe a test case for when this doesn't hold would be useful?

kpet (changes requested):
Thanks! A few changes needed.

src/program.hpp
rjodinchr: Done

src/program.hpp
rjodinchr: Done

src/program.cpp
rjodinchr: sorry for that, just the size of the offset. changed it for `sizeof(uint32_t)`

src/program.hpp
rjodinchr: indeed

src/kernel.hpp
rjodinchr: done

tests/api/images.cpp
rjodinchr: Sure. I stayed with formats that should be available everywhere.

tests/api/images.cpp
rjodinchr: Done

src/queue.cpp
rjodinchr: Your comment feels weird, as the code pushing pod args is not how you explain here.
I have modified the image_metadata implementation as well as the pod arg to be aggregated into one call to pushConstant.
Tell me if it was not what you were expecting.

src/program.cpp
rjodinchr: Yes, that's a first basic implementation which is only capable of sending those values through the pod. And yes it was not checking for the pod buffer size. It is now checking it and return an error if it exceeds it.

src/queue.cpp
rjodinchr: I'm not confident about what I have done here.
Maybe I should have at least 2 pushconstants, one for the args, one for the image metadata.
Otherwise it feels like the other information (global_size, num_workgroups) could be overridden by this last pushconstant.
What do you think?

rjodinchr: I have reverted this part to have no modifications regarding the args pushconstants

kpet (changes requested):
I've left a few more comments on the current implementation but I think you'll need to take a step back: I've played with the code and I think there are quite a few fundamental problems with the approach (see part of what I've tried on https://github.com/kpet/clvk/tree/image-metadata):
- You're not updating push constant ranges to take into account the image metadata
- The test as modified no longer works
- If POD arguments don't fit in push constants the offset into the pod buffer overlaps the offset into push constants for image metadata. This can't work in the general case (you could take my modified test in the branch and add enough POD arguments between `magic` and the image argument to test this).
- The validation layers complain that some of the image formats used in the test are not supported for me. Maybe the test should query supported formats?
- The validation layers complain about a mismatch between the component type expected by the SPIR-V module and that of the image (maybe that's a pre-existing issue?)

Overall, more work needed on this (and more tests :)!).

src/program.cpp
```diff
@@ -183,6 +187,28 @@ spv_result_t parse_reflection(void* user_data,
...
+            }
+            case NonSemanticClspvReflectionImageArgumentInfoChannelDataTypePushConstant: {
+                auto kernel = parse_data->strings[inst->words[5]];
+                auto ordinal = parse_data->constants[inst->words[6]];
+                auto offset = parse_data->constants[inst->words[7]];
+                parse_data->binary->add_image_channel_data_type_metadata(
+                    kernel, ordinal, offset);
+                auto pc = inst_to_push_constant(ext_inst);
+                parse_data->binary->add_push_constant(
+                    pc, {offset, sizeof(uint32_t)});
```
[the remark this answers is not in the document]
rjodinchr: Done

src/program.cpp
```diff
@@ -1530,6 +1561,31 @@ cl_int cvk_entry_point::init() {
         m_pod_buffer_size = round_up(m_pod_buffer_size, 4);
     }
 
+    // Take the size of image metadata into account for the pod buffer size
+    bool image_md_is_max = false;
+    for (auto md : m_image_channel_metadata) {
+        auto order_offset = md.second.order_offset;
+        auto data_type_offset = md.second.data_type_offset;
+        if (md.second.is_valid_order() && order_offset > m_pod_buffer_size) {
```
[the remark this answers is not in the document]
rjodinchr: reworked

src/program.cpp
```diff
@@ -1530,6 +1561,31 @@ cl_int cvk_entry_point::init() {
         m_pod_buffer_size = round_up(m_pod_buffer_size, 4);
     }
 
+    // Take the size of image metadata into account for the pod buffer size
+    bool image_md_is_max = false;
+    for (auto md : m_image_channel_metadata) {
```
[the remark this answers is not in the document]
rjodinchr: done

src/kernel.cpp
```diff
@@ -40,6 +40,8 @@ cl_int cvk_kernel::init() {
     // Store a copy of the arguments
     m_args = m_entry_point->args();
 
+    m_image_channel_metadata = m_entry_point->image_channel_metadata();
```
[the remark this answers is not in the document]
rjodinchr: done

src/program.hpp
```diff
@@ -128,6 +129,31 @@ struct spirv_validation_options {
...
+    void set_order(uint32_t order) {
+        order_offset = order;
+        valid_order = true;
+    }
+    void set_data_type(uint32_t data_type) {
+        data_type_offset = data_type;
+        valid_data_type = true;
+    }
+    bool is_valid_order() const { return valid_order; }
+    bool is_valid_data_type() const { return valid_data_type; }
```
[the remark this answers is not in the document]
rjodinchr: done

src/program.hpp
```diff
@@ -128,6 +129,31 @@ struct spirv_validation_options {
...
+    image_channel_metadata() : valid_order(false), valid_data_type(false) {}
+    uint32_t order_offset;
+    uint32_t data_type_offset;
+    void set_order(uint32_t order) {
+        order_offset = order;
+        valid_order = true;
+    }
+    void set_data_type(uint32_t data_type) {
+        data_type_offset = data_type;
+        valid_data_type = true;
```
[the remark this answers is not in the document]
rjodinchr: done

src/program.cpp
```diff
@@ -1454,6 +1480,11 @@ bool cvk_entry_point::
 cl_int cvk_entry_point::init() {
     VkResult res;
 
+    // Get the image metadata for the this entry point
+    if (auto* md = m_program->image_channel_metadata(m_name)) {
+        m_image_channel_metadata = *md;
```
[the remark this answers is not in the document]
rjodinchr: done

src/program.cpp
```diff
@@ -1530,6 +1561,31 @@ cl_int cvk_entry_point::init() {
...
+            image_md_is_max = true;
+        }
+        if (md.second.is_valid_data_type() &&
+            data_type_offset > m_pod_buffer_size) {
+            m_pod_buffer_size = data_type_offset;
+            image_md_is_max = true;
+        }
+    }
+    if (image_md_is_max) {
+        m_pod_buffer_size = round_up(m_pod_buffer_size + sizeof(uint32_t), 4);
```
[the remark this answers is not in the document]
rjodinchr: done

src/queue.cpp
```diff
@@ -410,6 +410,32 @@ void cvk_command_kernel::update_global_push_constants(
                            &num_workgroups);
     }
 
+    uint32_t min_pod_data_offset = UINT32_MAX;
+    uint32_t max_pod_data_offset = 0;
```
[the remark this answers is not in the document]
rjodinchr: done

rjodinchr:
> I've left a few more comments on the current implementation but I think you'll need to take a step back: I've played with the code and I think there are quite a few fundamental problems with the approach (see part of what I've tried on https://github.com/kpet/clvk/tree/image-metadata):
> 
> * You're not updating push constant ranges to take into account the image metadata
> * The test as modified no longer works
> * If POD arguments don't fit in push constants the offset into the pod buffer overlaps the offset into push constants for image metadata. This can't work in the general case (you could take my modified test in the branch and add enough POD arguments between `magic` and the image argument to test this).
> * The validation layers complain that some of the image formats used in the test are not supported for me. Maybe the test should query supported formats?
> * The validation layers complain about a mismatch between the component type expected by the SPIR-V module and that of the image (maybe that's a pre-existing issue?)
> 
> Overall, more work needed on this (and more tests :)!).

Indeed, there is a bug in the clspv implementation using the new ordinal while ordinals have changed when removing the pods from the kernel prototype.
I have a fix for it that I will push tomorrow.
For the part if POD arguments don't fit in push constants and the kernel is using the image metadata, it is just not supported at the moment.
I have never used the validation layers before and I am not sure where to look for their outputs.

rjodinchr:
https://github.com/google/clspv/pull/902

kpet (approved):
Looks good, thanks! As discussed there is still quite a bit of work for us to do around push constants, POD handling, etc but this change gives us a good starting point (and a great way to expose the scaling weaknesses of the current design!). To be continued :).

rjodinchr:
Yes, I'll do those things in the coming weeks.
Thanks

kpet (approved):
Thanks! Don't feel you have to do it all but feel free if you want :).

PR #388, opened 2022-07-25
Explicitly include tmp folder for ONLINE compilation

rjodinchr:
In ONLINE compilation mode, clspv does not include any path by
default (which it does with the input path in OFFLINE compilation
mode).
Explicitly add tmp folder to cover this case


kpet (approved):
This change is fine for now but ideally we'd add an interface for clspv to accept additional headers as strings so we don't need to save them to disk. I've created https://github.com/kpet/clvk/issues/393.

PR #389, opened 2022-07-25
device.cpp: add CapabilitySampledBuffer

rjodinchr:


kpet (commented):
Reporting the capability as supported is not really accurate since clvk does not support producing texel buffer descriptors. What we have today happens to work but is incorrect according to the specifications. We still need to agree a path forward for 1D Buffer images and fully align clvk and clspv (and/or):
- Use texel buffers
- Use linearly tiled 1D images

PR #390, opened 2022-07-27
Allow standard layout for UBOs when validating the SPIR-V if supported

callumfare:
It's possible for clspv to generate UBO layouts that require VK_KHR_uniform_buffer_standard_layout. If the device supports it, tell the SPIR-V validator.

This fixes `basic/parameter_types` from the CTS when combined with this clspv fix - https://github.com/google/clspv/pull/886


kpet (approved):
LGTM, thanks!

Issue #413, opened 2022-08-15
Image metadata support causing CTS failures


alan-baker:
#387 added support for image metadata being passed as push constants. This is causing CTS failures when I try to import it internally. At least api `min_max_parameter_size` is failing due to this. It appears that the check at https://github.com/kpet/clvk/blob/main/src/program.cpp#L1605 does not account for where the pod buffer will be passed as push constants or not. On my internal device clvk report support for 1024 parameters, but only 256B of push constant room. Clspv correctly uses a uniform buffer for the arguments, but check still fails.

CC: @rjodinchr

alan-baker:
Other failures potentially related to this:
```
buffers:buffer_read_array_barrier_char
buffers:buffer_read_array_barrier_long
```

rjodinchr:
```
buffers_buffer_migrate
buffers_buffer_read_array_barrier_char
buffers_buffer_read_array_barrier_float
buffers_buffer_read_array_barrier_int
buffers_buffer_read_array_barrier_long
buffers_buffer_read_array_barrier_short
buffers_buffer_read_array_barrier_uchar
buffers_buffer_read_array_barrier_uint
buffers_buffer_read_array_barrier_ulong
buffers_buffer_read_array_barrier_ushort
buffers_image_migrate
events_event_enqueue_barrier_with_event_list
events_event_enqueue_marker
events_event_enqueue_marker_with_event_list
```
Those ones already have an issue filed: https://github.com/kpet/clvk/issues/411

Which just leaves `min_max_parameter_size`
I will look at it

kpet:
@alan-baker @rjodinchr's PRs have now gone in and all the CTS tests mentioned here are now passing at my end.

PR #414, opened 2022-08-16
fix check for max push constants size

rjodinchr:
we need to check for only the push constant range, not the whole pod
buffer as every pod will not be in the push constants.

#413


kpet (approved):
LGTM

Issue #411, opened 2022-08-08
segfault running CTS tests


rjodinchr:
```
buffers_buffer_migrate
buffers_buffer_read_array_barrier_char
buffers_buffer_read_array_barrier_float
buffers_buffer_read_array_barrier_int
buffers_buffer_read_array_barrier_long
buffers_buffer_read_array_barrier_short
buffers_buffer_read_array_barrier_uchar
buffers_buffer_read_array_barrier_uint
buffers_buffer_read_array_barrier_ulong
buffers_buffer_read_array_barrier_ushort
buffers_image_migrate
events_event_enqueue_barrier_with_event_list
events_event_enqueue_marker
events_event_enqueue_marker_with_event_list
```
All of them are now failing with this error:
```
queue.hpp:355: virtual const std::vector<cvk_mem *> cvk_command::memory_objects() const: Assertion `false && "Should never be called"' failed.
```
Here is the backtrace from `buffer buffer_migrate`:
```
$ lldb -o r -- test_buffers buffer_migrate
(lldb) target create "test_buffers"
Current executable set to 'test_buffers' (x86_64).
(lldb) settings set -- target.run-args  "buffer_migrate"
(lldb) r
 Initializing random seed to 0.
Requesting Default device based on command line for platform index 0 and device index 0
Compute Device Name = SwiftShader Device (LLVM 10.0.0), Compute Device Vendor = FIXME, Compute Device Version = OpenCL 3.0 CLVK on Vulkan v1.3.0 driver 20971520, CL C Version = OpenCL C 1.2 CLVK on Vulkan v1.3.0 driver 20971520
Device latest conformance version passed: FIXME
Supports single precision denormals: NO
sizeof( void*) = 8  (host)
sizeof( void*) = 4  (device)
buffer_migrate...
test_buffers: /usr/local/google/home/rjodin/work/clvk/src/queue.hpp:355: virtual const std::vector<cvk_mem*> cvk_command::memory_objects() const: Assertion `false && "Should never be called"' failed.
Process 1117297 stopped
* thread #1, name = 'test_buffers', stop reason = hit program assert
    frame #4: 0x00007fffedbbf3f4 libOpenCL.so.1`cvk_command::memory_objects(this=0x00005555556a01b0) const at queue.hpp:355:9
   352      const std::vector<cvk_event*>& dependencies() const { return m_event_deps; }
   353
   354      virtual const std::vector<cvk_mem*> memory_objects() const {
[...]

PR #416, opened 2022-08-30
cvk_command_dep: add memory_objects method

rjodinchr:
This method is needed to avoid asserting on cvk_command virtual
implementation.

Ref #411


kpet (approved):
LGTM, thanks!

PR #418, opened 2022-09-04
fix build of api_tests

rjodinchr:
`tests/api/split_region.cpp` has a dependency on `utils.hpp` for
`ARRAY_SIZE`.
And `utils.hpp` has a dependency on `vulkan.h`. This header is not
always in the includes path, making the build fail.


kpet (approved):
What was the build that broke? Could we extend the CI?

rjodinchr:
Only the local build on my machine. I don't understand how it managed to build before. But after a full clean, I was not able to rebuild anymore.

kpet:
Was that a standalone build of the tests?

rjodinchr:
No, it was a full rebuild

kpet:
Ok, if you do find out what we're missing in the CI, feel free to create an issue.

PR #420, opened 2022-09-12
Implement printf

callumfare:
Clspv provides a printf definition that stores the value of each argument in a program-wide printf storage buffer. This implementation uses the appropriate ClspvReflection instructions to parse this data and print the resulting strings at runtime.

This fixes all failing printf CTS tests, other than one which requires physical addressing (to print the value of a pointer).

Related Pull Requests:
https://github.com/KhronosGroup/SPIRV-Registry/pull/168
https://github.com/KhronosGroup/SPIRV-Headers/pull/290
https://github.com/google/clspv/pull/925

### DIFF
--- a/src/kernel.cpp
+++ b/src/kernel.cpp
@@ -162,7 +162,7 @@ bool cvk_kernel_argument_values::setup_descriptor_sets() {
     // Setup module-scope variables
     if (program->module_constant_data_buffer() != nullptr &&
         program->module_constant_data_buffer_info()->type ==
-            constant_data_buffer_type::storage_buffer) {
+            module_buffer_type::storage_buffer) {
         auto buffer = program->module_constant_data_buffer();
         auto info = program->module_constant_data_buffer_info();
         cvk_debug_fn(
--- a/src/program.hpp
+++ b/src/program.hpp
@@ -32,6 +32,7 @@
 #include "init.hpp"
 #include "memory.hpp"
 #include "objects.hpp"
+#include "printf.hpp"
 
 const int SPIR_WORD_SIZE = 4;
 
@@ -124,6 +125,7 @@ enum class pushconstant
     region_group_offset,
     image_metadata,
     module_constants_pointer,
+    printf_buffer_pointer,
 };
 
 struct pushconstant_desc {
@@ -143,18 +145,26 @@ enum class spec_constant
     subgroup_max_size,
 };
 
-enum class constant_data_buffer_type
+enum class module_buffer_type
 {
     storage_buffer,
     pointer_push_constant,
 };
 
 struct constant_data_buffer_info {
-    constant_data_buffer_type type;
+    module_buffer_type type;
     uint32_t set;
     uint32_t binding;
     uint32_t pc_offset;
     std::vector<char> data;
+};
+
+struct printf_buffer_desc_info {
+    module_buffer_type type;
+    uint32_t set;
+    uint32_t binding;
+    uint32_t pc_offset;
+    uint32_t size = 0;
 };
 
 struct spirv_validation_options {
@@ -182,6 +192,7 @@ class spir_binary {
         std::unordered_map<std::string, std::vector<kernel_argument>>;
     using kernels_reqd_work_group_size_map =
         std::unordered_map<std::string, std::array<uint32_t, 3>>;
+    using kernels_flags_map = std::unordered_map<std::string, uint32_t>;
 
 public:
     spir_binary(spv_target_env env)
@@ -238,7 +249,13 @@ public:
         }
     }
 
-    void add_kernel(const std::string& name, uint32_t num_args) {
+    const printf_descriptor_map_t& printf_descriptors() const {
+        return m_printf_descriptors;
+    }
+
+    void add_kernel(const std::string& name, uint32_t num_args,
+                    uint32_t flags) {
+        m_flags[name] = flags;
         auto& args = m_dmaps[name];
         kernel_argument unused = {
             {}, 0, 0, 0, 0, 0, kernel_argument_kind::unused, 0, 0};
@@ -250,7 +267,6 @@ public:
         for (auto& arg : args) {
             arg.pos = pos++;
         }
-        m_reqd_work_group_sizes[name] = {0, 0, 0};
     }
 
     void add_kernel_argument(const std::string& name, kernel_argument&& arg) {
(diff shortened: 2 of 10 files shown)


src/program.hpp
```diff
@@ -261,13 +273,32 @@ class spir_binary {
...
+
 private:
     spv_context m_context;
     std::vector<uint32_t> m_code;
     std::vector<sampler_desc> m_literal_samplers;
     std::unordered_map<pushconstant, pushconstant_desc> m_push_constants;
     std::unordered_map<spec_constant, uint32_t> m_spec_constants;
     image_metadata_map m_image_metadata;
+    std::unordered_map<uint32_t, printf_descriptor> m_printf_descriptors;
+    std::unique_ptr<printf_buffer_desc_info> m_printf_buffer_info;
```
kpet, line 301: This doesn't really need to be dynamically allocated, does it? It's not really a performance concern given the path it's happening on but I thought I'd ask in case I missed a good reason for it.

src/queue.cpp
```diff
@@ -787,6 +787,20 @@ cvk_command_kernel::build_batchable_inner(cvk_command_buffer& command_buffer) {
...
+    // The queue has a cvk_printf_handler which will hold on to the cvk_mem of
+    // the printf buffer and process it after the queue has finished
+    if (m_kernel->program()->uses_printf()) {
+        // Zero the printf buffer
+        auto buffer = m_argument_values->get_printf_buffer();
+        std::vector<char> zeroes = {};
+        size_t size = buffer->size();
+        zeroes.resize(size);
+        memset(zeroes.data(), 0, size);
+        if (buffer->copy_from(zeroes.data(), 0, size))
```
kpet, line 799: Any reason you're not using memset directly on a host-mapped pointer to the buffer's memory? The way it's done looks very inefficient. Taking a step back: do you really need to zero the entire buffer? With the default size, we're talking about 1MB's worth of zeroing for each kernel instance! Am I missing something?

kpet (changes requested):
Many thanks for this contribution :). I'm starting to wrap my head around it and already have a few comments. I still need to do a full review but thought I'd share what I've found so far.

src/program.hpp
callumfare: It doesn't, I think my original thought here was that `m_printf_buffer_info` is optional so it may or may not be allocated for a program. I've updated this, now the buffer size defaults to 0 which is used to represent the absence of printf buffer info.

src/queue.cpp
callumfare: No good reason for not using memset on the mapped memory directly, I've fixed that now. An older version did require the whole buffer to be zeroed, but that's not the case any more. I've updated this to only zero the first 4 bytes (i.e. the allocation counter).

src/kernel.cpp
```diff
@@ -224,6 +224,31 @@ bool cvk_kernel_argument_values::setup_descriptor_sets() {
         descriptor_writes.push_back(writeDescriptorSet);
     }
 
+    if (m_entry_point->program()->uses_printf()) {
+        if (!create_printf_buffer()) {
```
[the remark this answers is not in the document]
callumfare: It definitely isn't optimal, I did it this way to closely match the existing implementation for program-scope constant data, since notionally the printf buffer is just another program-scope buffer (although not constant). This did get complicated by the possibility of different kernel instances writing to the same buffer, so it doesn't live in the `cvk_program`, but I'm not 100% sure that what I've done is always going to be correct.

Before I make any changes here, do you have any ideas about the best way to handle the buffer? I suspect it makes the most sense for there to be one per program per queue, although if there are multiple kernels in-flight then should the `cvk_executor_thread` actually own the buffer?

src/queue.cpp
kpet: Looks good, resolving.

src/kernel.cpp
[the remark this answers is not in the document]
kpet: There's a spectrum of possible options, with different trade-offs. Here are two options with some of the pros and cons:
1. One buffer per kernel instance, i.e. what this PR currently implements. With this option we definitely don't have to worry about concurrent accesses from multiple kernel instances and only concurrency between the invocations of a single kernel instance has to be managed. However this requires allocating one buffer per kernel instance that uses printf which is rather costly. We could add a pool of printf buffers to amortise the cost of those allocations but then we'd have the problem of managing the recycling of the memory (unless we kept as many buffers as the peak number of kernel instances reached; with 1K kernels in flight and a default buffer size of 1MiB, we're already talking about 1GiB of memory). It is possible to restrict the allocation of buffers to those kernels instances that do use printf which would limit the cost but could still result in undesirable pathological cases (e.g. lots of kernel instances making very little use of printf).
2. One buffer per queue. The allocation troubles essentially vanish but we now have to manage concurrent accesses from multiple kernel instances. clvk only supports in-order queues for now but a single kernel instance can result in multiple Vulkan compute dispatches that do run concurrently. Having said that, I think this could be managed correctly with the same approach currently (see clspv PR) used to manage concurrency between the work-items of a kernel instance provided the atomics use a `Device` scope. Alternatively clvk could serialise compute dispatches that make use of the printf buffer. This would also simplify the management of the completion logic to format and print those strings in clvk (only one buffer has to be considered). Maybe we'd want to increase the default size of the buffer a bit, at least until OpenCL standardises additional control over it like in the cl_arm_printf extension for example. The buffer could even be created lazily the first time a kernel that makes use of printf is enqueued (if we do that I'd quite like to see a bitfield declaring features with runtime impact individual kernels use added to the non-semantic instruction set).

Assuming I haven't made a mistake in the above reasoning, I can't really think of a clear downside to going for 2 right now (though it's Sunday evening :-)). Can you think of other issues with this approach?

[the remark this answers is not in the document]
callumfare: The second option definitely makes the most sense to me, so I've made some changes based on that.

- The printf buffer is now in the queue
- The printf buffer is lazily created when the first kernel that may use printf is enqueued
- Additional pipeline barriers are used to serialize execution of different region dispatches for kernels that use printf
- Kernels that use printf are considered non-batchable. This prevents the case where multiple kernels from different programs try to use printf in the same batched command group, as they may both write to the printf buffer and interfere with each other. 
- The actually printing happens as soon as the command is finished.

I agree with your idea of adding a bitfield to the non-semantic instruction set to allow things to work on a per-kernel basis. Are there any other flags you think should be added other than a 'uses printf' one? I can add any others to my SPIRV-Headers change in the same PR. There are a couple of TODOs in the latest commits that will get resolved when I make those changes.

callumfare:
I've updated this PR. As with the clspv change, it's based on top of the physical module constants change (#481) to make merging easier, so only the top commit is relevant here. It can be rebased when that change goes in. It also requires the SPIRV-Headers module bump from that PR.

callumfare:
I've updated/rebased this. I believe everything on the clvk side is finished, but the matching clspv changes are currently blocked by a regression after the latest rebase. Once the clspv PR is sorted and merged this should be good to go.

rjodinchr:
I am taking over this PR

lpavank:
@rjodinchr, Gowtham from Samsung did make progress (which is being internally reviewed) and plan to upload what we got sometime this week. Is that OK or still want to take over?

kpet:
Superseded by #565. @callumfare Many thanks for all the great work! You got us 99% there :).

PR #422, opened 2022-09-13
allow main thread to exec instead of executor when calling clFinish

rjodinchr:
Explicit inorder dependency between commands in order to allow main thread to steal commands from the executor when main thread reach a clFinish call, a clWaitForEvents call or a blocking enqueue command.

For clWaitForEvents, execute instead of executor only if all events are associated to one unique queue.

extract_cmds_dominated_by has a boolean to extract only non-batch commands. It will be very important when clvk will be able to use timeline semaphore.

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -1162,8 +1162,8 @@ cl_event CLVK_API_CALL clCreateUserEvent(cl_context context,
         }
     }
 
-    auto event = new cvk_event(icd_downcast(context), CL_SUBMITTED,
-                               CL_COMMAND_USER, nullptr);
+    auto event = new cvk_event(icd_downcast(context), nullptr, nullptr);
+    event->set_status(CL_SUBMITTED);
 
     if (errcode_ret != nullptr) {
         *errcode_ret = CL_SUCCESS;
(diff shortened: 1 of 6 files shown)


kpet (commented):
I've just had a first look at this. We will definitely need a fast path to execute commands directly in threads calling synchronisation APIs to remove thread hops between application threads calling those APIs and executor threads. As discussed, I'd started writing a change in parallel to yours. I can see a couple of issues with the approach in this PR:
- It duplicates a block of code from the executor
- It doesn't work with `clWaitForEvents` or blocking enqueue commands.

rjodinchr:
Do you prefer submitting yours, or should I update this one with your feedback?

kpet:
I don't feel too strongly about this. You're likely to be faster than me so if landing this quickly(ish) matters to you, it'd probably be better for you to work on it. If not, I could take a crack at it.

rjodinchr:
Alright, I will submit an updated version of it in the coming days

rjodinchr:
no regression on the full cts on swiftshader and nvidia

rjodinchr:
no regression on the full cts with swiftshader and nvidia

rjodinchr:
As we discussed, let's focus first on merging #424.
If you want a clearer view of this PR (without the diff of #424): https://github.com/rjodinchr/clvk/pull/3

rjodinchr:
I'll reopen one when #424 will be merged

PR #425, opened 2022-09-21
enable usage of perfetto with clvk

rjodinchr:


rjodinchr:
It would be great to have a test for it in the CI, I'll create a ticket, I am not sure if I will find the time to do it soon.

kpet:
Thanks for this PR. I'll have a bit of a play before reviewing it.

>also mention CLVK_MAX_CMD_{GROUP|BATCH}_SIZE in the readme

Maybe you could extract that into a separate PR that we could land very quickly. It really should have been in the previous PR as requested in review comments ;).

rjodinchr:
Just a thought here:
I think the trace macros introduced here could have a second backend using #281.

kpet:
>I think the trace macros introduced here could have a second backend using https://github.com/kpet/clvk/pull/281.

Yup. We might even want to allow both at the same time.

rjodinchr:
> Yup. We might even want to allow both at the same time.

Why not. At the moment perfetto is a complete overset of #281, obviously not with the same UI.
Do you have a use case in mind? Or maybe you foresee #281 having other features in the features that may not be covered by perfetto.

src/traces.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include "traces.hpp"
+#include "config.hpp"
+#include "queue.hpp"
+
+#ifdef CLVK_PERFETTO_ENABLE
+
+PERFETTO_TRACK_EVENT_STATIC_STORAGE();
+
+#ifdef CLVK_PERFETTO_BACKEND_INPROCESS
+std::unique_ptr<perfetto::TracingSession> tracing_session;
```
kpet, line 24: Can this be `static`? Maybe use the `gVarName` convention for that one?

src/traces.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+#include "cl_headers.hpp"
+#include "log.hpp"
+#include "perfetto.h"
+
+PERFETTO_DEFINE_CATEGORIES(
+    perfetto::Category("clvk").SetDescription("CLVK Events"));
+
+void init_tracing();
+void term_tracing();
```
kpet, line 27: I I'd rather we always define the prototypes have have an empty implementation by default.

src/traces.hpp
```diff
@@ -0,0 +1,67 @@
...
+#pragma once
+
+#ifdef CLVK_PERFETTO_ENABLE
+
+#include "cl_headers.hpp"
+#include "log.hpp"
+#include "perfetto.h"
+
+PERFETTO_DEFINE_CATEGORIES(
+    perfetto::Category("clvk").SetDescription("CLVK Events"));
```
kpet, line 24: We will probably want to tweak that later on but fine as a starting point. Maybe a macro would be useful for the `"clvk"` category string?

src/traces.hpp
```diff
@@ -0,0 +1,67 @@
+// Copyright 2022 The clvk authors.
```
kpet, line 1: Maybe rename this file to `tracing.hpp`?

kpet:
>Why not. At the moment perfetto is a complete overset of https://github.com/kpet/clvk/pull/281, obviously not with the same UI.
>Do you have a use case in mind? Or maybe you foresee https://github.com/kpet/clvk/pull/281 having other features in the features that may not be covered by perfetto.

I like that it could be used without a dependency on Perfetto. It probably would make it easier to aggregate and display global stats as well. This could be also be done with tools processing Perfetto traces I'm sure but maybe a bit harder to integrate direclty into clvk. We'll see :).

src/queue.cpp
```diff
@@ -43,6 +44,16 @@ cvk_command_queue::cvk_command_queue(
     if (properties & CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE) {
         cvk_warn_fn("out-of-order execution enabled, will be ignored");
     }
+
+    TRACE_CNT_VAR_INIT(batch_in_flight_counter,
+                       "clvk-queue_" + std::to_string((uintptr_t)this) +
```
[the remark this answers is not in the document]
rjodinchr: I agree it would be better.
The idea here is to be able to easily link the counter with other events with the queue in argument. Unfortunately perfetto viewer prints integer in decimal format. So it would make it difficult to follow at the moment to put it here in hex.
I don't see any way of forcing perfetto to print integers in hex at the moment...

rjodinchr:
I have improved traces of the API calls with more information that can be very useful.

src/queue.cpp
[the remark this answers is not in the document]
kpet: Ah right. Agree that consistency is better.

kpet (commented):
Thanks for all the changes. I'm not sure about the new way of managing api call traces you introduced at the same time. I can see the value of having additional data in the traces but it really feels of something we should integrate with the logging macros so we don't have to maintain both (and don't have to manage manual casts everywhere).

rjodinchr:
How do you want to work from here, should I revert the last commit, and open a new PR for it once this one is merged?
Should we discuss here a way forward with that?

To me, logging and tracing are different, both in usage and expectation.
It feels normal that they do not have the same information in arguments.
Merging them feels that we are going to either put useless information in the trace, or remove potential valuable information from the log.

kpet:
I'll merge it as is. You've done the work to add all the tracing macros and they don't really get in the way of anything. Let's keep discussing what to do going forward but no need to hold up this PR.

kpet (approved):

PR #428, opened 2022-09-22
Add -hack-mul-extended option for intel devices

omarahmed1111:
Intel devices fail with `OpUMulExtended` instruction. So, we should use `-hack-mul-extended` to avoid failures on Intel devices. This fixes `integer_ops/integer_mul_hi`, `integer_ops/integer_mad_hi` and `integer_ops/integer_mad_sat` CTS tests on Intel devices.

### DIFF
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -777,6 +777,9 @@ std::string cvk_program::prepare_build_options(const cvk_device* device) const {
     if (device->supports_ubo_stdlayout()) {
         options += " -std430-ubo-layout ";
     }
+    if (strstr(device->name(), "Intel") != nullptr) {
+        options += " -hack-mul-extended ";
+    }
 
     // Select target SPIR-V version
     options += " -spv-version=";


kpet (approved):
LGTM, thanks! I'm starting to think we're going to need a way to centralise and solidify how vendor detection works but this is perfectly fine for now.

@rjodinchr I know you care about Intel devices. Are you fine with this change?

rjodinchr (approved):
> LGTM, thanks! I'm starting to think we're going to need a way to centralise and solidify how vendor detection works but this is perfectly fine for now.
> 
> @rjodinchr I know you care about Intel devices. Are you fine with this change?

I am.
I have added this option to `clspv` for that.

I agree that having a way to centralise vendors specificity would be a plus.

kpet:
>I have added this option to clspv for that.

I was asking you to confirm because you hadn't enabled the option in clvk so I was wondering whether there was an issue with having it on by default.

>I agree that having a way to centralise vendors specificity would be a plus.

I'll have a think as to how we could do that and link it to the config code.

Issue #391, opened 2022-07-28
Support CTS compiler tests involving link


rjodinchr:
Most of the CTS compiler tests `clvk` does not support involve the `clLinkProgram` function.

At the moment, `clvk` implementation is based on the spirv-tools, using SPIRV as the intermediate language to link multiple programs. It feels quite complicated to have it working as the OpenCL intends it to.

Instead of using SPIRV as the intermediate language, I propose to just simply use the LLVM IR.
It means that we would need to be able to link LLVM IR. We could do that using llvm tools, or just have it integrated inside `clspv`.

I think in the end, it would be a better and easier to maintain solution. `clspv` often needs to have all the information about the kernel (and all the subfunctions) to be able to generate correct SPIRV code. That way, we would ensure that `clspv` always has that information.

- [ ] clCompileProgram needs to produce LLVM IR
- [ ] clLinkProgram needs to link LLVM IR
- [ ] CL_PROGRAM_BINARIES could either be a LLVM IR program (which would need to be compile with `clspv` to be executed), or a SPIRV program that is "ready" to be executed.

omarahmed1111:
This is a suggested alternative way:
* Add an optional flag to clspv that says it's safe to add the Linkage capability to the SPIR-V and then add function declarations with Internal/External linkage etc as expected
* clvk consumes the SPIR-V and passes it to the spv linker which will do proper linking
* clvk then verifies that there are no dangling links, then goes through the module manually to strip the OpCapability Linkage and any other related decorations
* when enqueued the binary is sent to the Vulkan driver with no traces of linkage to be found so it's considered a valid binary

I thought of adding it for maybe if LLVM IR method had any unexpected problems, or undesirable for some reason, or if you think this could be a better choice.

rjodinchr:
Yes, I have also considered this solution.
I think the actual code is closer to this solution than the one I have proposed above.

The issue is that `clspv` often needs to have all the information to know which function to inline in order to be able to generate the correct spirv code.
I think the solution you suggested might work for the CTS, but those are really simple kernels. I am not sure it will support real applications using this feature.

PR #429, opened 2022-10-05
add full support for clLinkProgram

rjodinchr:
Use LLVM IR as the intermediate language
clCompileProgram generated a LLVM IR binary.
clLinkProgram takes LLVM IR binaries as inputs and generates a SPIRV binary.
cvk_program stores the LLVM IR in the new argument 'm_ir'.

Based on google/clspv#932
Ref #391

### DIFF
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -720,6 +720,7 @@ std::string cvk_program::prepare_build_options(const cvk_device* device) const {
         {"-cl-strict-aliasing", ""},
         // clspv require entrypoint inlining for OpenCL 2.0
         {"-cl-std=CL2.0", "-cl-std=CL2.0 -inline-entry-points"},
+        {"-create-library", ""},
     };
 
     for (auto& subst : option_substitutions) {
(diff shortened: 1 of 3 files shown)


rjodinchr:
This is passing all 'test_compiler' of the CTS using a swiftshader backend (except 'pragma_unroll' which needs int8 support).
'features_macro' pass for CL1.2, but not for CL3.0, unrelated to link programs.

I'll add some more tests in clvk soon in this PR.

rjodinchr:
This PR is ready for review, but it needs https://github.com/google/clspv/pull/932.
I'll leave it as a draft for the time being.

src/program.hpp
```diff
@@ -539,20 +539,124 @@ struct cvk_program : public _cl_program, api_object<object_magic::program> {
...
+        }
+        if (version != clvk_binary_version) {
+            cvk_warn_fn("wrong version");
+            return CL_PROGRAM_BINARY_TYPE_NONE;
+        }
+        return binary_type;
+    }
+
+public:
     CHECK_RETURN bool read(const unsigned char* src, size_t size) {
```
kpet, line 598: Maybe this function and a couple of others are a bit big for the header file. I know there are existing big-ish functions in headers but I'd like to change this where there are no performance or other considerations.

tests/azure/steps-run-tests.yml
```diff
@@ -47,6 +47,15 @@ steps:
...
       condition: ${{ parameters.compilerAvailable }}
+    - script: |
+        set -ex
+        ./build/clspv -o simple.bc --emit-binary-ir ./tests/simple-from-llvm-ir-binary/simple.cl
+        export VK_ICD_FILENAMES=${{ parameters.icd_json_unix }}
+        export CLVK_LOG=3
+        ./build/simple_test_from_llvm_ir_binary
+        ./build/simple_test_from_llvm_ir_binary_static
+      displayName: Offline IR compilation simple tests
+      condition: ${{ parameters.compilerAvailable }}
```
kpet, line 58: Wrong CI files :). We've switched to Github Actions months ago but I hadn't removed the Azure Pipelines files. I'm doing this with https://github.com/kpet/clvk/pull/431. Sorry for the confusion.

kpet (commented):
Thanks! A couple comments after a cursory first look.

rjodinchr:
I have divided this PR into 3 parts. This one, #432 and https://github.com/rjodinchr/clvk/pull/1 (I'll submit it once this one has been merged)

src/program.hpp
```diff
@@ -653,6 +664,11 @@ struct cvk_program : public _cl_program, api_object<object_magic::program> {
     std::mutex m_lock;
     std::unique_ptr<std::thread> m_thread;
     std::string m_source;
+#ifndef CLSPV_ONLINE_COMPILER
+    std::vector<uint8_t> m_ir;
+#else
+    std::string m_ir;
+#endif
```
kpet, line 671: I don't think we need two different definitions here. Maybe handling vectors in the file saving utilities would be easier?

src/program.cpp
```diff
@@ -828,77 +829,124 @@ std::string cvk_program::prepare_build_options(const cvk_device* device) const {
...
+            cmd += clspv_input_file;
+            cmd += " ";
         }
     }
 
-    // Select operation
-    // TODO support building a library with clBuildProgram
-    if (m_operation == build_operation::compile) {
-        m_binary_type = CL_PROGRAM_BINARY_TYPE_COMPILED_OBJECT;
+    std::string spirv_file{tmp_folder + "/compiled.spv"};
```
kpet, line 911: We should probably change the name depending on the output format.

src/program.cpp
```diff
@@ -962,142 +987,126 @@ cl_build_status cvk_program::compile_source(const cvk_device* device) {
...
+        } else {
+            if (m_source.empty() && !m_ir.empty()) {
+                programs.push_back(m_ir);
+            } else {
+                programs.push_back(m_source);
+            }
+        }
+        m_build_log.clear();
+        if (build_to_ir) {
+            std::vector<uint32_t> ir;
```
kpet, line 1010: Looking at the whole chain from clspv to here, there are lots of conversions between uin32_t, steams, strings, etc. We should look at streamlining this a bit more.

src/program.hpp
```diff
@@ -634,8 +634,19 @@ struct cvk_program : public _cl_program, api_object<object_magic::program> {
...
+#if COMPILER_AVAILABLE
+#ifndef CLSPV_ONLINE_COMPILER
+    CHECK_RETURN cl_build_status
+    build_source_offline(bool build_to_ir, bool build_from_il,
+                         std::string& build_options, std::string& tmp_folder);
+#else
+    CHECK_RETURN cl_build_status build_source_online(
+        bool build_to_ir, bool build_from_il, std::string& build_options);
+#endif
+#endif
```
kpet, line 648: I like how you split what was becoming a monster into online and offline functions but I don't think `build_source*` is a good name any more (the naming problem existed before this change).

kpet (approved):
This could be polished a bit more but, in the grand scheme of things these are minor issues, this change is a really nice step forward in terms of functionality and it's been in review for a while already :).

src/program.cpp
```diff
@@ -828,77 +829,124 @@ std::string cvk_program::prepare_build_options(const cvk_device* device) const {
...
+        cvk_error_fn("failed to compile the program");
+        return CL_BUILD_ERROR;
+    }
+
+    if (build_to_ir) {
+        std::ifstream stream(spirv_file, std::ios::in | std::ios::binary);
+        m_ir.assign((std::istreambuf_iterator<char>(stream)),
+                    std::istreambuf_iterator<char>());
+        cvk_info("Loaded IR binary from \"%s\", size = %zu", spirv_file.c_str(),
+                 spirv_file.size());
```
kpet, line 928: This is reporting the binary's size as the length of the name of the file it was loaded from :)

PR #430, opened 2022-10-06
refactor management of push constants

rjodinchr:
Avoid calling vkCmdPushConstants multiple times by managing push constants in a user space buffer.
Flush the buffer by calling vkCmdPushConstants one time before vkCmdDispatch.

### DIFF
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -1528,7 +1528,7 @@ cl_int cvk_entry_point::init() {
     }
 
     // Calculate POD buffer size and update the push constant range.
-    VkPushConstantRange push_constant_range = m_program->push_constant_range();
+    VkPushConstantRange* push_constant_range = m_program->push_constant_range();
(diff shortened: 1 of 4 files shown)


rjodinchr:
Regarding our conversation about the copy of `VkPushConstantRange` (changed to a pointer in this PR):

I did that because I wanted to be able to create my buffer for the push constant with the right size at the beginning.
The VkPushConstantRange stored in the kernel does not have the right size because of the copy we are doing in `cvk_entry_point::init` (I think), where we are potentially increasing the size, but without keeping the value in the VkPushConstantRange object stored in the kernel.
Of course if we change the implementation to use an upper-bound constant value, and have a dynamic allocation as in llvm SmallVector, we would not care about that.

rjodinchr:
Let's close it for now. We can go back to it if we really feel the need for it in the future.

PR #432, opened 2022-10-07
fix save string to file when a folder is missing in the path

rjodinchr:
this will help pass all compiler tests of the CTS
Ref #391

### DIFF
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -14,9 +14,11 @@
 
 #include <algorithm>
 #include <cstring>
+#include <filesystem>
 #include <iostream>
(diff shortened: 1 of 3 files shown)


rjodinchr:
I don't understand the windows failure. This PR is not changing anything that could make it fail like it does... (`llvm-spirv` command is failing to translate SPIR-V to LLVM-IR).

kpet:
I think it's because you dropped `std::ios::binary` which means files will be open as text by default and Windows will mess with data that look like line endings.

kpet (approved):
Thanks!

PR #434, opened 2022-10-10
fix temp folder content deletion

rjodinchr:
PR #432 added the possibility to create folder inside the temp folder for include purpose. But the folders created for that are not added to the temp stack. This create the unwanted behavior of leaving the temp folder on disk everytime this feature is used.

I thought of just adding it to the deletion stack, but it was complicated a bit the code, and I did not see any good reason to erase everything in the temp folder invididualy.

I propose to remove everything in the folder using filesystem::remove_all.


kpet (approved):
LGTM, thanks! (Nice to see good improvements from C++17 features!)

PR #435, opened 2022-10-11
add subgroup extended types to extension to avoid validation error

rjodinchr:


rjodinchr:
I am not sure whether we can do something else if this extension is not supported to avoid having a error hard to understand if the kernel requires this extension to execute properly.

kpet (changes requested):
Thanks! We should only report `cl_khr_subgroups` when this extension is supported as that's a hard requirement to supporting the full CL extension.

kpet (approved):
LGTM, thanks!

PR #436, opened 2022-10-12
remove large packages to fix android CI

rjodinchr:


rjodinchr:
Previous run fixed it well, just cleaning a bit, and remove even more potential large folder.

kpet (approved):
That was on my list of things to look at :). It's not great to have to resort to that but at least it gets the Android jobs going again. Thanks!

PR #437, opened 2022-10-14
Used vulkan memory model to query atomic capabilities

FMarno:
I've added a query to the Vulkan physical device to check if atomics are actually supported on the device. Atomic semantics like Acquire and Release are described as part of the Vulkan memory model so I believe that this is the best way to test that.

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -780,17 +780,28 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
         break;
     case CL_DEVICE_ATOMIC_MEMORY_CAPABILITIES:
         val_atomic_capabilities =
-            CL_DEVICE_ATOMIC_ORDER_RELAXED | CL_DEVICE_ATOMIC_ORDER_ACQ_REL |
-            CL_DEVICE_ATOMIC_SCOPE_WORK_ITEM |
-            CL_DEVICE_ATOMIC_SCOPE_WORK_GROUP | CL_DEVICE_ATOMIC_SCOPE_DEVICE;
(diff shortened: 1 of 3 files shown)


FMarno:
This is somewhat related to issue #201.

kpet (approved):
Overall, I'm not really sure this is what we want to do. A few thoughts:

- `CL_DEVICE_ATOMIC_SCOPE_WORK_ITEM` should indeed be removed for `CL_DEVICE_ATOMIC_MEMORY_CAPABILITIES`, thanks!
- clspv doesn't currently use the **VulkanMemoryModel** capability or availability/visibility operations (see https://github.com/google/clspv/issues/268).
- When the Vulkan memory model is not supported, **AcquireRelease** as well as the **Device** scope are still legal to use but the behaviour is not defined formally.
- It's not immediately obvious to me that the Vulkan memory model satisfies all the requirements of the OpenCL memory model. Checking this might be non-trivial.
- This change means we will report much fewer capabilities on a number of devices. In particular no longer reporting `CL_DEVICE_ATOMIC_SCOPE_DEVICE` is a big loss as that's a very common use for atomics. Having said that I expect most applications do not check the device capabilities before using atomics so it might not matter that much in practice.

All in all, there's one obvious fix and it's good to start having those features at our disposal to make decisions so I'll get the patch in but I bet it's definitely not the last time we will be discussing atomics :).

FMarno:
Thank you @kpet. I agree this is a non-trivial problem so thank you for taking an interest.
I'm slightly new to the Vulkan specification, but when looking at it the only details on how atomics worked was found in the Memory Model appendix, so I made a bit of an assumption there. If I find a more accurate way to query this then I'll be back!

PR #440, opened 2022-10-18
api error case fixes

anholt:
A few minor fixes from trying out clvk on lavapipe using piglit's CL tests.  I can generate a bunch more, but figured I'd see how review goes.


src/api.cpp
```diff
@@ -2537,6 +2558,13 @@ cl_int CLVK_API_CALL clSetProgramReleaseCallback(
 // Kernel Object APIs
 cl_kernel cvk_create_kernel(cl_program program, const char* kernel_name,
                             cl_int* errcode_ret) {
+    if (kernel_name == nullptr) {
+        if (errcode_ret != nullptr) {
```
kpet, line 2562: In `cvk_*` functions that correspond to an OpenCL API function, you can assume that `errcode_ret` is always non-NULL. This is a pattern I've been using to remove the need to constantly check `errcode_ret` against `nullptr`. It wasn't done consistently but I think that's what we should do for all API functions that take an `errcode_ret`. Maybe adding asserts would be useful to clearly document and check this.

src/api.cpp
```diff
@@ -1931,10 +1940,22 @@ cl_program CLVK_API_CALL clCreateProgramWithSource(cl_context context,
...
+            *errcode_ret = CL_INVALID_VALUE;
+        }
+        return nullptr;
+    }
 
     cvk_program* prog = new cvk_program(icd_downcast(context));
 
     for (cl_uint i = 0; i < count; i++) {
+        if (!strings[i]) {
```
kpet, line 1953: minor style comment: explicit comparisons against `nullptr` are preferred.

kpet (changes requested):
Thanks for the PR! One minor change please. Feel free to create other similar PRs, there's still a fair amount of API error checking missing :). I was waiting for the relevant CTS negative tests to land before adding the bulk of the missing checks but it looks like it's going to be a while.

src/api.cpp
anholt: That's the kind of thing I wanted to get feedback on before I wrote 10 more patches.  Done!

src/api.cpp
anholt: Done.

kpet (approved):
LGTM, thanks!

Issue #441, opened 2022-10-19
[CI] simple_test: UNSUPPORTED: pCreateInfo->pNext 


rjodinchr:
https://github.com/kpet/clvk/commit/a55be50a0e8f5a2ea855d8832c7989cb385f246b introduces this error message when executing `simple_test` over swiftshader in the CI:
```
/home/runner/work/clvk/clvk/swiftshader/src/Vulkan/libVulkan.cpp:1094 WARNING: UNSUPPORTED: pCreateInfo->pNext sType = 1000177000
/home/runner/work/clvk/clvk/swiftshader/src/Vulkan/libVulkan.cpp:1094 WARNING: UNSUPPORTED: pCreateInfo->pNext sType = 1000082000
```
https://github.com/kpet/clvk/actions/runs/3257323026/jobs/5348521433#step:23:648

kpet:
Swiftshader reports a version of Vulkan it doesn't fully supports it would appear. This is definitely wrong but maybe we can tweak the logic in clvk to help. Are the warnings the only symptoms you're seeing?

rjodinchr:
It blocks google infra, as those warnings are considered as errors there.
But yes, it is the only symptom I have seen.

rjodinchr:
https://swiftshader-review.googlesource.com/c/SwiftShader/+/69328

kpet:
Thanks for that, I'm kicking off another build.

kpet:
The new build has succeeded, closing.

PR #443, opened 2022-10-20
only init features/properties if ext is enabled

rjodinchr:
Some driver (like swiftshader) seems to be exposing a vulkan version but do not support all the core features yet.
To avoid error message, make sure that the extension is enabled before initializing it.

Ref #441

### DIFF
--- a/src/device.cpp
+++ b/src/device.cpp
@@ -73,10 +73,9 @@ void cvk_device::init_vulkan_properties(VkInstance instance) {
         auto corever = std::get<0>(ver_ext_prop);
         auto ext = std::get<1>(ver_ext_prop);
         auto prop = std::get<2>(ver_ext_prop);
-        if ((corever != 0) && (m_properties.apiVersion >= corever)) {
-            prop->pNext = pNext;
-            pNext = prop;
-        } else if ((ext != nullptr) && is_vulkan_extension_enabled(ext)) {
+        if (((ext != nullptr) && is_vulkan_extension_enabled(ext)) ||
+            ((ext == nullptr) && (corever != 0) &&
+             (m_properties.apiVersion >= corever))) {
             prop->pNext = pNext;
             pNext = prop;
         }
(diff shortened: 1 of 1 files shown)


src/device.cpp
```diff
@@ -73,10 +73,9 @@ void cvk_device::init_vulkan_properties(VkInstance instance) {
         auto corever = std::get<0>(ver_ext_prop);
         auto ext = std::get<1>(ver_ext_prop);
         auto prop = std::get<2>(ver_ext_prop);
-        if ((corever != 0) && (m_properties.apiVersion >= corever)) {
-            prop->pNext = pNext;
-            pNext = prop;
-        } else if ((ext != nullptr) && is_vulkan_extension_enabled(ext)) {
+        if (((ext != nullptr) && is_vulkan_extension_enabled(ext)) ||
+            ((ext == nullptr) && (corever != 0) &&
+             (m_properties.apiVersion >= corever))) {
```
kpet, line 78: This doesn't look right to me. It means we're only using core features for which clvk has no knowledge of a corresponding extension. I don't think implementations are required to report extensions that were promoted (worth double-checking though).

rjodinchr: I understand.
It is an improvement over what we had before https://github.com/kpet/clvk/commit/a55be50a0e8f5a2ea855d8832c7989cb385f246b. But it is definitely not what we would prefer (which is the actual version).
If we are to keep it that way, I don't know what I will do for google infra. I don't want to disable all swiftshader tests, but I am not sure if anything else will be possible to do easily.

kpet: Just a thought: would it be possible to actually implement those features in swiftshader?

rjodinchr: Good point, I'll look at it

rjodinchr: One of those has already been implemented upstream, we just need to rebuild swiftshader.
The other is `PhysicalDevice8BitStorageFeatures`.

kpet: I'll kick off an update of swiftshader.

rjodinchr: > Functionality in this extension is included in core Vulkan 1.2, with the KHR suffix omitted. However, if Vulkan 1.2 is supported and this extension is not, the StorageBuffer8BitAccess capability is optional. The original type, enum and command names are still available as aliases of the core functionality.
https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_8bit_storage.html#_promotion_to_vulkan_1_2

kpet: So I was right and reporting the extension is not required (at least for that one but that's a good hint that there's no blanket requirement). We can't make the change proposed in this PR then.

rjodinchr: If I understand correctly, our way forward is to add support for `VkPhysicalDevice8BitStorageFeatures` in swiftshader. But it will just report that nothing is supported, right?

kpet: Yes, that's the cheapest possible change but maybe swiftshader actually supports those features? You'd have to check (with them). BTW, I've now updated switshader in the CI, that should remove one of the failures.

rjodinchr: Update:
I was mistaken, nothing has been added upstream. I was only seeing one of them in google infra because of the stop at first error.
The other one is `PhysicalDeviceShaderFloat16Int8Features`.

Both of them are not supported in swiftshader. But I feel like they are just missing from one switch-case.
I am testing it, and will propose a change in swiftshader.

kpet: Cool, thanks for the update!

PR #445, opened 2022-10-25
Fix cyclical refcount between command queues and executors

### LINKED ISSUE #433: Memory leak caused by clEnqueueWriteBuffer
I've been trying to run the OpenCL CTS on the project but hit an out of resources error and then lead me to believe there may be a memory leak.
I've tried to create a minimal example but I've struggled to get to very small. Code at the end.

I'm going to be investigating this but I thought I would report just in case the issue was obvious.

Here is the code I'm running. After the kernel is compiled it will quick start eating up your resources! When clEnqueueWriteBuffer is removed then the memory usage stays consistent.

```
// OpenCL test
//
#define CL_TARGET_OPENCL_VERSION 120
#include <CL/cl.h>

#include <array>
#include <iostream>
#include <optional>
#include <variant>

constexpr size_t platform_choice = 0;
constexpr size_t device_choice = 0;

auto get_cl_device() -> std::variant<cl_int, cl_device_id> {
	std::array<cl_platform_id, 16> platform_ids;
	cl_uint platform_count;
	cl_int error =
	    clGetPlatformIDs(static_cast<cl_uint>(platform_ids.size()),
			     platform_ids.data(), &platform_count);
	if (error != CL_SUCCESS) {
		std::puts("GetPlatformsFailed");
		return error;
	}

	if (platform_count == 0) {
		std::puts("no platforms found");
		return 0;
	}

	std::array<cl_device_id, 16> device_ids;
	cl_uint device_count;
	error =
	    clGetDeviceIDs(platform_ids[platform_choice], CL_DEVICE_TYPE_ALL,
			   static_cast<cl_uint>(device_ids.size()),
			   device_ids.data(), &device_count);
	if (error != CL_SUCCESS) {
		std::puts("get devices failed");
		return error;
	}

	if (device_count == 0) {
		std::puts("no devices found");
		return 0;
	}
	return device_ids[device_choice];
}

struct kernel_state {
	cl_program program;
	cl_kernel kernel;
};

const char *square_src =
    "__kernel void test(__global int* input_array, __global int* "
    "output_array) {"
    "    int i = get_global_id(0);"
    "    int value = input_array[i];"
    "    output_array[i] = value * value;"
    "}";

auto create_kernel_state(cl_device_id device, cl_context ctx)
    -> std::variant<kernel_state, cl_int> {
	cl_int error;
	auto program =
	    clCreateProgramWithSource(ctx, 1, &square_src, nullptr, &error);
	if (program == nullptr) return error;

	error = clBuildProgram(program, 1, &device, nullptr, nullptr, nullptr);
	if (error != CL_SUCCESS) return error;

	auto kernel = clCreateKernel(program, "test", &error);
	if (kernel == nullptr) return error;
	return kernel_state{program, kernel};
}
auto free_kernel_state(kernel_state state) {
	clReleaseKernel(state.kernel);
	clReleaseProgram(state.program);
}
auto free_queue(cl_command_queue queue) {
	clFlush(queue);
	clFinish(queue);
	clReleaseCommandQueue(queue);
}

constexpr std::size_t local_item_size = 64;

int main() {
	auto device = std::get<cl_device_id>(get_cl_device());
	// create buffer
	using arr_ty = std::int32_t;
	std::array<arr_ty, 1024> input_array;

	cl_int error;
	auto ctx =
	    clCreateContext(nullptr, 1, &device, nullptr, nullptr, &error);
	if (ctx == nullptr) return error;

	auto maybe_square_k_state = create_kernel_state(device, ctx);
	auto k_state = std::get<kernel_state>(maybe_square_k_state);

	auto input_buffer = clCreateBuffer(
	    ctx, CL_MEM_READ_ONLY, sizeof(input_array), nullptr, &error);
	if (input_buffer == nullptr) return error;

	for (int i = 0; true; ++i) {
		auto command_queue =
		    clCreateCommandQueue(ctx, device, 0, &error);
		if (command_queue == nullptr) return error;

		// Fill input buffer
		error = clEnqueueWriteBuffer(command_queue, input_buffer,
					     CL_TRUE, 0, sizeof(input_array),
					     &input_array, 0, nullptr, nullptr);
		if (error != CL_SUCCESS) return error;

		// release
		free_queue(command_queue);
		std::cout << i << '\r';
		std::cout.flush();
	}
	clReleaseMemObject(input_buffer);
	free_kernel_state(k_state);
	clReleaseContext(ctx);
}
```

kpet:
Since #421, command queue executors hold a refcount on the last command queue they were associated with. Since executors are pooled and not destroyed until the context is destroyed, they keep all command queues that ever were created alive, which in turn means that executors are never returned to the pool.

After this change executors only hold that refcount temporarily during the execution of individual command groups.

Fixes #433

### DIFF
--- a/src/queue.cpp
+++ b/src/queue.cpp
@@ -259,7 +259,6 @@ cl_int cvk_command_queue::wait_for_events(cl_uint num_events,
 }
 
 void cvk_executor_thread::set_queue(cvk_command_queue* queue) {
-    m_queue.reset(queue);
     m_profiling = queue->has_property(CL_QUEUE_PROFILING_ENABLE);
 }
 
@@ -286,6 +285,9 @@ void cvk_executor_thread::executor() {
 
         lock.unlock();
 
+        CVK_ASSERT(group->commands.size() > 0);
+        cvk_command_queue_holder queue = group->commands.front()->queue();
+
         while (group->commands.size() > 0) {
 
             cvk_command* cmd = group->commands.front();
@@ -311,8 +313,7 @@ void cvk_executor_thread::executor() {
             delete cmd;
         }
 
-        CVK_ASSERT(m_queue);
-        m_queue->group_completed();
+        queue->group_completed();
 
         lock.lock();
     }
--- a/src/queue.hpp
+++ b/src/queue.hpp
@@ -36,8 +36,7 @@ struct cvk_command_group {
 struct cvk_executor_thread {
 
     cvk_executor_thread()
-        : m_thread(nullptr), m_shutdown(false), m_profiling(false),
-          m_queue(nullptr) {
+        : m_thread(nullptr), m_shutdown(false), m_profiling(false) {
         m_thread =
             std::make_unique<std::thread>(&cvk_executor_thread::executor, this);
     }
@@ -74,7 +73,6 @@ private:
     bool m_shutdown;
     std::deque<std::unique_ptr<cvk_command_group>> m_groups;
     bool m_profiling;
-    cvk_command_queue_holder m_queue;
 };
 
 struct cvk_command_pool {


src/queue.cpp
```diff
@@ -259,7 +259,6 @@ cl_int cvk_command_queue::wait_for_events(cl_uint num_events,
 }
 
 void cvk_executor_thread::set_queue(cvk_command_queue* queue) {
```
rjodinchr, line 261: I think we should remove this function. And get the profiling property from the queue of the first command in `cvk_executor_thread::executor`.

kpet: I thought about it but decided to keep it as it provides a better separation of concerns, i.e. the executor pool doesn't need to know that executor threads need to be associated to queues because of profiling, they just have to know an association is needed.

rjodinchr: Note that #424 removes the need of `m_profiling` in the `cvk_executor_thread` (even so it seems I forget to remove it in the PR). It would leave `set_queue` completely empty.

kpet: Alright, let's remove it under #424 then.

rjodinchr (approved):

PR #446, opened 2022-10-26
fix warnings threaded as errors in google infra

rjodinchr:


kpet (commented):
I'm seeing the same issue locally but I'm not sure about this solution. The variables are used by the `TRACE_CNT` macro so maybe it should be calling `UNUSED` when tracing is not enabled?

PR #447, opened 2022-10-29
Update deps and CTS

kpet:
Now need to ask clspv to enable the __opencl_c_read_write_images feature macro.


kpet:
@rjodinchr It seems clspv regressed some of the IL tests on Windows. I haven't bisected it but I think it's likely because of https://github.com/google/clspv/pull/932. I see the output generation logic has changed. Are we now missing a flag similar to `std::ios::binary` somewhere?

rjodinchr:
I am working on it

kpet:
@rjodinchr Thanks, all passing now.

PR #449, opened 2022-11-04
Preserve arg set status when cloning a kernel

callumfare:
This fixes test_api/clone_kernel from the CTS

### DIFF
--- a/src/kernel.cpp
+++ b/src/kernel.cpp
@@ -44,9 +44,6 @@ cl_int cvk_kernel::init() {
         m_image_metadata = md;
     }
 
-    // Mark all arguments as unset
-    m_args_set.resize(m_args.size(), false);
-
     // Init argument values
     m_argument_values = cvk_kernel_argument_values::create(m_entry_point);
     if (m_argument_values == nullptr) {
@@ -119,11 +116,6 @@ cl_int cvk_kernel::set_arg(cl_uint index, size_t size, const void* value) {
 
     cl_int ret = m_argument_values->set_arg(arg, size, value);
 
-    if (ret == CL_SUCCESS) {
-        // Mark argument as set
-        m_args_set[index] = true;
-    }
-
     // if the argument is an image, we need to set its metadata
     // (channel_order/channel_data_type).
     if (arg.kind == kernel_argument_kind::ro_image ||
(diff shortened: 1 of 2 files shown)


kpet (changes requested):
I don't think that's correct: not all arguments are necessarily set at the time where cloning happens. Could you add a test, here or, even better, in the CTS, for the following scenario:
- Create kernel object for a given kernel in a program
- Set only part of its arguments
- Attempt to enqueue it, check that `CL_INVALID_KERNEL_ARGS` is returned
- Clone the kernel
- Attempt to enqueue the cloned kernel, check that `CL_INVALID_KERNEL_ARGS` is returned

As to how to implement it, I think we should probably move `cvk_kernel::m_args_set` into the `cvk_kernel_argument_values` object and let the latter's copy constructor handle the copy of the exact state.

callumfare:
Thanks for the review, that's what happens when I try to fix something last thing on a Friday! I've updated this PR to implement it as you suggested, and I've opened a new PR for the CTS to add an additional test case for `CL_INVALID_KERNEL_ARGS` - https://github.com/KhronosGroup/OpenCL-CTS/pull/1563

src/kernel.cpp
```diff
@@ -134,6 +126,12 @@ cl_int cvk_kernel::set_arg(cl_uint index, size_t size, const void* value) {
     return ret;
 }
 
+bool cvk_kernel::args_valid() const {
+    return std::all_of(m_argument_values->args_set().cbegin(),
+                       m_argument_values->args_set().cend(),
+                       [](bool b) { return b; });
```
kpet, line 132: You could delegate this to the arguments object to reduce what it exposes to just an `args_valid` function.

kpet (approved):

kpet:
Thanks! The change looks good. I've left a minor comment in case you're up for some more polishing but I'm fine merging the change as is and will do in a day or so if you don't push an update.

src/kernel.cpp
```diff
@@ -134,6 +126,8 @@ cl_int cvk_kernel::set_arg(cl_uint index, size_t size, const void* value) {
     return ret;
 }
 
+bool cvk_kernel::args_valid() const { return m_argument_values->args_valid(); }
```
kpet, line 129: That could have been in the header but we've iterated enough on that change now :).

kpet (approved):
Thanks!

Issue #451, opened 2022-11-09
Stop reporting support for cl_khr_subgroups?


aarongreig:
Now that we're on CL 3.0 we can report support for subgroup operations without supporting the extension. The motivation for this is that Vulkan doesn't really make any forward progress guarantees at the subgroup level, and supporting subgroups as the core feature only would allow us to use `CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS` to communicate this and skip the forward progress tests in the CTS. The change is simple enough, I'm opening an issue rather than a PR because honestly I'm finding the specs involved a bit light on explicit language, so I'm interested to know if anyone thinks we can in fact guarantee independent forward progress, or if anyone has a device that passes the relevant CTS tests (I can vouch for two that don't).

rjodinchr:
We (Google) are relying on the extension to be able to trigger optimisation in [tensorflow](https://github.com/tensorflow/tensorflow/blob/b965b91fc9565d2e9d32c6cb48ea2acd254e937b/tensorflow/lite/delegates/gpu/common/tasks/conv_generic.cc#L1775) (you can also see that we have a special if statement for `clvk` there).

I have not looked too much into the CTS for this feature. Are you saying that the Vulkan subgroups feature is not enough to be able to support CL subgroups in `clvk`?

I only know that the Vulkan subgroups feature is enough to run the tensorflow lite workload we have internally.

aarongreig:
> Are you saying that the Vulkan subgroups feature is not enough to be able to support CL subgroups in clvk?

It can support subgroups as defined in the core (3.0) spec, but not as the extension defines due to the lack of `CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS` in the extension. Vulkan can support all the subgroup operations like the scan and broadcast type stuff but there isn't anything I can see that says vulkan implementations need to guarantee that subgroups can make independent forward progress like it says the CL spec. [This blog post](https://www.khronos.org/blog/comparing-the-vulkan-spir-v-memory-model-to-cs#_limited_forward_progress_guarantees) highlights the lack of a guarantee, and it specifically says user defined blocking conditions probably won't work, which is a pattern central to how the CL CTS tests independent forward progress. The blog post is obviously about the memory model, which clvk doesn't currently enable, but without the memory model enabled you're working with even fewer guarantees wrt operation ordering.

[...]

PR #454, opened 2022-11-10
Stop reporting support for cl_khr_subgroups extension.

aarongreig:
Removing this extension allows us to stop reporting CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS as true while still reporting support for core subgroup functionality. This is more in line with Vulkan subgroups, specifically Vulkan's lack of any such forward progress guarantees. See #451 for more detail.


kpet (approved):
Thanks!

PR #455, opened 2022-11-10
Rename image kernel arg kind enums.

aarongreig:
I think this makes it clearer that the enum is used for both read_write
and write_only (and that both are represented with a vulkan read/write
capable type) while adhering to the existing convention.


kpet (commented):
If we're renaming these, we probably should go for `sampled_image` and `storage_image` instead. What do you think?

aarongreig:
agreed, that looks better

kpet (approved):
Thanks!

PR #458, opened 2022-11-14
Rework binaries with a header to support all types

rjodinchr:
types: executable/compiled_objects/library.
Still supports clCreateProgramWithBinary using a SPIRV binary Also add support for a LLVM IR binary (looking at the magic at the begining of the file).

Ref #391


tests/api/compiler.cpp
```diff
@@ -212,15 +212,160 @@ TEST_F(WithCommandQueue, ModuleScopeConstantData) {
...
+    ASSERT_EQ(buffer_size, sizeof(result));
+
+    CompileProgram(program);
+    ASSERT_EQ(GetProgramBinaryType(program),
+              CL_PROGRAM_BINARY_TYPE_COMPILED_OBJECT);
+    auto compiled_binary = GetProgramBinary(program);
+    auto binary_program = CreateProgramWithBinary(compiled_binary);
+    ASSERT_EQ(GetProgramBinaryType(binary_program),
+              CL_PROGRAM_BINARY_TYPE_COMPILED_OBJECT);
+
```
kpet, line 236: We could check that the kernel creation is correctly rejected prior to calling clBuildProgram.

tests/api/compiler.cpp
```diff
@@ -212,15 +212,160 @@ TEST_F(WithCommandQueue, ModuleScopeConstantData) {
...
+    EnqueueNDRangeKernel(kernel, 1, nullptr, &gws, nullptr);
+    EnqueueReadBuffer(buffer, CL_BLOCKING, 0, buffer_size, result);
+
+    ASSERT_EQ(result[0], 0);
+    ASSERT_EQ(result[1], 1);
+    ASSERT_EQ(result[2], 2);
+    ASSERT_EQ(result[3], 3);
+}
+
+TEST_F(WithCommandQueue, ProgramBinaryExecutable) {
```
kpet, line 324: All these tests have a lot in common. It might be worth introducing a dedicated fixture.

tests/api/testcl.hpp
```diff
@@ -226,6 +223,35 @@ class WithContext : public ::testing::Test {
...
+    holder<cl_program> CreateAndBuildProgram(const char* source,
+                                             const char* options = nullptr) {
+        auto program = CreateProgram(source);
+
+        BuildProgram(program, options);
+
+        return program;
+    }
+
+    holder<cl_program> CreateProgramWithBinary(std::vector<uint8_t>& binary) {
```
kpet, line 237: `binary`  could be const.

tests/api/testcl.hpp
```diff
@@ -226,6 +223,35 @@ class WithContext : public ::testing::Test {
...
+        cl_int binary_status;
+        auto program = clCreateProgramWithBinary(m_context, 1, &gDevice, &size,
+                                                 &data, &binary_status, &err);
+        EXPECT_CL_SUCCESS(err);
+        EXPECT_EQ(binary_status, CL_BUILD_SUCCESS);
+        return program;
+    }
+
+    holder<cl_program>
+    CreateAndBuildProgramWithBinary(std::vector<uint8_t>& binary,
```
kpet, line 250: Same here.

kpet (approved):
Awesome, thanks! A few minor comments but nothing that warrants holding this PR up.

PR #460, opened 2022-11-20
group device specific properties

rjodinchr:


src/device_properties.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include "cl_headers.hpp"
+
+struct cvk_device_properties {
+    virtual cl_ulong get_global_mem_cache_size() const { return 0; }
+    virtual cl_ulong get_num_compute_units() const { return 1; }
+
+    virtual cl_uint get_max_cmd_batch_size() const { return 10000; }
+    virtual cl_uint get_max_first_cmd_batch_size() const { return 10000; }
+    virtual cl_uint get_max_cmd_group_size() const { return UINT32_MAX; }
+    virtual cl_uint get_max_first_cmd_group_size() const { return UINT32_MAX; }
```
kpet, line 29: Please use the default values from the config so we don't have default values in two places.

kpet (changes requested):
I was thinking I'd extend the config management to do this, possibly by adding support for config sections and/or profiles, but this approach works too and is definitely a step in the right direction. We can always rework it later, which will have been made much easier by the abstraction you're introducing. Please fix the source of the default values and it's good to go. Thanks!

src/device_properties.hpp
rjodinchr: Done

kpet (approved):
Thanks!

PR #461, opened 2022-11-23
enable CapabilityShaderNonUniform when using appropriate vk version

rjodinchr:


kpet (approved):
Thanks! This is correct but we could also accept the capability when VK_EXT_descriptor_indexing is supported, I'll add that.

PR #469, opened 2022-11-28
allow to override where temporaries files are saved

rjodinchr:


src/config.def
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 OPTION(std::string, cache_dir, "")
+OPTION(std::string, build_cache_dir, "")
```
kpet, line 16: Not really a "cache" since we never really reuse data from there. It's always used as a one-off. If we recompile 50 times the same program with the same options, we'll create 50 temp folders. Maybe `compiler_temp_dir` or better name?

src/program.cpp
```diff
@@ -1187,9 +1187,12 @@ cl_build_status cvk_program::do_build_inner(const cvk_device* device) {
     std::string tmp_folder;
     if (use_tmp_folder) {
         // Create temporary folder
-        std::string tmp_template{"clvk-XXXXXX"};
+        std::filesystem::path tmp_prefix(config.build_cache_dir());
+        std::filesystem::path tmp_suffix("clvk-XXXXXX");
+        std::string tmp_template = (tmp_prefix / tmp_suffix).string();
```
kpet, line 1192: Doesn't that expand to `/clvk-XXXXX`? Is it really using the current directory in that case? You were using the temp dir as determined by std::filesystem in a previous patch. What was the problem with that?

src/program.cpp
```diff
@@ -1187,9 +1187,12 @@ cl_build_status cvk_program::do_build_inner(const cvk_device* device) {
...
         const char* tmp = cvk_mkdtemp(tmp_template);
         if (tmp == nullptr) {
+            cvk_error_fn("Could not create temporary folder");
```
kpet, line 1195: Maybe give the path for which creation failed in the error message.

kpet (changes requested):
Thanks! I've thought about doing this a few times :).

src/program.cpp
rjodinchr: I am not sure, but my understanding is that this "/' is just a concatenate. In the case of an empty first string, I think it is considered as the current directory (to be confirmed).

Initially, I wanted to have the default temp directory to be `std:filesystem::temp_directory_path` (meant to be `/tmp` on linux). And use `std::filesystem::current_directory` if `build_cache_dir` ( :p `compiler_temp_dir`) is not set and `CLVK_KEEP_TEMPORARIES` is set, to have it right where you are.
But windows with the online compiler kept failing. I don't have a windows setup to reproduce. I tried a few things and it seems that `clspv` was not happy with the include `-I C:\dirA\dirB\dirB` and/or the include in the source `#include dir/header.h` which are using different separators.

kpet: I tested and it is a bit smarter than a simple concatenation. It removes the trailing slash when `tmp_prefix` is an empty string and the current directory is used. Ah right, I don't think it's strictly needed: the temporaries only stick around when the user requests it for debugging so I'm not too fussed about the default path not being under the system's temporary folder. The separator issue on Windows might require some form of canonicalisation of paths. We can chase that later as the need arises.

rjodinchr: I wanted to have the default be the temporary directory because I encounter issues in the chromeOS infra with the current folder not being writable.
Of course I have many other ways to fix that, but this one seemed like a proper one matching my needs.
I'll fall back onto another one.
Thanks for merging this one.

PR #470, opened 2022-11-28
decorate non uniform pointers when possible

rjodinchr:


rjodinchr:
`clspv` implementation is breaking something. Most probably in `selectFromSameObject` that is now used more when enabling this feature.

rjodinchr:
https://github.com/google/clspv/pull/967

src/device.cpp
```diff
@@ -845,9 +845,7 @@ bool cvk_device::supports_capability(spv::Capability capability) const {
     case spv::CapabilityVulkanMemoryModel:
         return m_features_vulkan_memory_model.vulkanMemoryModel;
     case spv::CapabilityShaderNonUniform:
-        return m_properties.apiVersion >= VK_MAKE_VERSION(1, 2, 0) ||
-               is_vulkan_extension_enabled(
-                   VK_EXT_DESCRIPTOR_INDEXING_EXTENSION_NAME);
+        return supports_non_uniform_pointers();
```
kpet, line 848: They're not really non-uniform "pointers". Maybe `supports_non_uniform_decoration`?

kpet (changes requested):
One name change please. Also this conflicts with https://github.com/kpet/clvk/pull/475 I've just merged, you'll need to rebase.

src/device.cpp
rjodinchr: Done

kpet (approved):
Thanks!

src/device.hpp
```diff
@@ -270,6 +270,12 @@ struct cvk_device : public _cl_device_id,
 
     bool supports_subgroups() const { return m_has_subgroups_support; }
 
+    bool supports_non_uniform_decoration() const {
+        return m_properties.apiVersion >= VK_MAKE_VERSION(1, 2, 0) ||
+               is_vulkan_extension_enabled(
+                   VK_EXT_DESCRIPTOR_INDEXING_EXTENSION_NAME);
+    }
+
```
kpet, line 278:
```suggestion

```

PR #471, opened 2022-11-28
align atomic feature macros to api query

FMarno:

Sorry I missed this when doing the previous PR. This change should make the feature macro list line up with more with the API query as seen here
https://github.com/kpet/clvk/blob/782818384cec4d09cc6b3a75396959475572e1af/src/api.cpp#L791-L797


kpet (commented):
Thanks, I did think about this and then had forgotten by the time you sent the other PR. Maybe we should wrap the tests in cvk_device member functions to guarantee the conditions used here and to report device infos are the same?

FMarno:
> Maybe we should wrap the tests in cvk_device member functions to guarantee the conditions used here and to report device infos are the same?

Good idea, is this what you are looking for? I see I'm not quite following the practice of creating a boolean variable, but I hoped this query was simple enough. I can change it if you like.

kpet (approved):
Exactly what I had in mind, thanks!

PR #474, opened 2022-11-29
fix device_properties + add amd

rjodinchr:
strcmp if statement was incorrectly moved by previous commit use explicit comparison for strcmp
use strncmp instead of strstr for Intel

overriden functions were not used because of copy of struct. Use a unique_ptr instead.


kpet (approved):
I should have been more careful in the review :).

PR #475, opened 2022-11-29
Take advantage of clspv's --use-native-builtins flag

aarongreig:
Also refactor build option handling so that options which don't vary per program are only set up once in the device rather than for each program built.

Depends on https://github.com/kpet/clvk/pull/474 (I'll pull the fixes soon)

### DIFF
--- a/src/device.cpp
+++ b/src/device.cpp
@@ -363,6 +363,100 @@ void cvk_device::init_features(VkInstance instance) {
     m_features.pNext = pNext;
 }
 
+void cvk_device::init_compiler_options() {
+    m_device_compiler_options = "";
+
+    if (!devices_support_images()) {
+        m_device_compiler_options += " -images=0 ";
+    }
+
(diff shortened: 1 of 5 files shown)


src/device.cpp
```diff
@@ -363,6 +363,99 @@ void cvk_device::init_features(VkInstance instance) {
...
+    }
+    if (supports_int8()) {
+        m_device_compiler_options += " -int8 ";
+    }
+    if (supports_ubo_stdlayout()) {
+        m_device_compiler_options += " -std430-ubo-layout ";
+    }
+
+    // Device specific options
+    m_device_compiler_options += m_clvk_properties->get_compile_options();
```
rjodinchr, line 412: `get_compile_options` does not guarantee to return a white-space surrounded string.

aarongreig: fixed

kpet (approved):
LGTM, thanks!

PR #478, opened 2022-12-07
Implement clSetProgramSpecializationConstant

aarongreig:

Depends on https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1758 (and a submodule bump which I will add here once that gets merged)

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -5437,7 +5437,14 @@ cl_program cvk_create_program_with_il(cl_context context, const void* il,
 
     auto program = new cvk_program(icd_downcast(context), il, length);
 
-    *errcode_ret = CL_SUCCESS;
+    if (!program->parse_user_spec_constants()) {
+        // If we fail to parse spec constants we can assume the IL stream is
+        // wonky.
+        *errcode_ret = CL_INVALID_VALUE;
+    } else {
+        *errcode_ret = CL_SUCCESS;
+    }
+
     return program;
 }
 
@@ -5476,13 +5483,20 @@ cl_program CLVK_API_CALL clCreateProgramWithIL(cl_context context,
 }
 
 cl_int CLVK_API_CALL
-clSetProgramSpecializationConstant(cl_program program, cl_uint spec_id,
+clSetProgramSpecializationConstant(cl_program prog, cl_uint spec_id,
                                    size_t spec_size, const void* spec_value) {
-    TRACE_FUNCTION("program", (uintptr_t)program, "spec_id", spec_id,
-                   "spec_size", spec_size);
+    TRACE_FUNCTION("program", (uintptr_t)prog, "spec_id", spec_id, "spec_size",
+                   spec_size);
     LOG_API_CALL("program = %p, spec_id = %u, spec_size = %zu, spec_value = %p",
-                 program, spec_id, spec_size, spec_value);
-    return CL_INVALID_OPERATION;
+                 prog, spec_id, spec_size, spec_value);
+
+    auto program = icd_downcast(prog);
+
+    if (!is_valid_program(program)) {
+        return CL_INVALID_PROGRAM;
+    }
+
+    return program->set_user_spec_constant(spec_id, spec_size, spec_value);
 }
 
 // Shared Virtual Memory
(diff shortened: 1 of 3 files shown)


aarongreig:
My llvm-spirv patch merged but only after some breaking changes that'll necessitate a clspv llvm bump, I'm not going to have time to do the testing to make that change that so I'll wait until it happens and update here when it does.

aarongreig:
I think the remaining fail might be an issue with the CI, it looks like it's failing before the build gets to anything touched by my changes.

src/program.hpp
```diff
@@ -142,6 +142,35 @@ enum class spec_constant
...
+    union {
+        uint8_t i8;
+        uint16_t i16;
+        uint32_t i32;
+        uint64_t i64;
+    } data;
+
+    user_spec_constant_data(std::string type, uint32_t size)
+        : type(type), size(size), set(false) {}
+    user_spec_constant_data() : type(""), size(0), set(false) {}
```
kpet, line 158: Why is that constructor needed? Can we remove it?

src/program.hpp
```diff
@@ -142,6 +142,35 @@ enum class spec_constant
...
+    void init_data(size_t size, const void* value) {
+        if (type == "i8" || type == "i1") {
+            std::memcpy(&data.i8, value, size);
+        } else if (type == "i16" || type == "f16") {
+            std::memcpy(&data.i16, value, size);
+        } else if (type == "i32" || type == "f32") {
+            std::memcpy(&data.i32, value, size);
+        } else if (type == "i64" || type == "f64") {
+            std::memcpy(&data.i64, value, size);
+        }
```
kpet, line 169: We should probably have an else case to catch errors here or at least an assert.

src/api.cpp
```diff
@@ -5476,13 +5483,20 @@ cl_program CLVK_API_CALL clCreateProgramWithIL(cl_context context,
...
-                 program, spec_id, spec_size, spec_value);
-    return CL_INVALID_OPERATION;
+                 prog, spec_id, spec_size, spec_value);
+
+    auto program = icd_downcast(prog);
+
+    if (!is_valid_program(program)) {
+        return CL_INVALID_PROGRAM;
+    }
+
```
kpet, line 5498: A few more errors need to be handled here. Please handle them or add TODOs:

• CL_INVALID_OPERATION if no devices associated with program support intermediate language programs.
• CL_COMPILER_NOT_AVAILABLE if program is created with clCreateProgramWithIL and a compiler is not available, i.e. CL_DEVICE_COMPILER_AVAILABLE specified in the Device Queries table is set to CL_FALSE.
• CL_INVALID_VALUE if spec_value is NULL.

src/program.cpp
```diff
@@ -894,9 +894,92 @@ std::string cvk_program::prepare_build_options(const cvk_device* device) const {
...
+        delim = line.find(",");
+        auto size_string = line.substr(0, delim);
+        line.erase(0, delim + 1);
+        size_string = size_string.substr(size_string.find("=") + 1);
+        auto size = static_cast<uint32_t>(atoi(size_string.c_str()));
+
+        // We need to consume the "=" and the " " here since we aren't
+        // converting the value from a string.
+        auto type_string = line.substr(line.find("=") + 2);
+        m_user_spec_constants[id] = user_spec_constant_data(type_string, size);
```
kpet, line 949: The lack of error handling is making me a bit nervous. Hopefully the translator does not change its output format too frequently and never produces output that would be invalid in a way that would trip this code over. Maybe a regex would be better? I think I'm fine with no change here.

kpet (changes requested):
Thanks! It's almost there, a few minor changes please (also merge conflicts need to be resolved).

kpet:
Also, I can confirm that the macOS CI error is spurious, I'll deal with this.

kpet:
@aarongreig Do you want to update the PR or should I take it over?

kpet:
@aarongreig It's gone in as https://github.com/kpet/clvk/pull/503. I think I've managed to retain your authorship. Thanks for the contribution!

aarongreig:
thanks!

PR #485, opened 2022-12-12
Add support to use a percentage of total memory 

omarahmed1111:
Add an option so that users could control the amount of total memory used by clvk on allocations.

### DIFF
--- a/src/device.hpp
+++ b/src/device.hpp
@@ -205,7 +205,14 @@ struct cvk_device : public _cl_device_id,
         heap_index = m_mem_properties.memoryTypes[type_index].heapIndex;
         size = std::min(size, m_mem_properties.memoryHeaps[heap_index].size);
 
-        return size;
+        double percentage_of_available_memory_reported =
+            static_cast<double>(
+                config.percentage_of_available_memory_reported()) /
+            100;
+        cvk_info("Using %u%% of total memory size",
+                 config.percentage_of_available_memory_reported());
+
+        return size * percentage_of_available_memory_reported;
     }
 
     uint64_t memory_size() const {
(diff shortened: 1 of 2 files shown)


src/device.hpp
```diff
@@ -205,7 +205,12 @@ struct cvk_device : public _cl_device_id,
         heap_index = m_mem_properties.memoryTypes[type_index].heapIndex;
         size = std::min(size, m_mem_properties.memoryHeaps[heap_index].size);
 
-        return size;
+        double percentage_of_available_memory =
+            static_cast<double>(config.percentage_of_available_memory) / 100;
+        cvk_info("Using %lf% of total memory size",
+                 config.percentage_of_available_memory);
+
+        return size * percentage_of_available_memory;
```
omarahmed1111, line 213: I moved the percentage here as I think `max_alloc_size()` should also address that percentage change.

kpet: Yes, that's a better location for it.

src/device.hpp
```diff
@@ -205,7 +205,12 @@ struct cvk_device : public _cl_device_id,
         heap_index = m_mem_properties.memoryTypes[type_index].heapIndex;
         size = std::min(size, m_mem_properties.memoryHeaps[heap_index].size);
 
-        return size;
+        double percentage_of_available_memory =
+            static_cast<double>(config.percentage_of_available_memory) / 100;
+        cvk_info("Using %lf% of total memory size",
+                 config.percentage_of_available_memory);
```
kpet, line 211:
```suggestion
        cvk_info("Using %u%% of total memory size",
                 config.percentage_of_available_memory());
```

Fixes build warnings

kpet (commented):
I'm not sure about the name but I can't think of something much better (maybe we could add "_reported" to the name. Let's get this in as a stepping stone.

kpet (changes requested):
Seems I can't commit the suggestion myself so: please make commit it :). Thanks!

kpet (approved):
LGTM, thanks!

PR #486, opened 2022-12-12
rework api_tests image write invalidate

rjodinchr:
The test is known to be a little flaky.
This PR is changing a bit the nature of the test.
It is checking that the map region is not written by clvk before being given to the user. But it does not check for other copies that could happen inside clvk.


kpet (approved):
Nice tidy-up. Let's see how this goes. As discussed we can think of more changes if it's still flaky.

Issue #480, opened 2022-12-08
Memory Leak in clEnqueueMapImage


FMarno:
I've observed a memory leak when using clEnqueueMapImage. Following is the reproducing code:
```cpp
#define CL_TARGET_OPENCL_VERSION 120
#include <CL/cl.h>

#include <array>
#include <iostream>
#include <optional>
#include <variant>

constexpr size_t platform_choice = 0;
constexpr size_t device_choice = 0;

auto get_cl_device() -> std::variant<cl_int, cl_device_id> {
	std::array<cl_platform_id, 16> platform_ids;
	cl_uint platform_count;
	cl_int error =
	    clGetPlatformIDs(static_cast<cl_uint>(platform_ids.size()),
			     platform_ids.data(), &platform_count);
	if (error != CL_SUCCESS) {
		std::puts("GetPlatformsFailed");
		return error;
	}

	if (platform_count == 0) {
		std::puts("no platforms found");
		return 0;
	}

	std::array<cl_device_id, 16> device_ids;
	cl_uint device_count;
	error =
	    clGetDeviceIDs(platform_ids[platform_choice], CL_DEVICE_TYPE_ALL,
			   static_cast<cl_uint>(device_ids.size()),
			   device_ids.data(), &device_count);
	if (error != CL_SUCCESS) {
		std::puts("get devices failed");
		return error;
	}

	if (device_count == 0) {
		std::puts("no devices found");
		return 0;
	}
	return device_ids[device_choice];
}

auto free_queue(cl_command_queue queue) {
	clFlush(queue);
	clFinish(queue);
	clReleaseCommandQueue(queue);
}

int main() {
	auto device = std::get<cl_device_id>(get_cl_device());
	// create buffer
	using arr_ty = std::int32_t;
	std::array<arr_ty, 100> input_array;

	cl_int error;
	auto ctx =
	    clCreateContext(nullptr, 1, &device, nullptr, nullptr, &error);
	if (ctx == nullptr) return error;

	cl_image_format imageFormat{CL_RGB, CL_SIGNED_INT32};

	cl_image_desc imageDesc{};

	imageDesc.image_type = CL_MEM_OBJECT_IMAGE1D;
	imageDesc.image_width = 100;
	imageDesc.image_height = 0;
	imageDesc.image_depth = 0;

	size_t origin[4] = {0, 0, 0, 0};
	size_t region[3] = {imageDesc.image_width, 1, 1};
	size_t offset[3] = {0, 0, 0};

[...]

PR #487, opened 2022-12-13
added missing virtual from command_batchable destructor

FMarno:
clvk_command_batchable is being inherited from so it should have a virtual destructor.


kpet (approved):
How did you find this?

FMarno:
Looking into issue #480. It didn't fix anything but I believe it's technically correct.

kpet:
I don't think it's strictly necessary but it certainly doesn't hurt :).

PR #488, opened 2022-12-20
use vendorID/deviceID to choose device properties

rjodinchr:
vendorID/deviceID are more reliable to identify which device properties to select.


kpet (approved):
LGTM, thanks!

PR #489, opened 2022-12-22
Implement clGetKernelInfo with CL_KERNEL_ATTRIBUTES

FMarno:
This depends on this change to CLSPV - https://github.com/google/clspv/pull/990, which itself depends on changes to SPIRV-Headers - https://github.com/KhronosGroup/SPIRV-Headers/pull/308.

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -2826,6 +2826,7 @@ cl_int CLVK_API_CALL clGetKernelInfo(cl_kernel kern, cl_kernel_info param_name,
     cl_int ret = CL_SUCCESS;
     size_t ret_size = 0;
     const void* copy_ptr = nullptr;
+    // TODO could this be a union
     cl_uint val_uint;
     cl_context val_context;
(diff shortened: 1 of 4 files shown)


FMarno:
This isn't in a completed state, but I'm going on holiday for two weeks. One of my colleagues may finish the work while I'm away.

FMarno:
Unfortunately I no longer have time to complete this work but it would be a shame if this work was lost since these addition will help toward 100% CTS conformance, specifically the api/kernel_attributes test. If someone would like to pick up this work and finish it then please do. First https://github.com/google/clspv/pull/990 will need to be completed but then I think all that will need to be done is to account for the attribute being optional.

rjodinchr:
I am taking over this PR

rjodinchr:
This PR can be closed in favor of #562

kpet:
@rjodinchr Ok, closing. @FMarno Thanks for setting this into motion!

Issue #479, opened 2022-12-07
Unused kernel arguments


alan-baker:
As part of the transition to opaque pointers, clspv needs to handle unused kernel arguments a bit differently. For example, consider the following kernel from the intmath basic CTS test:
```
  __kernel void test_add(__global TYPE *srcA,
                         __global TYPE *srcB,
                         __global TYPE *srcC,
                         __global TYPE *dst)
  {
      int  tid = get_global_id(0);
      dst[tid] = srcA[tid] + srcB[tid];
  };
```

Here `srcC` is unused. There is no information for clspv to generate an appropriate variable from. The compiler cannot simply generate a storage buffer because it might not be compatible with anything that is bound. For example, a modified version of the same kernel:
```
  __kernel void test_add(__global TYPE *srcA,
                         __global TYPE *srcB,
                         read_only image2d_t srcC,
                         __global TYPE *dst)
  {
      int  tid = get_global_id(0);
      dst[tid] = srcA[tid] + srcB[tid];
  };
```

The LLVM IR is the same for the unused argument: `ptr addrspace(1) %srcC`.

My preference would be to not generate any variable for unused kernel arguments going forwards, but clvk would require changes to support that. Currently when I implement that change it fails the test trying to set the kernel arguments. I can think of a few ways forward:
1. Clspv does not generate any variable and clvk changes to accommodate that.
2. Clspv generates some reflection information about the number of arguments.
3. Wait for LLVM to add some more info.

I'd prefer the first option, but I'm not certain how clvk would generate the diagnostics. The third option isn't palatable unless it happens very soon. That leaves something along the lines of the second option. That wouldn't be sufficient for `clGetKernelArgInfo`, but we can tackle that later.

Thoughts @kpet?

alan-baker:
I've prototyped this through various flows and it doesn't require too many changes to clvk. The changes in clspv can be localized to descriptor allocation.

PR #492, opened 2023-01-10
Handle kernel arguments not generated by clspv reflection

alan-baker:
Fixes https://github.com/kpet/clvk/issues/479

* Preparation for clspv switching to opaque pointers by default
* Parse the NumArguments kernel operand from clspv reflection
  * Use that to pre-generate all of the arguments as unused arguments
  * Used arguments get overridden by the correct data after parsing
* Support unused arguments (i.e. don't error) in a few places in clvk:
  * allow nullptr arg values
  * ignore them when creating descriptor layouts
  * ignore them when generating push constant ranges

The committed version of clspv will generate the NumArguments reflection info, but still generates reflection for unused arguments. I have tested a version of clspv that no longer generates reflection arguments in those cases and it works as expected. This will be necessary for some opaque pointer cases. Adding this support before clspv removes generation breaks a dependency for me internally.

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -2793,8 +2793,12 @@ cl_int CLVK_API_CALL clSetKernelArg(cl_kernel kern, cl_uint arg_index,
         return CL_INVALID_ARG_INDEX;
     }
 
+    // With opaque pointers, clspv is unable to infer the type of an unused
+    // kernel argument so just allow nullptr for its value. It will not have an
+    // affect on the kernel's operation.
     if ((arg_value == nullptr) &&
-        (kernel->arg_kind(arg_index) != kernel_argument_kind::local)) {
+        !((kernel->arg_kind(arg_index) != kernel_argument_kind::local) ||
+          (kernel->arg_kind(arg_index) != kernel_argument_kind::unused))) {
         cvk_error_fn("passing a null pointer to clSetKernelArg is only "
                      "supported for local arguments");
         return CL_INVALID_ARG_VALUE;
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -2794,11 +2794,11 @@ cl_int CLVK_API_CALL clSetKernelArg(cl_kernel kern, cl_uint arg_index,
     }
 
     // With opaque pointers, clspv is unable to infer the type of an unused
-    // kernel argument so just allow nullptr for its value. It will not have an
+    // kernel argument so allow nullptr for its value. It will not have an
     // affect on the kernel's operation.
     if ((arg_value == nullptr) &&
-        !((kernel->arg_kind(arg_index) != kernel_argument_kind::local) ||
-          (kernel->arg_kind(arg_index) != kernel_argument_kind::unused))) {
+        !((kernel->arg_kind(arg_index) == kernel_argument_kind::local) ||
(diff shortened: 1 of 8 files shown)


alan-baker:
Always stripping the non-semantic info passes the previously failing test, but now the compilation from IL seems to generate an 8-bit integer. No idea why that would change unless I managed to pull in a bad translator change.

alan-baker:
> Always stripping the non-semantic info passes the previously failing test, but now the compilation from IL seems to generate an 8-bit integer. No idea why that would change unless I managed to pull in a bad translator change.

To be honest I'm confused when I check the device options code. Clspv defaults to supporting int8 so I'd expect there to be code disabling it, but instead it only enables it. Not sure how these are ever passing.

alan-baker:
@kpet please consider this before #493. In my mind the last issue is stripping reflection in the unavailable compiler case. That would require building SPIRV-Tools for such a use.

src/api.cpp
```diff
@@ -2793,8 +2793,12 @@ cl_int CLVK_API_CALL clSetKernelArg(cl_kernel kern, cl_uint arg_index,
         return CL_INVALID_ARG_INDEX;
     }
 
+    // With opaque pointers, clspv is unable to infer the type of an unused
+    // kernel argument so just allow nullptr for its value. It will not have an
```
kpet, line 2797:
```suggestion
    // kernel argument so allow nullptr for its value. It will not have an
```
I think the "just" is confusing as non-nullptr values AFAIU are accepted for unused arguments (they should be!).

src/api.cpp
```diff
@@ -2793,8 +2793,12 @@ cl_int CLVK_API_CALL clSetKernelArg(cl_kernel kern, cl_uint arg_index,
...
+    // affect on the kernel's operation.
     if ((arg_value == nullptr) &&
-        (kernel->arg_kind(arg_index) != kernel_argument_kind::local)) {
+        !((kernel->arg_kind(arg_index) != kernel_argument_kind::local) ||
+          (kernel->arg_kind(arg_index) != kernel_argument_kind::unused))) {
```
kpet, line 2801: Unless I'm mis-parsing this, I think the logic is wrong here. Didn't you intend `is_null && !(is_local || is_unused)`? Also, I don't think we can do this with the spec as currently written. Negative tests passing say an invalid (but unused) buffer or sampler would fail.

kpet (changes requested):
Thanks for dealing with the update and fixing a couple of things as well! Unless I've misunderstood, I think this will need changes.

src/api.cpp
alan-baker: Unfortunately with opaque pointers I'm not sure there's much we can do. We can either error for an unused argument and have a false positive for local memory arguments or not error and have a false negative for buffers, images and samplers. I thought it would be better to have the false negatives since the unused argument has no effect.

kpet: Yeah, we will need a solution to this (maybe using arg info in clvk, maybe by generating descriptor info from arg info in clspv, maybe changing the spec?) at some point but it should not get in the way of real-world applications so I'd rather we unblock you and get forward progress now. Problem for future us :).

PR #493, opened 2023-01-13
Update deps and CTS

kpet:


alan-baker:
@kpet this seems to hit similar problems as #492. I've tried to integrate the last change there, but I have encountered other issues too.
1. Swiftshader using older headers for reflection than clspv generates. My solution was to always strip non-semantic info, but that fails when the compiler is not available.
2. Clspv generating bad array strides for some i8 without i8 cases. This is hopefully fixed by your last commit.
3. Clvk not handling the -int8 option properly. My PR has a fix for this (treat it like -fp16, etc).

How would you like to proceed? We're sort doing the same work. I'd like to unblock #492 as it's a prerequisite for https://github.com/google/clspv/pull/1001 and enabling opaque pointers in clspv by default.

alan-baker:
The spirv-llvm-translator is changing the kernel argument from i32* to i8* in the translation. It is also generating invalid code that the validator is not catching (a bitcast no-op).

alan-baker:
> The spirv-llvm-translator is changing the kernel argument from i32* to i8* in the translation. It is also generating invalid code that the validator is not catching (a bitcast no-op).

The sad part is I can make this flow work with opaque pointers (if llvm-spirv also runs `-spirv-mem2reg`), but I don't see an easy way of making transparent pointers work correctly.

alan-baker:
> > The spirv-llvm-translator is changing the kernel argument from i32* to i8* in the translation. It is also generating invalid code that the validator is not catching (a bitcast no-op).
> 
> The sad part is I can make this flow work with opaque pointers (if llvm-spirv also runs `-spirv-mem2reg`), but I don't see an easy way of making transparent pointers work correctly.

I think this is actually a presubmit issue. I'm testing a fix in #492. The initial translation through llvm-spirv needs to also disable opaque pointers for now.

PR #500, opened 2023-03-21
Revert LLVM CMake Workaround

rjodinchr:
Building with LLVM_TARGETS_TO_BUILD set to an empty string works now

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,13 +93,10 @@ if (CLVK_BUILD_SPIRV_TOOLS)
 endif()
 
 # clspv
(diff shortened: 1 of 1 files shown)


kpet (approved):
LGTM! Let's try. Worst case we re-re-re-vert (it's been broken and fixed multiple times over the years) :).

rjodinchr:
Thanks!
One thing I don't understand is that I don't see any commits in `clspv` to do this workaround. It seems to have always been set to the empty string...

kpet:
I don't know why that would be.

PR #502, opened 2023-03-30
Transition to opaque pointers

### LINKED ISSUE #380: Transition to opaque pointers
- [ ] Re-enable opaque pointers in offline compilation tests
- [ ] Re-enable opaque pointers in SPIR-V consumption

kpet:
Also update deps and CTS.

Fixes #380

Change-Id: I592f304f17278565351d1ffe5e7d7945008cd56c


src/program.cpp
```diff
@@ -1055,8 +1053,6 @@ cl_build_status cvk_program::do_build_inner_online(bool build_to_ir,
             int llvmArgc = 2;
```
rjodinchr, line 1053: shouldn't `llvmArgv` be updated to `1`?

kpet: Yes, good catch!

rjodinchr (approved):

PR #506, opened 2023-04-12
remove extra parentheses

rjodinchr:
This is breaking in Google infra


kpet (approved):
What flagged the issue? Maybe we could add this to the CI on Github?

rjodinchr:
`-Wparentheses-equality` is not supported by the CI compilers. Not sure what to do to help avoid it in the future here.

kpet:
Seems it's only supported by Clang and enabled by `-Wall` that we've already got enabled. Maybe we could add clang build? It's probably time to update the version of Ubuntu too. Anyway, this PR can go in as is.

kpet:
Also: the above reasoning assumes we'd enable `-Werror`, I'll look into that.

Issue #494, opened 2023-02-10
Attempt at use ffmpeg opencl filters result in compilation failure (termux)


Randrianasulu:
I have Android device with vendor-provided Vulkan driver (somewhere deep in system, I am not root so I can't change much)

in termux I have

clvk/stable,now 0.0.20230124.132813g9635d8e0 aarch64 [installed]
  Experimental implementation of OpenCL on Vulkan

it gives clinfo like this:

```
~ $ clinfo
Number of platforms                               1
  Platform Name                                   clvk
  Platform Vendor                                 clvk
  Platform Version                                OpenCL 3.0 clvk
  Platform Profile                                FULL_PROFILE
  Platform Extensions                             cl_khr_icd cl_khr_extended_versioning
  Platform Extensions with Version                cl_khr_icd                                                       0x400000 (1.0.0)
                                                  cl_khr_extended_versioning                                       0x400000 (1.0.0)
  Platform Numeric Version                        0xc00000 (3.0.0)
  Platform Extensions function suffix             clvk
  Platform Host timer resolution                  0ns

  Platform Name                                   clvk
Number of devices                                 1
  Device Name                                     PowerVR Rogue GE8322
  Device Vendor                                   FIXME
  Device Vendor ID                                0x1010
  Device Version                                  OpenCL 3.0 CLVK on Vulkan v1.1.131 driver 5776728
  Device UUID                                     32322038-3620-3130-3420-323138000000
  Driver UUID                                     35373736-3732-3800-0000-000000000000
  Valid Device LUID                               No
  Device LUID                                     0000-000000000000
  Device Node Mask                                0
  Device Numeric Version                          0xc00000 (3.0.0)
  Driver Version                                  3.0 CLVK on Vulkan v1.1.131 driver 5776728
  Device OpenCL C Version                         OpenCL C 1.2 CLVK on Vulkan v1.1.131 driver 5776728
  Device OpenCL C Numeric Version                 0x402000 (1.2.0)
  Device OpenCL C all versions                    OpenCL C                                                         0x400000 (1.0.0)
                                                  OpenCL C                                                         0x401000 (1.1.0)
                                                  OpenCL C                                                         0x402000 (1.2.0)
                                                  OpenCL C                                                         0xc00000 (3.0.0)
  Device OpenCL C features                        __opencl_c_images                                                0xc00000 (3.0.0)
                                                  __opencl_c_read_write_images                                     0xc00000 (3.0.0)
[...]

Issue #498, opened 2023-02-26
Report name of unsupported SPIR-V capabilities and Vulkan features in error messages


kpet:

kpet:
We should get SPIRV-Headers to provide strings for capabilities.

PR #507, opened 2023-04-12
print capability name in logs

rjodinchr:
Ref #494


rjodinchr:
Ref #498

kpet (approved):
I'd rather this be provided by the SPIRV-Headers so we don't have to maintain this here but it's too useful not to merge :).

PR #508, opened 2023-04-12
add device properties for nvidia and fix swiftshader

rjodinchr:
isSwiftShaderDevice was bug, this is fixing it.
Also update the list of native builtins passing the CTS.

add device properties for nvidia withthe list of native builtins passing the CTS.


kpet (approved):
LGTM, thanks! Four new tests passing at my end with this.

PR #509, opened 2023-04-12
Remove extra parentheses

alan-baker:


alan-baker:
Didn't see #506. Closing in favour of it.

PR #517, opened 2023-05-03
Update device properties for Samsung device

lpavank:


rjodinchr:
I'm completely ok with this CL going in, but it feels like a lot of native builtins. I have never seen that many being OpenCL compliant in a Vulkan driver (even on AMD ones that should be close to yours). Are you sure you have tested them with the full math test (not with the wimpy option)?
As said before I don't mind having it merged as we can update it easily in the future if needs be.

lpavank:
@rjodinchr I share the concerns you have (especially compared with other vendors). But, we have some driver issues (memory and/or compiler related) that's blocking us for a "full run" of bruteforce (non wimpy mode) at this time, so I did not get a chance to try non-wimpy mode. So, only tried wimpy mode, and it was fine. 

And, the rationale I had is, I can always come back and tweak in future as needed, but honestly just "fma" is wrong.

thoughts/concerns?

rjodinchr:
Sounds good to me

rjodinchr (approved):

kpet (approved):

PR #518, opened 2023-05-05
allow to modify gtest libraries used to link for tests

rjodinchr:


rjodinchr:
This will be useful in Chromeos infrastructure as we build the test separately from clvk. Thus llvm's `gtest` libraries are not available.

kpet (approved):
LGTM, glad this worked. One less patch you'll have to carry :).

PR #520, opened 2023-05-09
add clangd cache and expected folders in external to gitignore

rjodinchr:


kpet (commented):
I'm wondering whether these (and the existing `/build`) shouldn't only exist in local `.git/info/exclude` files.

rjodinchr:
good point, I'll do that. Thanks

PR #526, opened 2023-05-15
remove cvk_command_combine in favor of cvk_event_combine

rjodinchr:
cvk_command_combine prevents optimization where commands could be batch.
It is needed to have a single event representing the execution of the group of commands. This PR creates a new event (cvk_event_combine) which allows to have a single event grouping several events, thus keeping the single event for the group of commands.

It also allows to make do_action a private function, thus having it called only in one place. This is very convenient to implement future optimization (using VkSemaphore).

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -1263,7 +1263,7 @@ cl_event CLVK_API_CALL clCreateUserEvent(cl_context context,
         }
     }
 
-    auto event = new cvk_event(icd_downcast(context), nullptr, nullptr);
+    auto event = new cvk_event_command(icd_downcast(context), nullptr, nullptr);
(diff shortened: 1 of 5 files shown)


rjodinchr:
It can feel like a lot of change, but most of it is just code being moved around.

It is passing the full CTS on Nvidia and Swiftshader on my side.

rjodinchr:
I have updated the CL following #776 which brought merge conflicts.

rjodinchr:
> mem_host_no_access_image (all tested GPUs)

That one should be fixed now.

> atomic_flag (AMD only)

I'm not able to reproduce, could you share what's going wrong?

PR #530, opened 2023-05-17
Replace atexit function with static destroyer object for global state

bmanga:
This PR changes how/when the `gGlobalState` gets deleted. Instead of using `at_exit`, which destroys the object at the end of main, it uses the destructor a static global variable. 

I'm not sure about whether this supersedes the WIN32 workaround, so I removed to see whether the CI would pick up any issue.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -255,20 +255,14 @@ clvk_global_state::~clvk_global_state() {
     term_logging();
 }
 
-static clvk_global_state* gGlobalState;
+static clvk_global_state* gGlobalState = nullptr;
 static std::once_flag gInitOnceFlag;
 
-static void destroy_global_state() { delete gGlobalState; }
+static struct global_state_deleter {
+    ~global_state_deleter() { delete gGlobalState; }
+} gs_deleter;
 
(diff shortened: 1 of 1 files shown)


kpet (changes requested):
First of all, thanks for the contribution! I agree this is a problem with some libraries. Strictly speaking the OpenCL specification states that using any API functions from global constructors and destructors is not portable: https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_global_constructors_and_destructors but attempting to change every library out there is a losing battle.

I think we should keep the current lazy initialisation scheme because that works well in all cases and also enables clvk to not perform its initialisation sequence in applications that link with or load OpenCL but don't use it.

To deal with the destructor problem we could add a config option (e.g. a boolean `CLVK_DESTROY_GLOBAL_STATE` or better name) to opt-out of destroying the global state. Would that work for you? It pushes the problem to the user but at least they now have a lever they can use.

EDIT: As you can see this change is not passing the CI on Windows because the ordering of global destructors across libraries is undefined.

PR #532, opened 2023-05-18
Update program.cpp

rhusain1:
Added kernel name in pipeline creation error message.

### DIFF
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -2006,7 +2006,7 @@ cvk_entry_point::create_pipeline(const cvk_spec_constant_map& spec_constants) {
                                  &createInfo, nullptr, &pipeline);
 
     if (res != VK_SUCCESS) {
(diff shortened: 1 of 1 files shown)


src/program.cpp
```diff
@@ -2006,8 +2006,8 @@ cvk_entry_point::create_pipeline(const cvk_spec_constant_map& spec_constants) {
                                  &createInfo, nullptr, &pipeline);
 
     if (res != VK_SUCCESS) {
-        cvk_error_fn("Could not create compute pipeline: %s",
-                     vulkan_error_string(res));
+        cvk_error_fn("Could not create compute pipeline: %s for kernel %s",
```
kpet, line 2009:
```suggestion
        cvk_error_fn("Could not create compute pipeline for kernel %s: %s",
```

kpet (approved):
Thanks!

PR #533, opened 2023-05-19
Update the MAX_MEM_ALLOC_SIZE query 

lpavank:
Change the MAX_MEM_ALLOC_SIZE using the VK Caps which is more closer to OpenCL MAX_MEM_ALLOC_SIZE.
Also, for devices stick with minimum of the spec by default and provide a override (via config.def) to end user (similar to percentage), if he wants to override it.


rjodinchr (changes requested):
This PR is changing the permission of `src/device_properties.cpp`. Please leave the permission of this file untouched.

lpavank:
> This PR is changing the permission of `src/device_properties.cpp`. Please leave the permission of this file untouched.

done

src/device.hpp
```diff
@@ -217,13 +227,37 @@ struct cvk_device : public _cl_device_id,
...
+                     (uint64_t)(32 * 1024 * 1024));
+
+        // Max memory allocation for single buffer is set in config.def to 1024
+        // mb - minimum required by spec. For single allocation this value can
+        // be adjusted with environment variable CLVK_MAX_ALLOC_SIZE_MB For
+        // multiple allocations(total memory allocations), environment var
+        // CLVK_PERCENTAGE_OF_AVAILABLE_MEMORY_REPORTED can be adjusted.
+        auto maxAllocSz =
+            std::min(m_mtn3Properties.maxMemoryAllocationSize,
+                     (uint64_t)config.max_alloc_size_mb() * 1024 * 1024);
```
kpet, line 248: I'd rather we set the default value of the option to something that's obviously wrong like zero and only apply the min if the option is set (you can test `config.max_alloc_size_mb.set`). The spec does not really say 1024MB as clearly...

src/device.hpp
```diff
@@ -217,13 +227,37 @@ struct cvk_device : public _cl_device_id,
         return size * percentage_of_available_memory_reported;
     }
 
-    uint64_t memory_size() const {
+    uint64_t global_memory_size() const {
         return std::min(max_alloc_size() * 4, actual_memory_size());
```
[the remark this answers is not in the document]
lpavank: yeah, if you are saying, MAX_MEM_ALLOC_SIZE should be derived from GLOBAL_MEM_SIZE, not the other way around. That's exactly my opinion as well, but we were unsure why CLVK in the first place (even before this change) was deriving it the other way around. Did not want to change it without understanding the reason why it was done the other way.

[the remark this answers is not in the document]
kpet: Yeah, I can't remember why it was done this way but we should revisit and if we discover there was a good reason (I doubt that there's still a valid reason), document it :).

lpavank:
Thanks @kpet for the comments. Addressed most of your comments, in the latest commit, if you have any more concerns, please let me know.

src/device.hpp
```diff
@@ -545,6 +564,8 @@ struct cvk_device : public _cl_device_id,
     VkPhysicalDevice m_pdev;
     // Properties
     VkPhysicalDeviceProperties m_properties;
+    VkPhysicalDeviceProperties2 m_properties2;
```
kpet, line 567: This is not needed with the change requested above.

src/device.cpp
```diff
@@ -88,6 +88,16 @@ void cvk_device::init_vulkan_properties(VkInstance instance) {
...
     func(m_pdev, &properties);
+
+    //--- Get maxMemoryAllocationSize for figuring out the  max single buffer
+    // allocation size.
+    m_maintenance3_properties.sType =
+        VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_MAINTENANCE_3_PROPERTIES;
+    m_maintenance3_properties.pNext = nullptr;
+    m_properties2.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PROPERTIES_2;
+    m_properties2.pNext = &m_maintenance3_properties;
+    vkGetPhysicalDeviceProperties2(m_pdev, &m_properties2);
```
kpet, line 99: You don't need to call `vkGetPhysicalDeviceProperties2` here. This function is designed so that the queries are all chained in a single call. You can move the init line to the top of the function, along with the other struct initialisations and add an entry to `coreversion_extension_properties`. Everything else is automatic, including checking for the extension presence (I didn't spot that in my previous review but you should add the extension to the list of desired extensions in https://github.com/kpet/clvk/blob/main/src/device.cpp#L207).

src/device.hpp
```diff
@@ -217,13 +217,32 @@ struct cvk_device : public _cl_device_id,
...
     }
 
-    uint64_t memory_size() const {
-        return std::min(max_alloc_size() * 4, actual_memory_size());
-    }
+    uint64_t max_mem_alloc_size() const {
+        // Min memory as per the specs.
+        auto specMinAllocSz =
+            std::max(std::min((uint64_t)(1024 * 1024 * 1024),
+                              (uint64_t)((1 / 4) * global_mem_size())),
```
kpet, line 224: This code will perform the integer division `1 / 4` which will always be 0. You probably want to write `global_mem_size() / 4` instead and maybe add an assert checking that `global_mem_size()` returns a multiple of ':

```
CVK_ASSERT(global_mem_size() % 4 == 0);
```

src/device.hpp
```diff
@@ -217,13 +217,32 @@ struct cvk_device : public _cl_device_id,
...
+                     (uint64_t)(32 * 1024 * 1024));
+
+        // Max memory allocation for single buffer is set in config.def to 1024
+        // mb - minimum required by spec. For single allocation this value can
+        // be adjusted with environment variable CLVK_MEM_MAX_ALLOC_SIZE_MB For
+        // multiple allocations(total memory allocations), environment var
+        // CLVK_PERCENTAGE_OF_AVAILABLE_MEMORY_REPORTED can be adjusted.
+        auto maxAllocSz =
+            std::min(m_maintenance3_properties.maxMemoryAllocationSize,
+                     (uint64_t)config.max_mem_alloc_size_mb() * 1024 * 1024);
```
kpet, line 234: I think we should handle the case where the maintenance3 extension is not present too. Value-initialising `m_maintenance3_properties` would be good enough but maybe a bit subtle to reason about. I suggest you either do that or check for the extension's presence.

src/device.hpp
```diff
@@ -217,13 +217,32 @@ struct cvk_device : public _cl_device_id,
...
-    }
+    uint64_t max_mem_alloc_size() const {
+        // Min memory as per the specs.
+        auto specMinAllocSz =
+            std::max(std::min((uint64_t)(1024 * 1024 * 1024),
+                              (uint64_t)((1 / 4) * global_mem_size())),
+                     (uint64_t)(32 * 1024 * 1024));
+
+        // Max memory allocation for single buffer is set in config.def to 1024
+        // mb - minimum required by spec. For single allocation this value can
```
kpet, line 228:
```suggestion
        // MiB - minimum required by spec. For single allocation this value can
```

kpet (changes requested):
Thanks for the changes and the nice tidy-up of existing issues. A couple more tweaks needed but we're converging.

lpavank:
Thanks again Kevin, all/most of your new comments should be addressed, let me know if I missed any.

src/device.cpp
```diff
@@ -48,6 +48,13 @@ void cvk_device::init_vulkan_properties(VkInstance instance) {
     m_subgroup_properties.sType =
         VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SUBGROUP_PROPERTIES;
 
+    //--- Get maxMemoryAllocationSize for figuring out the  max single buffer
+    // allocation size.
+    m_maintenance3_properties.sType =
+        VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_MAINTENANCE_3_PROPERTIES;
+    m_maintenance3_properties.pNext = nullptr;
```
kpet, line 55:
```suggestion
```

You don't need this, pNext pointers are all handled automatically by the code below :).

kpet (changes requested):
Thanks!

src/device.cpp
```diff
@@ -61,6 +68,8 @@ void cvk_device::init_vulkan_properties(VkInstance instance) {
                          m_pci_bus_info_properties),
             VER_EXT_PROP(VK_MAKE_VERSION(1, 1, 0), nullptr,
                          m_subgroup_properties),
+            VER_EXT_PROP(VK_MAKE_VERSION(1, 0, 0), nullptr,
```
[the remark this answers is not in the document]
lpavank: yeah I understood the code there as follows. 
Like sub group become core in 1.2 (if I am not mistaken), and was an extension in 1.1.
Similarly, this feature(maintainence3) was core in 1.1 and extension in 1.0. That's why I put it like that.
Did I mis interpret something here?

src/device.hpp
lpavank: IIRC we tried on a device (hacked our device that we dont support this extension), and made sure we gracefully handled this case. I remember in our local testing, it was gracefully handled. 
But, will look at your suggestion and change accordingly.

src/device.cpp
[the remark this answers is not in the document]
kpet: It's the version in which the feature became core. The code below only checks for the extension if `m_properties.apiVersion >= corever` was false. What this whole exchange has taught me is that there should be more documentation/comments about this. It should have been easier for  you to make that change.

src/device.hpp
kpet: This test does not mean much as its outcome depends on the value of an uninitialised memory location. If you run the same test under valgrind it'll complain about this.

src/device.cpp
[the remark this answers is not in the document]
lpavank: done

src/device.cpp
lpavank: done

lpavank:
I should have addressed 3 of your comments from yesterday and the test on our end (single 1 buffer) still works.

kpet (approved):
LGTM, thanks!

PR #535, opened 2023-05-31
add option to disable dependency on LLVM-SPIRV-Translator

rjodinchr:
Keep it enabled by default
According to the OpenCL specification, just removing it from the list returned by CL_DEVICE_IL_VERSION should be enough.

This is useful for ChromeOS to remove dependency on LLVM-SPIRV-Translator as it is not strictly needed for OpenCL compliance and we do not have a use for it.

### DIFF
--- a/src/device.cpp
+++ b/src/device.cpp
@@ -521,7 +521,9 @@ void cvk_device::build_extension_ils_list() {
         // Add always supported extensions
         MAKE_NAME_VERSION(1, 0, 0, "cl_khr_extended_versioning"),
         MAKE_NAME_VERSION(1, 0, 0, "cl_khr_create_command_queue"),
+#ifdef ENABLE_SPIRV_IL
         MAKE_NAME_VERSION(1, 0, 0, "cl_khr_il_program"),
+#endif
         MAKE_NAME_VERSION(1, 0, 0, "cl_khr_spirv_no_integer_wrap_decoration"),
         MAKE_NAME_VERSION(1, 0, 0, "cl_arm_non_uniform_work_group_size"),
         MAKE_NAME_VERSION(1, 0, 0, "cl_khr_suggested_local_work_size"),
--- a/src/device.cpp
+++ b/src/device.cpp
@@ -575,7 +575,9 @@ void cvk_device::build_extension_ils_list() {
 
     // Build list of ILs
     m_ils = {
+#ifdef ENABLE_SPIRV_IL
         MAKE_NAME_VERSION(1, 0, 0, "SPIR-V"),
+#endif
     };
 
     for (auto& il : m_ils) {
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -1020,8 +1020,8 @@ cl_build_status cvk_program::do_build_inner_offline(bool build_to_ir,
     // Save input program to a file
     if (build_from_il) {
 #ifndef ENABLE_SPIRV_IL
-        cvk_error_fn(
-            "Trying to build SPIR-V IL while SPIR-V IL has been disabled");
+        cvk_error_fn("Could not build from il because clvk has been built with "
+                     "CLVK_ENABLE_SPIRV_IL=OFF");
         return CL_BUILD_ERROR;
 #else  // ENABLE_SPIRV_IL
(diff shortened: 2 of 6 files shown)


src/device.cpp
```diff
@@ -575,7 +575,9 @@ void cvk_device::build_extension_ils_list() {
 
     // Build list of ILs
     m_ils = {
+#ifdef ENABLE_SPIRV_IL
         MAKE_NAME_VERSION(1, 0, 0, "SPIR-V"),
+#endif
```
kpet, line 580: You also need to stop reporting the cl_khr_il_program extension as SPIR-V is a mandatory IL for implementations that support the extension.

src/program.cpp
```diff
@@ -995,6 +995,13 @@ bool cvk_program::parse_user_spec_constants() {
     return true;
 #endif // CLSPV_ONLINE_COMPILER
 #else
+#ifndef COMPILER_AVAILABLE
```
kpet, line 998:
```suggestion
#if !COMPILER_AVAILABLE
```

src/program.cpp
```diff
@@ -1012,6 +1019,11 @@ cl_build_status cvk_program::do_build_inner_offline(bool build_to_ir,
     std::string clspv_input_file{tmp_folder + "/source"};
     // Save input program to a file
     if (build_from_il) {
+#ifndef ENABLE_SPIRV_IL
+        cvk_error_fn(
+            "Trying to build SPIR-V IL while SPIR-V IL has been disabled");
```
kpet, line 1024: I'd rather you use a more explicit message as above. It might not be clear to a user what "SPIR-V IL has been disabled means" and the error message is not as actionable.

src/program.cpp
```diff
@@ -1170,6 +1183,11 @@ cl_build_status cvk_program::do_build_inner_online(bool build_to_ir,
     cvk_info_fn("build_from_il %u - build_to_ir %u", build_from_il,
                 build_to_ir);
     if (build_from_il) {
+#ifndef ENABLE_SPIRV_IL
+        cvk_error_fn(
+            "Trying to build SPIR-V IL while SPIR-V IL has been disabled");
```
kpet, line 1188: Same here

kpet (changes requested):
Looks reasonable but we might want to add a build to the CI.

rjodinchr:
> Looks reasonable but we might want to add a build to the CI.

Do you want to have it in the actual matrix, meaning we kind of double everything? Or should I just add one build with it?

kpet (approved):
We definitely don't want a full matrix in the CI and I'm not even sure this build should be in the presubmit. I'll capture an issue for it, I think we're reaching the point where it makes sense to introduce another workflow that runs less frequently or at least ask ourselves the question. Maybe the embryo of a release workflow?

alan-baker:
I'm seeing CTS failures due to this. For example in `consistency_il_programs`:
```
[CLVK] parse_user_spec_constants: Could not parse user spec constants because clvk has been built with CLVK_ENABLE_SPIRV_IL=OFF
ERROR: Device does not support IL Programs but clCreateProgramWithIL did not return CL_INVALID_OPERATION! (Got CL_INVALID_VALUE, expected CL_INVALID_OPERATION from [third_party/opencl_cts/test_conformance/api/test_api_consistency.cpp:967](https://cs.corp.google.com/piper///depot/google3/third_party/opencl_cts/test_conformance/api/test_api_consistency.cpp?l=967&ws=tap-presubmit-server/56248946&snapshot=2))
consistency_il_programs FAILED
```

It appears the wrong error code is being returned.

alan-baker:
> @rjodinchr Was the above failure Alan reported observed in a new job with the translator disabled? If not, you shouldn't see the error message at all. This smells fishy. Agree the error code needs fixing too.

I found this before updating our internal builds to include the newly added define.

kpet:
Oh, that makes sense, then the PR should be all that's needed to fix the issue. Is it Bazel-based? Would it make sense to upstream this to clvk?

rjodinchr:
It is bazel-based, but it contains references to other projects (the dependencies and others), which are specific to google infrastructure (it would need to be rework to consider the layout with the submodules, which is not what we have internally), so I don't think it makes sense to upstream that.

PR #541, opened 2023-06-01
Protect changes to memory object mappings with a mutex

kpet:
Both command enqueue and execution need to mutate mappings and they can be concurrent.

Change-Id: Ic94b2a4ddb6170e1ed1b74399d51e9f39df1d3b7

### DIFF
--- a/src/memory.hpp
+++ b/src/memory.hpp
@@ -373,6 +373,7 @@ struct cvk_buffer : public cvk_mem {
     }
 
     bool insert_mapping(const cvk_buffer_mapping& mapping) {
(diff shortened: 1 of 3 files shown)


rjodinchr (approved):
I'll update https://github.com/kpet/clvk/pull/526 afterwards as it also contains part of this PR.

kpet:
Ah, yes, down to the names :). I'm amazed we haven't hit this before, it was so obviously wrong. I've been getting TSAN up and running and it's flagging a lot of issues beyond that one. I'm working on fixing them before we make more changes to execution or threading.

PR #552, opened 2023-06-20
Tidy up mappings when buffer map commands fail to execute

kpet:
Buffer mappings are created at command creation time. When the command fails execution, either directly or because its dependencies failed, the mapping needs to be destroyed.

Caught by the Vulkan validation layers.

Change-Id: I53eef3c4bc4c7f4c5f94ade39b16bde558e9da2c

### DIFF
--- a/src/memory.hpp
+++ b/src/memory.hpp
@@ -388,6 +388,14 @@ struct cvk_buffer : public cvk_mem {
         return mapping;
     }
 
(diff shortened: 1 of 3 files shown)


kpet:
Image mapping seem to suffer from similar issues. deferring to https://github.com/kpet/clvk/issues/553.

rjodinchr (approved):

PR #554, opened 2023-06-21
Only load SPIR-V reflection information once

kpet:
Doing it multiple times leads to duplicated bindings for literal samplers.

Keep the call that is in the path common to online and offline compilation.

Caught by the Vulkan validation layers.

Change-Id: I8cc43d0d4f84931a85ea616ad3ba7487fcc1a8a4

### DIFF
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -497,7 +497,7 @@ bool spir_binary::load(std::istream& istream, uint32_t size) {
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


kpet:
This fixes the last error reported by the validation layers for api_tests. I'll enable validation layers in the CI once this goes in.

rjodinchr (approved):

PR #561, opened 2023-06-27
remove unneeded build in satisfy_data_dependencies

rjodinchr:
building the command in satisfy_data_dependencies is not needed.

The command is enqueued just after. During the enqueue the command will be either built as a batchable command (in
`cvk_command_queue::enqueue_command`), or built when it will be added to a batch (in `cvk_command_batch::add_command`).


rjodinchr:
No regression running the cts on Nvidia and Swiftshader

kpet (approved):
Nice catch! I'll add an assert that would have caught this.

PR #562, opened 2023-06-27
Implement clGetKernelInfo with CL_KERNEL_ATTRIBUTES 

### LINKED ISSUE #398: Pass api:kernel_attributes

rjodinchr:
This PR is a clean rebase of #489.

With an update of clspv to get https://github.com/google/clspv/pull/1143

I have checked the implementation on both swiftshader and nvidia. It fixes `test_api kernel_attributes`.

Fixes #398

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -2858,7 +2858,6 @@ cl_int CLVK_API_CALL clGetKernelInfo(cl_kernel kern, cl_kernel_info param_name,
     cl_int ret = CL_SUCCESS;
     size_t ret_size = 0;
     const void* copy_ptr = nullptr;
-    // TODO could this be a union
     cl_uint val_uint;
     cl_context val_context;
     cl_program val_program;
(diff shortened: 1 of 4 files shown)


src/api.cpp
```diff
@@ -2858,6 +2858,7 @@ cl_int CLVK_API_CALL clGetKernelInfo(cl_kernel kern, cl_kernel_info param_name,
     cl_int ret = CL_SUCCESS;
     size_t ret_size = 0;
     const void* copy_ptr = nullptr;
+    // TODO could this be a union
```
kpet, line 2861: Yeah, maybe we can save a few tens of bytes of stack depending on what the compiler does :). This would apply to all the clGet*Info functions. Probably not worth losing sleep over...

src/api.cpp
```diff
@@ -2893,7 +2894,12 @@ cl_int CLVK_API_CALL clGetKernelInfo(cl_kernel kern, cl_kernel_info param_name,
         copy_ptr = &val_program;
         ret_size = sizeof(val_program);
         break;
-    case CL_KERNEL_ATTRIBUTES: // TODO implement
+    case CL_KERNEL_ATTRIBUTES: {
+        const auto& attrs = kernel->attributes();
+        copy_ptr = attrs.c_str();
+        ret_size = attrs.size() + 1;
```
kpet, line 2900: You could use `api_query_string` to avoid having to add the + 1 manually but this would mean a copy unless we use it in the objects directly. I think the copy is acceptable for this query.

kpet (approved):
I'm fine with the change as is but added a suggestion. I've also edited the PR description so the issue tracking the CTS test this fixes auto-closes when we merge the PR.

kpet:
Oh, and I checked that it does fix the attributes CTS test with no regressions at my end.

src/api.cpp
rjodinchr: I removed this comment.

src/api.cpp

kpet (approved):
Thanks!

PR #563, opened 2023-06-28
add support for roundingmoderte capability

rjodinchr:


rjodinchr:
This PR depends on https://github.com/google/clspv/pull/1146

rjodinchr:
In fact, as swiftshader does not support `RoundingModeRTE`, this PR could pass the CI it seems.
But it would break all platforms supporting `RoundingModeRTE` with an unknown option when calling `clspv`.
Let's wait for `clspv` PR to land ^^

src/device.cpp
```diff
@@ -426,6 +432,23 @@ void cvk_device::init_compiler_options() {
...
+        roundingModeRTE.push_back("32");
+    }
+    if (m_float_controls_properties.shaderRoundingModeRTEFloat64) {
+        roundingModeRTE.push_back("64");
+    }
+    if (roundingModeRTE.size() > 0) {
+        m_device_compiler_options += " -rounding-mode-rte=";
+        std::stringstream stream;
+        std::copy(roundingModeRTE.begin(), roundingModeRTE.end(),
+                  std::ostream_iterator<std::string>(stream, ","));
```
kpet, line 449: This results in a trailing comma which clspv does not accept. You're apparently going to have to wait for  https://en.cppreference.com/w/cpp/experimental/ostream_joiner to make it into a future standard for a solution from the standard library :P. I'm guessing it passed the CI because swiftshader does not support float controls.

kpet (changes requested):
Good call on making this optional. Requiring it would have prevented clvk from running on a number of older Vulkan implementations.

src/device.cpp
rjodinchr: fixed

kpet:
I'm seeing the following in conformance runs (along with many failures):

>error: Cannot set RoundingModeRTE for fp64 if fp64 is not supported

rjodinchr:
It is quite weird that `shaderRoundingModeRTEFloat64` could be set without `shaderFloat64`...

kpet:
I think it's because we don't enable fp64 support in clspv based on `shaderFloat64` since clspv doesn't have full support for doubles.

kpet (approved):

PR #565, opened 2023-06-29
Implement printf

rjodinchr:
Clspv provides a printf definition that stores the value of each argument in a program-wide printf storage buffer. This implementation uses the appropriate ClspvReflection instructions to parse this data and print the resulting strings at runtime.

This is a rebase of https://github.com/kpet/clvk/pull/420 + some fixes

### DIFF
--- a/src/kernel.hpp
+++ b/src/kernel.hpp
@@ -19,6 +19,8 @@
 #include <unordered_map>
 #include <vector>
 
+#include "spirv/unified1/NonSemanticClspvReflection.h"
+
 #include "memory.hpp"
 #include "objects.hpp"
 #include "program.hpp"
@@ -142,6 +144,13 @@ struct cvk_kernel : public _cl_kernel, api_object<object_magic::kernel> {
     cl_kernel_arg_type_qualifier arg_type_qualifier(cl_uint arg_index) const {
         return m_args.at(arg_index).info.type_qualifier;
     }
+
+    bool uses_printf() const {
+        return m_program->kernel_flags(m_name) &
+               NonSemanticClspvReflectionMayUsePrintf;
+    }
+
+    bool requires_serialized_execution() const { return uses_printf(); }
 
 private:
     friend cvk_kernel_argument_values;
--- a/src/kernel.hpp
+++ b/src/kernel.hpp
@@ -145,10 +145,7 @@ struct cvk_kernel : public _cl_kernel, api_object<object_magic::kernel> {
         return m_args.at(arg_index).info.type_qualifier;
     }
 
-    bool uses_printf() const {
-        return m_program->kernel_flags(m_name) &
-               NonSemanticClspvReflectionMayUsePrintf;
-    }
+    bool uses_printf() const { return m_entry_point->uses_printf(); }
 
     bool requires_serialized_execution() const { return uses_printf(); }
 
--- a/src/printf.cpp
+++ b/src/printf.cpp
@@ -1,0 +1,247 @@
+// Copyright 2022 The clvk authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <sstream>
+
+#include "printf.hpp"
+
+// Extract the conversion specifier from a format string
+char get_fmt_conversion(std::string_view fmt) {
+    auto conversionSpecPos = fmt.find_first_of("diouxXfFeEgGaAcsp");
+    return fmt.at(conversionSpecPos);
+}
+
+// Read type T from given pointer
+template <typename T> T read_buff(const char* data) {
+    return *(reinterpret_cast<const T*>(data));
+}
+
+// Read type T from given pointer then increment the pointer
+template <typename T> T read_inc_buff(char*& data) {
+    T out = *(reinterpret_cast<T*>(data));
+    data += sizeof(T);
+    return out;
+}
+
+// Extract the optional vector flag and return a modified format string suitable
+// for calling snprintf on individual vector elements
+std::string get_vector_fmt(std::string fmt, int& vector_size, int& element_size,
+                           std::string& remaining_fmt) {
+    // Consume flags (skipping initial '%')
+    auto pos = fmt.find_first_not_of(" +-#0", 1ul);
+    // Consume precision and field width
+    pos = fmt.find_first_not_of("123456789.", pos);
+
+    if (fmt.at(pos) != 'v') {
+        vector_size = 1;
+        return std::string{fmt};
+    }
+
+    // Trim the data after the conversion specifier and store it in
+    // `remaining_fmt`
+    auto pos_conversion = fmt.find_first_of("diouxXfFeEgGaAcsp");
+    auto fmt_specifier = fmt.substr(0, pos_conversion + 1);
+    remaining_fmt = fmt.substr(pos_conversion + 1);
+    fmt = fmt_specifier;
+
+    size_t vec_length_pos_start = ++pos;
+    size_t vec_length_pos_end =
+        fmt.find_first_not_of("23468", vec_length_pos_start);
+    auto vec_length_str = fmt.substr(vec_length_pos_start,
+                                     vec_length_pos_end - vec_length_pos_start);
+    int vec_length = std::atoi(vec_length_str.c_str());
+
+    auto fmt_pre_vec_len = fmt.substr(0, vec_length_pos_start - 1);
+    auto fmt_post_vec_len = fmt.substr(vec_length_pos_end, fmt.size());
+    fmt = fmt_pre_vec_len + fmt_post_vec_len;
+
+    // The length modifier is required with vectors
+    if (fmt_post_vec_len.find("hh") != std::string::npos) {
+        element_size = 1;
+    } else if (fmt_post_vec_len.find("hl") != std::string::npos) {
+        element_size = 4;
+    } else if (fmt_post_vec_len.find("h") != std::string::npos) {
+        element_size = 2;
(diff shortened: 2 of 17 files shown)


src/program.cpp
```diff
@@ -1787,6 +1826,34 @@ bool cvk_entry_point::
     return true;
 }
 
+bool cvk_entry_point::build_descriptor_sets_layout_bindings_for_printf_buffer(
+    binding_stat_map& smap) {
+    std::vector<VkDescriptorSetLayoutBinding> layoutBindings;
+    if (m_program->printf_buffer_info().size > 0 &&
+        (m_program->kernel_flags(m_name) &
+         NonSemanticClspvReflectionMayUsePrintf)) {
```
kpet, line 1834: `cvk_kernel::uses_printf` should probably become a member of `cvk_entry_point`.

src/program.hpp
```diff
@@ -770,6 +822,7 @@ struct cvk_program : public _cl_program, api_object<object_magic::program> {
     VkPipelineCache m_pipeline_cache;
     std::unique_ptr<cvk_buffer> m_module_constant_data_buffer;
     std::unordered_map<uint32_t, user_spec_constant_data> m_user_spec_constants;
+    std::unique_ptr<cvk_buffer> m_printf_buffer;
```
kpet, line 825:
```suggestion
```

The buffer is now owned by queues.

src/queue.cpp
```diff
@@ -593,6 +593,25 @@ void cvk_command_kernel::update_global_push_constants(
...
+
+        auto buffer = m_queue->printf_buffer();
+        VkBufferDeviceAddressInfo info{};
+        info.buffer = buffer->vulkan_buffer();
+        info.sType = VK_STRUCTURE_TYPE_BUFFER_DEVICE_ADDRESS_INFO;
+        info.pNext = NULL;
+
+        auto dev_addr = vkGetBufferDeviceAddress(
+            m_kernel->context()->device()->vulkan_device(), &info);
+        dev_addr += buffer->vulkan_buffer_offset();
```
kpet, line 608:
```suggestion
        auto dev_addr = buffer->device_address();
```

src/queue.cpp
```diff
@@ -757,6 +776,32 @@ cl_int cvk_command_kernel::dispatch_uniform_region_within_vklimits(
     vkCmdDispatch(command_buffer, num_workgroups[0], num_workgroups[1],
                   num_workgroups[2]);
 
+    // If we have a kernel that requires serial execution (i.e. regions are not
+    // executed in parallel with other regions or other kernels) then serialize
+    // the command buffer
+    if (m_kernel->requires_serialized_execution()) {
```
kpet, line 782: I'd have to dive into clspv to confirm but I don't think we need to serialise individual regions.

src/printf.cpp
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    printf("%s", printf_out.str().c_str());
+}
+
+void cvk_printf(cvk_mem* printf_buffer,
+                const printf_descriptor_map_t& descriptors) {
+    if (!printf_buffer->map()) {
+        cvk_error("Could not map printf buffer");
+        return;
```
kpet, line 236: We should probably propagate errors.

src/queue.hpp
```diff
@@ -733,6 +747,13 @@ struct cvk_command_kernel final : public cvk_command_batchable {
     CHECK_RETURN cl_int
     build_batchable_inner(cvk_command_buffer& cmdbuf) override final;
 
+    CHECK_RETURN cl_int do_post_action() override final;
+
+    bool can_be_batched() const override final {
+        return !m_kernel->uses_printf() &&
```
kpet, line 753: Fine as a first step but we should be able to support batching.

src/queue.hpp
```diff
@@ -171,6 +172,16 @@ struct cvk_command_queue : public _cl_command_queue,
         return m_command_pool.free_command_buffer(cmdbuf);
     }
 
+    cvk_buffer* printf_buffer() {
```
kpet, line 175:
```suggestion
    cvk_buffer* get_or_create_printf_buffer() {
```

To make the behaviour obvious from the name. We could also have a separate getter for the code running on completion where we could assert the buffer exists. With the current code we could end up creating a buffer in the completion code if we mess up the setup somehow.

Also, note to self: no locking is needed here because we are always holding the queue lock when creating the buffer.

src/queue.cpp
```diff
@@ -912,6 +957,41 @@ cvk_command_kernel::build_batchable_inner(cvk_command_buffer& command_buffer) {
...
     }
 
+    // Setup printf buffer descriptor if needed
+    if (m_kernel->program()->uses_printf()) {
+        // Create and initialize the printf buffer
+        auto buffer = m_queue->printf_buffer();
+        if (buffer->map()) {
+            memset(buffer->host_va(), 0, 4);
+            buffer->unmap();
+        }
```
kpet, line 967: Maybe this could be encapsulated into a `reset_printf_buffer`?

src/printf.cpp
```diff
@@ -0,0 +1,249 @@
...
+    }
+    char* data = static_cast<char*>(printf_buffer->host_va());
+    auto buffer_size = printf_buffer->size();
+    auto bytes_written = read_inc_buff<uint32_t>(data) * 4;
+    auto* data_start = data;
+
+    while (static_cast<size_t>(data - data_start) < bytes_written &&
+           static_cast<size_t>(data - data_start) < buffer_size) {
```
kpet, line 244: Should this be `buffer_size - 4`?

src/printf.cpp
```diff
@@ -0,0 +1,249 @@
...
+    if (!printf_buffer->map()) {
+        cvk_error("Could not map printf buffer");
+        return;
+    }
+    char* data = static_cast<char*>(printf_buffer->host_va());
+    auto buffer_size = printf_buffer->size();
+    auto bytes_written = read_inc_buff<uint32_t>(data) * 4;
+    auto* data_start = data;
+
+    while (static_cast<size_t>(data - data_start) < bytes_written &&
```
kpet, line 243: Does this mean that we never print anything if the kernel filled the whole buffer and we wrapped around (I haven't checked the clspv side)?

kpet (changes requested):
Thanks! We can leave removing the serialisation or adding support for batching to a future PR but there are a few things we should IMO fix as part of this one. Ideally we'd add a few tests (I'm thinking of exercising the case where we wrap-around the buffer at least) but I'm fine with deferring to a separate PR if you want.

FWIW, it's passing all the printf conformance tests at my end and the validation layers do not report anything.

src/program.cpp
rjodinchr: Done

src/printf.cpp
rjodinchr: Done

src/queue.cpp
rjodinchr: Done

src/printf.cpp
rjodinchr: I agree. The 4bytes being for the first element being the buffer offset as I understand it.

src/printf.cpp
rjodinchr: I have added a test showing that we print as much as possible.

rjodinchr:
As you proposed I have left removing the serialisation and adding support for batching to a future PR.

I tried to address every other comment. And I have added some tests. Especially to convert to wrapping behaviour.

src/kernel.hpp
```diff
@@ -19,6 +19,8 @@
 #include <unordered_map>
 #include <vector>
 
+#include "spirv/unified1/NonSemanticClspvReflection.h"
+
```
kpet, line 23:
```suggestion
```

Should no longer be necessary.

src/queue.cpp
```diff
@@ -958,6 +1039,19 @@ cvk_command_kernel::build_batchable_inner(cvk_command_buffer& command_buffer) {
     return CL_SUCCESS;
 }
 
+cl_int cvk_command_kernel::do_post_action() {
+    if (m_kernel->uses_printf()) {
+        auto buffer = m_queue->get_printf_buffer();
+        if (buffer == nullptr) {
+            cvk_error_fn("printf buffer was not created");
```
kpet, line 1046: I'm not sure this makes sense as a user-visible error. The only to hit this is if we made a mistake in clvk. Maybe an assert would have been enough/more appropriate.

src/queue.hpp
```diff
@@ -171,6 +172,33 @@ struct cvk_command_queue : public _cl_command_queue,
...
+                context(), 0, config.printf_buffer_size, nullptr, &status);
+            CVK_ASSERT(status == CL_SUCCESS);
+        }
+        return m_printf_buffer.get();
+    }
+
+    cvk_buffer* get_printf_buffer() {
+        if (!m_printf_buffer) {
+            return nullptr;
+        }
```
kpet, line 188: Don't think you need this.

tests/api/printf.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    char* ptr = stdoutBuffer;
+    do {
+        ptr += strlen(ptr);
+        ptr = fgets(ptr, BUFFER_SIZE, f);
+    } while (ptr != nullptr);
+    fclose(f);
+
+    return stdoutBuffer;
+}
```
kpet, line 69: We should introduce a callback to get access to the output of printf from unit tests. As discussed, I want to make this standard in a future version of OpenCL, similarly to how we've done it in the cl_arm_printf extension.

src/queue.cpp
```diff
@@ -915,6 +958,41 @@ cvk_command_kernel::build_batchable_inner(cvk_command_buffer& command_buffer) {
         return CL_OUT_OF_RESOURCES;
     }
 
+    // Setup printf buffer descriptor if needed
+    if (m_kernel->program()->uses_printf()) {
+        // Create and initialize the printf buffer
+        auto buffer = m_queue->get_or_create_printf_buffer();
+        auto err = m_queue->reset_printf_buffer();
```
kpet, line 965: This means we can be writing to the printf buffer concurrently to executing kernels (or maybe even the post-action code) with no protection whatsoever.

kpet (approved):
Thanks for all the changes. This is still suboptimal and broken in some cases but good enough for many common use-cases and the CTS. Let's put a stake in the ground.

kpet:
I've created https://github.com/kpet/clvk/issues/578 to track further work.

PR #566, opened 2023-07-01
update vulkan package to 1.3.250.1 for windows

rjodinchr:
1.3.250.0 is not available on the website anymore


kpet (approved):
Hmm, I was hoping for more stability. We might want to store our own copies of the SDK so the CI doesn't randomly break when someone at LunarG decides to remove a version.

PR #569, opened 2023-07-03
fix condition to use nonuniform decoration

rjodinchr:


kpet (commented):
So, after reading through specs a bit more, the original code should be all that we need and the reason you needed this was probably because of a broken implementation. I'm fine with getting a hack in to help solve issues with existing drivers but we should probably use device properties to enable it. We should also make sure that the Vulkan CTS has good coverage for these cases so newer drivers are fixed.

What were the symptoms? What vendor(s) did you see the issue with?

rjodinchr:
The issue was with mali devices where the driver reported the following message:
```
==>[WARN] *** COMPILATION FAILURE *** (compute shader)
0:1: SPV01: Unsupported capability (5301)
```

rjodinchr:
code formatting check will be fixed by #589

src/device.hpp
```diff
@@ -300,10 +300,31 @@ struct cvk_device : public _cl_device_id,
...
     bool supports_non_uniform_decoration() const {
-        return m_properties.apiVersion >= VK_MAKE_VERSION(1, 2, 0) ||
-               is_vulkan_extension_enabled(
-                   VK_EXT_DESCRIPTOR_INDEXING_EXTENSION_NAME);
+        bool supported = m_properties.apiVersion >= VK_MAKE_VERSION(1, 2, 0) ||
+                         is_vulkan_extension_enabled(
+                             VK_EXT_DESCRIPTOR_INDEXING_EXTENSION_NAME);
+
+        // This is a hack to work-around a driver bug in mali device using the
+        // bifrost driver.
```
kpet, line 308:
```suggestion
```

This is at risk of becoming out of date. I suggest to delete the comment a replace it by something more generic.

src/device_properties.cpp
```diff
@@ -25,18 +25,35 @@ struct cvk_device_properties_mali : public cvk_device_properties {
     std::string vendor() const override final { return "ARM"; }
     cl_uint get_max_first_cmd_batch_size() const override final { return 10; }
     cl_uint get_max_cmd_group_size() const override final { return 1; }
+
+    cvk_device_properties_mali(const uint32_t deviceID)
+        : m_deviceID(deviceID) {}
+
+    bool get_non_uniform_decoration_unsupported() const override final {
```
kpet, line 32:
```suggestion
    bool is_non_uniform_decoration_broken() const override final {
```

src/device.hpp
```diff
@@ -300,10 +300,31 @@ struct cvk_device : public _cl_device_id,
...
+                           m_features_descriptor_indexing
+                               .shaderStorageImageArrayNonUniformIndexing &&
+                           m_features_descriptor_indexing
+                               .shaderUniformBufferArrayNonUniformIndexing &&
+                           m_features_descriptor_indexing
+                               .shaderStorageBufferArrayNonUniformIndexing)) {
+            cvk_warn_fn("Device should support non-uniform decorations, but "
+                        "some expected features are set to false. This is "
+                        "unexpected and might lead to errors");
+        }
```
kpet, line 324:
```suggestion
```

It's perfectly valid for an implementation to support non-uniform indexing for only some descriptor types or even none but to correctly support the non-uniform decoration. Better to remove this IMO.

src/device_properties.cpp
```diff
@@ -25,18 +25,35 @@ struct cvk_device_properties_mali : public cvk_device_properties {
...
+#define GPU_ID2_ARCH_MAJOR_SHIFT 28
+#define GPU_ID2_ARCH_MAJOR (0xF << GPU_ID2_ARCH_MAJOR_SHIFT)
+        const uint32_t bifrost_arch_major = 9 << GPU_ID2_ARCH_MAJOR_SHIFT;
+        return (m_deviceID & GPU_ID2_ARCH_MAJOR) < bifrost_arch_major;
```
kpet, line 36: 9 is the first Valhall device. Less than 9 (i.e. before Valhall) means Bifrost.

src/device_properties.hpp
```diff
@@ -50,6 +50,12 @@ struct cvk_device_properties {
         return std::set<std::string>();
     }
 
+    // This is a hack to work-around a driver bug in mali device using the
+    // bifrost driver.
```
kpet, line 54:
```suggestion
```

Same comment as above.

src/device_properties.cpp
kpet: You could add the comment about Bifrost drivers being broken here.

kpet (changes requested):
Thanks! Please rebase to pick up your CI fix as well.

kpet (approved):
Thanks!

PR #570, opened 2023-07-04
print build log in debug log when failing to compile

rjodinchr:


kpet (approved):
That'll come in handy for apps that don't print build errors. If this gets annoying, we can always assign this to a group that can be filtered out.

PR #571, opened 2023-07-04
do not override build_batchable_inner error code

rjodinchr:
build_batchable_inner always return a proper CL error code. Use it instead of overwriting to CL_OUT_OF_RESOURCES to keep as much information as possible.

Also add a error log


kpet (approved):
That's better, thanks!

PR #572, opened 2023-07-07
Set CL_DEVICE_IMAGE_PITCH_ALIGNMENT to 1 when images are supported.

robquill:
The spec says that it should be 0 if images are not supported. Otherwise it should be a power of 2. 0 isn't a power of 2 and causes problems with tests which expect to be able to use CL_DEVICE_IMAGE_PITCH_ALIGNMENT to calculate memory allocation sizes etc.

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -769,6 +769,16 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
         size_ret = sizeof(val_sizet);
         break;
     case CL_DEVICE_IMAGE_PITCH_ALIGNMENT:
+        // Alignment of zero only makes sense if images are not supported.
+        if (device->supports_images()) {
+            val_uint = 1;
+        } else {
+            val_uint = 0;
+        }
+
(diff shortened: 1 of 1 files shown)


kpet (commented):
Thanks for the contribution! Assuming I'm reading the specification correctly, this change looks incorrect. The spec states:

>The value returned must be a power of 2.
>Support for 2D images created from a buffer is required for an OpenCL 2.0, 2.1, or 2.2 device if CL_DEVICE_IMAGE_SUPPORT is CL_TRUE.
>This value must be 0 for devices that do not support 2D images created from a buffer.

clvk is an OpenCL 3.0 implementation and does not currently support creating 2D images from buffers (see #43). As far as I can tell, reporting 0 is the correct behaviour.

Furthermore, this change fails the `consistency_2d_image_from_buffer` CTS test.

My guess would be that the failing application is making incorrect assumptions and not checking for support before using the value returned by `CL_DEVICE_IMAGE_PITCH_ALIGNMENT`.

PR #574, opened 2023-07-11
free descriptor sets in cvk_kernel_argument_values earlier

rjodinchr:
We used to free descriptor sets when cvk_kernel_argument_values is destroyed. But as cvk_kernel has a ref on it, it is only freed when clReleaseKernel is called.

To be able to free them earlier, keep a refcount between retain_resources and release_resources. When the counter reach zero, free the descriptor sets, and set m_is_enqueued to false to make sure we reallocate the descriptor set if we need to use them once again.

As descriptor sets can be allocated after being freed, reset them to VK_NULL_HANDLE.

### DIFF
--- a/src/kernel.hpp
+++ b/src/kernel.hpp
@@ -168,7 +168,8 @@ struct cvk_kernel_argument_values {
           m_args(m_entry_point->args()), m_pod_arg(nullptr),
           m_kernel_resources(m_entry_point->num_resource_slots()),
           m_local_args_size(m_entry_point->args().size(), 0),
-          m_args_set(m_args.size(), false), m_descriptor_sets{VK_NULL_HANDLE} {}
+          m_args_set(m_args.size(), false), m_descriptor_sets{VK_NULL_HANDLE},
+          m_descriptor_sets_refcount(0) {}
 
     cvk_kernel_argument_values(const cvk_kernel_argument_values& other)
         : m_entry_point(other.m_entry_point), m_is_enqueued(false),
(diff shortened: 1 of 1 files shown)


src/kernel.hpp
```diff
@@ -350,6 +346,16 @@ struct cvk_kernel_argument_values {
             if (resource)
                 resource->release();
         }
+        std::lock_guard<std::mutex> lock(m_lock);
+        if (--m_descriptor_sets_refcount == 0) {
+            m_is_enqueued = false;
+            for (auto ds : m_descriptor_sets) {
```
kpet, line 352:
```suggestion
            for (auto& ds : m_descriptor_sets) {
```

I think you intended `ds` to be a reference.

src/kernel.hpp
```diff
@@ -350,6 +346,16 @@ struct cvk_kernel_argument_values {
             if (resource)
                 resource->release();
         }
+        std::lock_guard<std::mutex> lock(m_lock);
+        if (--m_descriptor_sets_refcount == 0) {
+            m_is_enqueued = false;
```
kpet, line 351: This variable (and the associated getters) should really have been named: `m_was_used_by_in_flight_command`. We were never resetting it. With your change `m_is_used_by_in_flight_command` would become more appropriate. To be clear I'm not requesting you make those changes.

kpet (changes requested):
Your PR title says "fix" but, AFAICT (shout if I misunderstood something, I have not explored this very thoroughly), nothing was really broken. We were not leaking descriptor sets or accumulating them and your problem only arose because kernels kept the last argument values and its descriptor sets alive after the last enqueue (we should discuss this a bit more). It was only reused if none of the kernel arguments were changed (which shouldn't be that infrequent for kernels used in the middle of complex neural nets for example). I'm not opposed to this change but I want to make sure I fully understand what the issue you were facing was. I've created https://github.com/kpet/clvk/issues/580 to look into allocating much fewer descriptor sets.

rjodinchr:
You are right this is not really a fix. I changed the name.

I feel that my explanation in the commit message is not clear.
The reason for those descriptor_sets kept alive was not clear, and some guesses were made. But trying to fix it I found that `cvk_kernel` holds a `shared_ptr` on `cvk_kernel_argument_values`. Thus as we need to destroy `cvk_kernel_argument_values` to release the `descriptor_sets`, we need to wait for a `clReleaseKernel` to do it.

src/kernel.hpp
rjodinchr: Thanks!
I'll add back the destructor as this bug was making it segfault, but I think we should really keep the destructor.

kpet (approved):
This should not really be needed but since it helps with some drivers (at the cost of forcing more frequent descriptor set allocations in some cases), let's get it in. We can revisit under #580

PR #575, opened 2023-07-12
Add timing.hpp and make it a backend of the tracing system

rjodinchr:


README.md
```diff
@@ -227,6 +227,47 @@ Once the libOpenCL.so library has been built, verify with:
 LD_LIBRARY_PATH=/path/to/build clinfo
 ```
 
+### With global timing information
+
+Global timing information about API functions as well as some internal functions can be printed on stdout at the end of the execution by clvk.
```
kpet, line 232:
```suggestion
Global timing information about API functions as well as some internal functions can be logged at the end of the execution.
```

src/timing.hpp
```diff
@@ -0,0 +1,101 @@
...
+
+class cvk_timer {
+public:
+    cvk_timer(std::string name) : m_name(name) {}
+
+#ifdef CVK_ENABLE_TIMING
+
+    ~cvk_timer() {
+        double average = m_count ? (m_total_time / 1.0e6) / m_count : 0.0;
+        cvk_warn("%.2f ms -> %s (%d blocks, avg %.3f ms)", m_total_time / 1.0e6,
```
kpet, line 30: We could maybe assign a logging group to these messages. They're all printed at the end though so not critical at all to be able to filter them out.

kpet (approved):
Thanks!

PR #576, opened 2023-07-12
add properties for llvmpipe

rjodinchr:


kpet (approved):
Thanks! This will come in handy when we add lavapipe to the CI.

PR #577, opened 2023-07-14
manage spirv_arch & physical_addressing in device/device_properties

rjodinchr:
This will help once we have something detected whether or not to use physical addressing.
It will also help force some device to not use physical addressing if the feature is bugged.

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -529,7 +529,7 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
         size_ret = sizeof(val_fpconfig);
         break;
     case CL_DEVICE_ADDRESS_BITS:
-        val_uint = (config.spirv_arch() == "spir64") ? 64 : 32;
+        val_uint = (device->get_spirv_arch() == "spir64") ? 64 : 32;
         copy_ptr = &val_uint;
         size_ret = sizeof(val_uint);
         break;
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -529,7 +529,7 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
         size_ret = sizeof(val_fpconfig);
         break;
     case CL_DEVICE_ADDRESS_BITS:
-        val_uint = (device->get_spirv_arch() == "spir64") ? 64 : 32;
+        val_uint = device->address_bits();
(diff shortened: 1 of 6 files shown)


rjodinchr:
We need to rebuild swiftshader. It seems the artifact has expired thus removed.

src/device.hpp
```diff
@@ -528,6 +528,9 @@ struct cvk_device : public _cl_device_id,
         return m_max_first_cmd_group_size;
     }
 
+    const std::string get_spirv_arch() const { return m_spirv_arch; }
```
kpet, line 531:
```suggestion
```

No longer needed if we introduce `address_bits`.

src/device.hpp
```diff
@@ -528,6 +528,9 @@ struct cvk_device : public _cl_device_id,
...
+    bool get_physical_addressing() const { return m_physical_addressing; }
```
kpet, line 532:
```suggestion
    bool uses_physical_addressing() const { return m_physical_addressing; }
```

kpet (changes requested):
Agree this is a needed change and good direction. Just a few minor changes please.

src/api.cpp
```diff
@@ -529,7 +529,7 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
         size_ret = sizeof(val_fpconfig);
         break;
     case CL_DEVICE_ADDRESS_BITS:
-        val_uint = (config.spirv_arch() == "spir64") ? 64 : 32;
+        val_uint = (device->get_spirv_arch() == "spir64") ? 64 : 32;
```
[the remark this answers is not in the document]

src/memory.hpp
```diff
@@ -62,6 +64,7 @@ struct cvk_memory_allocation {
     VkDeviceSize m_size;
     VkDeviceMemory m_memory;
     uint32_t m_memory_type_index;
+    bool m_physical_addressing;
```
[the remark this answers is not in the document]
rjodinchr: Done

src/device.hpp
rjodinchr: Done

src/device.hpp
rjodinchr: Done

kpet (approved):
Thanks for the changes!

PR #584, opened 2023-07-19
fix set_arg with invalid pod pointer

rjodinchr:
I found this bug running api_tests on llvmpipe with physical addressing enabled.
api_tests: WithContext.SetUnusedMemeObjectArgWithInvalidObject failed

### DIFF
--- a/src/kernel.hpp
+++ b/src/kernel.hpp
@@ -268,6 +268,8 @@ struct cvk_kernel_argument_values {
                 set_pod_data(arg.offset, arg.size, &null);
             } else {
                 auto mem_downcast = icd_downcast(mem);
+                if (!mem_downcast->is_valid())
+                    return CL_INVALID_MEM_OBJECT;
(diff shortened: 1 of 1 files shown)


src/kernel.hpp
```diff
@@ -268,6 +268,8 @@ struct cvk_kernel_argument_values {
                 set_pod_data(arg.offset, arg.size, &null);
             } else {
                 auto mem_downcast = icd_downcast(mem);
+                if (!mem_downcast->is_valid())
+                    return CL_INVALID_MEM_OBJECT;
```
kpet, line 272:
```suggestion
                if (!mem_downcast->is_valid()) {
                    return CL_INVALID_MEM_OBJECT;
                }
```

kpet (approved):
We need testing with physical addressing in the CI.  I've created https://github.com/kpet/clvk/issues/586

PR #585, opened 2023-07-20
fix printf when reaching the end of the printf buffer

rjodinchr:
stop printing if the args size overflows the printf buffer size

add a new test


kpet (approved):
Thanks! This issue was flagged in the original review. I should have been more careful.

PR #587, opened 2023-07-24
clang-format: InsertBraces = true

rjodinchr:
Since clang-format-15 InsertBraces can be used to force brace around single line control statement.


kpet (approved):
Thanks for following up on that :)

PR #591, opened 2023-07-27
Add benchmark test for float-32

Rekt3421:
The idea here is to have micro-benchmarks similar to these [opencl](https://github.com/ProjectPhysX/OpenCL-Benchmark/blob/c6a154e8ae744f5f12a96b450ac55f271a34af72/src/kernel.cpp#L18) benchmarks. This is the first in a series of other benchmarks which will each have its own binary.

This benchmark finds the first GPU device and conduct a float 32 benchmark on the kernel for whatever number iterations are specified on the cmdline.


Rekt3421:
closing to fix issues.

PR #593, opened 2023-08-01
force flush fatal and error logs

rjodinchr:
While stdout is line buffered and stderr is unbuffered, file are just buffered by default.
Which means that fatal and error logs might not be printed (or completely printed) in some scenario.


rjodinchr:
It feels like there is an issue with azure

kpet (approved):

Issue #599, opened 2023-08-13
Quote/escape application-provided build options for the shell used to invoke clspv


kpet:
In builds where clspv is a separate binary, we're currently using the shell to execute it. Some applications (e.g. hashcat) use build options that the shell interprets. Those need to be quoted/escaped or we need to stop using the shell when invoking clspv. See #598.

giridhart:
@zwang20 The fix merged it expects option to have space after `-D` 
For example, `-DM2S(x)=XM2S(x)`  is not handled.

PR #601, opened 2023-08-13
Quote application-provided build options for the shell used to invoke clspv

zwang20:
See #599


kpet (approved):
You beat me to it! I'd started working on it  with a slightly different approach that would extend better to supporting quoting of arguments but this works for hashcat and does not seem to regress anything else, merging as a first step. Thanks!

PR #606, opened 2023-09-04
make OpenCL version configurable

rjodinchr:
This would allow us to test CL1.2/CL3.0 compliance without having to recompile clvk.

The end goal is to remove a patch in ChromeOS


kpet (changes requested):
The option looks good but please document it in the README. Thanks!

kpet (approved):
Looking at this again, we should probably take an X.Y.Z version number via the option rather than the encoded value in hex but this would require special handling. Fine as a first step.

PR #607, opened 2023-09-04
Avoid using VK_FORMAT_B8G8R8A8

rjodinchr:
This is following work on 1dbuffer.
Some Intel drivers do not support STORATE_TEXEL_BUFFERS and VK_FORMAT_B8G8R8A8_. But this is needed as supporting {CL_BGRA, CL_UNORM_INT8} is required by OpenCL.

In that case we could just use VK_FORMAT_R8G8B8A8_ which is supported and swizzle the component to have everything as expected for OpenCL.

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -4712,10 +4712,10 @@ std::unordered_map<cl_image_format, image_format_support, ClFormatMapHash,
         {{CL_RGBA, CL_FLOAT}, VK_FORMAT_R32G32B32A32_SFLOAT},
 
         // BGRA formats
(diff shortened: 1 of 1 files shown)


rjodinchr:
The question then would be, what would be a proper way to integrate that?
- Just merged this path
- Have a config per device to should whether to use BGRA or not.
- Detect dynamically that this translation is needed

I am not sure of the cost of swizzling. If this is minor, maybe just merging it as it would be a good enough solution.

rjodinchr:
I have just realised that this implementation does not work with samplerless accesses.
First idea coming to me would be to add a literal sampler to avoid samplerless accesses.

PR #609, opened 2023-09-12
add support for image1d_buffer

rjodinchr:
This implementation has been tested using the OpenCL-CTS patch with https://github.com/KhronosGroup/OpenCL-CTS/pull/1806

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -710,9 +710,7 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
         size_ret = sizeof(val_uint);
         break;
     case CL_DEVICE_IMAGE_MAX_BUFFER_SIZE:
-        val_sizet =
-            std::min((uint64_t)device->vulkan_limits().maxTexelBufferElements,
-                     device->max_mem_alloc_size());
+        val_sizet = device->image_max_buffer_size();
         copy_ptr = &val_sizet;
         size_ret = sizeof(val_sizet);
         break;
@@ -3577,9 +3575,9 @@ cl_int CLVK_API_CALL clEnqueueFillBuffer(
 
     // TODO check sub-buffer alignment
 
-    auto cmd = new cvk_command_fill_buffer(command_queue,
-                                           static_cast<cvk_buffer*>(buffer),
-                                           offset, size, pattern, pattern_size);
+    auto cmd = new cvk_command_fill_buffer(
+        command_queue, static_cast<cvk_buffer*>(buffer), offset, size, pattern,
+        pattern_size, CL_COMMAND_FILL_BUFFER);
 
     return command_queue->enqueue_command_with_deps(
         cmd, num_events_in_wait_list, event_wait_list, event);
@@ -3725,8 +3723,9 @@ void* cvk_enqueue_map_buffer(cvk_command_queue* cq, cvk_buffer* buffer,
                              cl_map_flags map_flags,
                              cl_uint num_events_in_wait_list,
                              const cl_event* event_wait_list, cl_event* event,
-                             cl_int* errcode_ret) {
-    auto cmd = new cvk_command_map_buffer(cq, buffer, offset, size, map_flags);
+                             cl_int* errcode_ret, cl_command_type type) {
+    auto cmd =
+        new cvk_command_map_buffer(cq, buffer, offset, size, map_flags, type);
 
     void* map_ptr;
     cl_int err = cmd->build(&map_ptr);
@@ -3830,7 +3829,8 @@ void* CLVK_API_CALL clEnqueueMapBuffer(cl_command_queue cq, cl_mem buf,
     cl_int err;
     auto map_ptr = cvk_enqueue_map_buffer(
         command_queue, buffer, blocking_map, offset, size, map_flags,
-        num_events_in_wait_list, event_wait_list, event, &err);
+        num_events_in_wait_list, event_wait_list, event, &err,
+        CL_COMMAND_MAP_BUFFER);
 
     if (errcode_ret != nullptr) {
         *errcode_ret = err;
@@ -4890,6 +4890,16 @@ cl_int cvk_enqueue_image_copy(
     size_t slice_pitch, void* ptr, cl_uint num_events_in_wait_list,
     const cl_event* event_wait_list, cl_event* event) {
 
+    auto img = static_cast<cvk_image*>(image);
+    if (img->is_backed_by_buffer_view()) {
+        auto cmd = new cvk_command_buffer_host_copy(
+            queue, command_type, static_cast<cvk_buffer*>(img->buffer()), ptr,
+            origin[0] * img->element_size(), region[0] * img->element_size());
+        auto err = queue->enqueue_command_with_deps(
+            cmd, blocking, num_events_in_wait_list, event_wait_list, event);
+        return err;
+    }
+
     // Create image map command
     std::array<size_t, 3> orig = {origin[0], origin[1], origin[2]};
     std::array<size_t, 3> reg = {region[0], region[1], region[2]};
@@ -4901,7 +4911,6 @@ cl_int cvk_enqueue_image_copy(
         map_flags = CL_MAP_READ;
     }
 
-    auto img = static_cast<cvk_image*>(image);
     auto cmd_map = std::make_unique<cvk_command_map_image>(queue, img, orig,
                                                            reg, map_flags);
     void* map_ptr;
@@ -5005,23 +5014,10 @@ cl_int CLVK_API_CALL clEnqueueReadImage(
         return CL_INVALID_OPERATION;
     }
 
-    auto internal_image = static_cast<cvk_image*>(image);
-    if (internal_image->is_backed_by_buffer_view()) {
-        auto cmd = new cvk_command_buffer_host_copy(
-            command_queue, CL_COMMAND_READ_BUFFER,
-            static_cast<cvk_buffer*>(internal_image->buffer()), ptr,
-            origin[0] * internal_image->element_size(),
-            region[0] * internal_image->element_size());
-        auto err = command_queue->enqueue_command_with_deps(
-            cmd, blocking_read, num_events_in_wait_list, event_wait_list,
-            event);
-        return err;
-    } else {
-        return cvk_enqueue_image_copy(
-            command_queue, CL_COMMAND_READ_IMAGE, image, blocking_read, origin,
-            region, row_pitch, slice_pitch, ptr, num_events_in_wait_list,
-            event_wait_list, event);
-    }
+    return cvk_enqueue_image_copy(command_queue, CL_COMMAND_READ_IMAGE, image,
+                                  blocking_read, origin, region, row_pitch,
+                                  slice_pitch, ptr, num_events_in_wait_list,
+                                  event_wait_list, event);
 }
 
 cl_int CLVK_API_CALL clEnqueueWriteImage(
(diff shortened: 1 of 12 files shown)


rjodinchr:
This PR depends on an update of clspv

src/api.cpp
```diff
@@ -710,7 +710,9 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
         size_ret = sizeof(val_uint);
         break;
     case CL_DEVICE_IMAGE_MAX_BUFFER_SIZE:
-        val_sizet = device->vulkan_limits().maxImageDimension1D;
+        val_sizet =
+            std::min((uint64_t)device->vulkan_limits().maxTexelBufferElements,
+                     device->max_mem_alloc_size());
```
kpet, line 715: Let's push this into a `cvk_device::image_max_buffer_size` function.

src/api.cpp
```diff
@@ -4809,29 +4841,37 @@ cl_int CLVK_API_CALL clGetSupportedImageFormats(cl_context context,
...
         if ((features & required_format_feature_flags) ==
             required_format_feature_flags) {
+            VkComponentMapping components_sampled, components_storage;
+            get_component_mappings_for_channel_order(clfmt.image_channel_order,
+                                                     &components_sampled,
+                                                     &components_storage);
+            if ((components_sampled != components_storage) &&
+                (image_type == CL_MEM_OBJECT_IMAGE1D_BUFFER)) {
+                continue;
+            }
```
kpet, line 4863: I'm a bit confused by this condition. Those swizzles are unused for 1D buffer images. Why is this needed?

src/api.cpp
```diff
@@ -4965,10 +5005,23 @@ cl_int CLVK_API_CALL clEnqueueReadImage(
...
     }
 
-    return cvk_enqueue_image_copy(command_queue, CL_COMMAND_READ_IMAGE, image,
-                                  blocking_read, origin, region, row_pitch,
-                                  slice_pitch, ptr, num_events_in_wait_list,
-                                  event_wait_list, event);
+    auto internal_image = static_cast<cvk_image*>(image);
+    if (internal_image->is_backed_by_buffer_view()) {
+        auto cmd = new cvk_command_buffer_host_copy(
+            command_queue, CL_COMMAND_READ_BUFFER,
```
kpet, line 5011:
```suggestion
            command_queue, CL_COMMAND_READ_IMAGE,
```

Otherwise, the application could see the wrong command type.

src/api.cpp
```diff
@@ -5028,11 +5081,24 @@ cl_int CLVK_API_CALL clEnqueueWriteImage(
...
+    if (internal_image->is_backed_by_buffer_view()) {
+        auto cmd = new cvk_command_buffer_host_copy(
+            command_queue, CL_COMMAND_WRITE_BUFFER,
+            static_cast<cvk_buffer*>(internal_image->buffer()), ptr,
+            origin[0] * internal_image->element_size(),
+            region[0] * internal_image->element_size());
+        auto err = command_queue->enqueue_command_with_deps(
+            cmd, blocking_write, num_events_in_wait_list, event_wait_list,
+            event);
+        return err;
```
kpet, line 5094: I think you probably want to push this code to `cvk_enqueue_image_copy` so it's the same for read image and write image commands.

src/api.cpp
```diff
@@ -5102,18 +5168,49 @@ clEnqueueCopyImage(cl_command_queue cq, cl_mem src_image, cl_mem dst_image,
     if (!src_img->has_same_format(dst_img)) {
         return CL_IMAGE_FORMAT_MISMATCH;
     }
+    assert(src_img->element_size() == dst_img->element_size());
```
kpet, line 5171: This should not be an assert. OpenCL specifies the following error condition:
>CL_IMAGE_FORMAT_MISMATCH if src_image and dst_image do not use the same image format.

src/api.cpp
```diff
@@ -5102,18 +5168,49 @@ clEnqueueCopyImage(cl_command_queue cq, cl_mem src_image, cl_mem dst_image,
...
     std::array<size_t, 3> dst_orig = {dst_origin[0], dst_origin[1],
                                       dst_origin[2]};
     std::array<size_t, 3> reg = {region[0], region[1], region[2]};
 
-    auto cmd = std::make_unique<cvk_command_image_image_copy>(
-        command_queue, src_img, dst_img, src_orig, dst_orig, reg);
+    if (src_img->is_backed_by_buffer_view() &&
+        dst_img->is_backed_by_buffer_view()) {
+        auto cmd = new cvk_command_copy_buffer(
+            command_queue, CL_COMMAND_COPY_BUFFER,
```
kpet, line 5182:
```suggestion
            command_queue, CL_COMMAND_COPY_IMAGE,
```

That's what the application should see.

src/api.cpp
```diff
@@ -5102,18 +5168,49 @@ clEnqueueCopyImage(cl_command_queue cq, cl_mem src_image, cl_mem dst_image,
...
+            static_cast<cvk_buffer*>(dst_img->buffer()),
+            src_origin[0] * src_img->element_size(),
+            dst_origin[0] * dst_img->element_size(),
+            region[0] * src_img->element_size());
+
+        return command_queue->enqueue_command_with_deps(
+            cmd, num_events_in_wait_list, event_wait_list, event);
+    } else if (src_img->is_backed_by_buffer_view()) {
+        auto cmd = std::make_unique<cvk_command_buffer_image_copy>(
+            CL_COMMAND_COPY_BUFFER_TO_IMAGE, command_queue,
```
kpet, line 5193:
```suggestion
            CL_COMMAND_COPY_IMAGE, command_queue,
```

src/api.cpp
```diff
@@ -5102,18 +5168,49 @@ clEnqueueCopyImage(cl_command_queue cq, cl_mem src_image, cl_mem dst_image,
...
+            static_cast<cvk_buffer*>(src_img->buffer()), dst_img,
+            src_origin[0] * src_img->element_size(), dst_orig, reg);
+
+        return command_queue->enqueue_command_with_deps(
+            cmd.release(), num_events_in_wait_list, event_wait_list, event);
+    } else if (dst_img->is_backed_by_buffer_view()) {
+        auto cmd = std::make_unique<cvk_command_buffer_image_copy>(
+            CL_COMMAND_COPY_IMAGE_TO_BUFFER, command_queue,
```
kpet, line 5201:
```suggestion
            CL_COMMAND_COPY_IMAGE, command_queue,
```

src/api.cpp
```diff
@@ -5168,11 +5265,27 @@ cl_int CLVK_API_CALL clEnqueueFillImage(
...
 
+    auto img = static_cast<cvk_image*>(image);
+
+    // Fill
+    cvk_image::fill_pattern_array pattern;
+    size_t pattern_size;
+    img->prepare_fill_pattern(fill_color, pattern, &pattern_size);
+
+    if (img->is_backed_by_buffer_view()) {
+        auto cmd = new cvk_command_fill_buffer(
```
kpet, line 5276: This will expose the wrong command type to applications.

src/api.cpp
```diff
@@ -5279,12 +5387,23 @@ cl_int CLVK_API_CALL clEnqueueCopyImageToBuffer(
                                     src_origin[2]};
     std::array<size_t, 3> reg = {region[0], region[1], region[2]};
 
-    auto cmd = std::make_unique<cvk_command_buffer_image_copy>(
-        CL_COMMAND_COPY_IMAGE_TO_BUFFER, command_queue, buffer, image,
-        dst_offset, origin, reg);
+    if (image->is_backed_by_buffer_view()) {
+        auto cmd = new cvk_command_copy_buffer(
+            command_queue, CL_COMMAND_COPY_BUFFER,
```
kpet, line 5392: Wrong command type

src/api.cpp
```diff
@@ -5355,12 +5474,23 @@ cl_int CLVK_API_CALL clEnqueueCopyBufferToImage(
                                     dst_origin[2]};
     std::array<size_t, 3> reg = {region[0], region[1], region[2]};
 
-    auto cmd = std::make_unique<cvk_command_buffer_image_copy>(
-        CL_COMMAND_COPY_BUFFER_TO_IMAGE, command_queue, buffer, image,
-        src_offset, origin, reg);
+    if (image->is_backed_by_buffer_view()) {
+        auto cmd = new cvk_command_copy_buffer(
+            command_queue, CL_COMMAND_COPY_BUFFER, buffer,
```
kpet, line 5479: Same here

src/api.cpp
```diff
@@ -5499,10 +5629,20 @@ void* CLVK_API_CALL clEnqueueMapImage(
...
 
     cl_int err;
-    auto ret = cvk_enqueue_map_image(command_queue, image, blocking_map,
-                                     map_flags, origin, region, image_row_pitch,
-                                     image_slice_pitch, num_events_in_wait_list,
-                                     event_wait_list, event, &err);
+    void* ret;
+    auto img = static_cast<cvk_image*>(image);
+    if (img->is_backed_by_buffer_view()) {
+        ret = cvk_enqueue_map_buffer(
```
kpet, line 5635: Wrong command type

kpet (changes requested):
This change does not introduce CTS regressions at my end. As discussed I'm not a big fan of handling the buffer view special case in the API layer as this breaks encapsulation. In the interest of getting forward progress, I'm with this approach as a first step (and I think it likely makes the code a bit shorter).

src/memory.cpp
```diff
@@ -485,6 +490,55 @@ bool cvk_image::init() {
...
+
+    auto vkbuf = static_cast<cvk_buffer*>(buffer())->vulkan_buffer();
+
+    VkBufferViewCreateInfo createInfo = {
+        VK_STRUCTURE_TYPE_BUFFER_VIEW_CREATE_INFO,
+        nullptr,
+        0,            // flags
+        vkbuf,        // buffer
+        format,       // format
+        0,            // offset
```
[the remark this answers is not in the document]
rjodinchr: I've tried to add something in the CTS ( in `kernel_read_write`), but it just got too complicated and got lost in this code. So I have just added a test for it in clvk.

src/api.cpp

src/api.cpp
rjodinchr: I am not sure, this is coming from your initial commit I believe. Thinking about it, I don't see a reason to have it, maybe we should get rid of it?

src/api.cpp

src/api.cpp

src/api.cpp

src/api.cpp
rjodinchr: I have added a `cl_command_type` in `cvk_command_buffer_image_copy` so that we have the reported command type as well as the one specifying what copy to perform.

src/api.cpp
rjodinchr: same as above

src/api.cpp
rjodinchr: Fixed

src/api.cpp
rjodinchr: Fixed

src/api.cpp
rjodinchr: Fixed

src/api.cpp
rjodinchr: Fixed

kpet:
@rjodinchr In case you hadn't noticed: it looks like the `1DBufferImageFromSubBuffer` test is hitting an assert.

rjodinchr:
The tests are expected to fail one way or another as this PR is not updating clspv.

kpet:
Ok, makes sense. Can you include the clspv update here (just the one commit we need please) so the CI passes? Thanks!

src/api.cpp
kpet: yeah, if we pass the CTS without it, then we should remove it. If we somehow need this to pass the CTS, we should find a clearer expression of the conditions.

src/queue.hpp
```diff
@@ -957,6 +967,7 @@ struct cvk_command_buffer_image_copy final : public cvk_command_batchable {
     size_t m_offset;
     std::array<size_t, 3> m_origin;
     std::array<size_t, 3> m_region;
+    cl_command_type m_copy_type;
```
kpet, line 970: Reusing `cl_command_type` is a bit confusing but it'll do for a first implementation.

kpet (commented):
Almost there. I think we just need to resolve the one open thread about component mapping and update clspv (just the one commit this PR needs) so the CI can pass. Everything else looks nice!

src/memory.cpp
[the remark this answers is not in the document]
kpet: No problem. The unit test is enough.

kpet:
@rjodinchr I've had a look at the CI failures. The newly added test creates a sub-buffer at an offset that does not satisfy Vulkan constraints. clvk should have rejected the creation of the sub-buffer and texel buffer alignment constraints should be taken into account when calculating `CL_DEVICE_MEM_BASE_ADDR_ALIGN`. I suggest we don't fix sub-buffer error handling in this PR. In the interest of getting forward progress, you could just modify the offsets used in the test so that it works in the CI or query `CL_DEVICE_MEM_BASE_ADDR_ALIGN`, use it to pick the sub-buffer offset and change its implementation to take `storageTexelBufferOffsetAlignmentBytes`, `storageTexelBufferOffsetSingleTexelAlignment`, `uniformTexelBufferOffsetAlignmentBytes`, and `uniformTexelBufferOffsetSingleTexelAlignment` into account.

kpet (commented):
It seems swiftshader does not support `VK_FORMAT_FEATURE_2_STORAGE_WRITE_WITHOUT_FORMAT_BIT_KHR` on any format for texel buffers so you're going to have to disable the test for swiftshader.

src/api.cpp
rjodinchr: It seems that this test is needed. Otherwise the CTS thinks that we support `CL_INTENSITY` and `CL_LUMINANCE` with 1d buffer, which does not work with the actual implementation. As they are not mandatory, it feels ok to remove them with this test. What do you think?

kpet: If they're not supported we should remove them but the condition used to do that is a hack (you said it was part of my draft patch, I'm assuming I'd done that as a quick way of excluding those formats). I've added an entry to #625 so we don't forget to revisit this.

kpet:
Merging this as is to unblock clspv updates. I will fix the CI with a separate commit on top.

PR #610, opened 2023-09-15
remove unnecessary eol in cvk_debug_fn

rjodinchr:


kpet (approved):
Thanks! The message is not very clear, I'll change it with a suggestion before merging this PR.

src/device.hpp
```diff
@@ -133,7 +133,7 @@ struct cvk_device : public _cl_device_id,
             auto k = memory_type_index_for_resource(valid_memory_type_bits,
                                                     supported_memory_types[i]);
             if (k != VK_MAX_MEMORY_TYPES) {
-                cvk_debug_fn("outer returning %u\n", k);
+                cvk_debug_fn("outer returning %u", k);
```
kpet, line 136:
```suggestion
                cvk_debug_fn("selected %u", k);
```

PR #611, opened 2023-09-16
fix print when forgetting vector modifier

rjodinchr:
According to the OpenCL specification, when printing a vector if the modifier is not used, it is an undefined behavior. The current implement prints the first element only: "%v4u", (1, 2, 3, 4) => '1, 1, 1, 1'

But in our case we can figure out the element size using information coming from clspv. Use it only when we are in this undefined behavior to print something making more sense:
"%v4u", (1, 2, 3, 4) => '1, 2, 3, 4'

### DIFF
--- a/src/printf.cpp
+++ b/src/printf.cpp
@@ -217,6 +217,13 @@ void process_printf(char*& data, const printf_descriptor_map_t& descs,
             data += size;
         } else {
             // Vector argument
+            if (el_size == 0) {
+                // 'ele_size == 0' means that no modifier has been used.
+                // According to the spec, this is an undefined behavior. Let's
+                // use the size coming from clspv and the vec_len to figure out
+                // the element size then.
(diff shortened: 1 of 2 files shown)


src/printf.cpp
```diff
@@ -217,6 +217,13 @@ void process_printf(char*& data, const printf_descriptor_map_t& descs,
             data += size;
         } else {
             // Vector argument
+            if (el_size == 0) {
+                // 'ele_size == 0' means that no modifier has been used.
```
kpet, line 221:
```suggestion
                // 'ele_size == 0' means that no length modifier has been used.
```

kpet (changes requested):
The change looks good. Agree this is a more helpful behaviour in that case. Since this cannot be covered by the CTS, we should add unit tests in clvk.

kpet (approved):
Thanks!

PR #612, opened 2023-09-19
fix cvk_command_image_init for 3D images

rjodinchr:


kpet (commented):
This looks like an obvious fix which makes me suspicious as to why it wasn't done like that in the first place. Maybe the 1 was used for debugging and changing that back was missed when merging to main. At least we clearly don't have adequate test coverage :). How did you find it?

rjodinchr:
I find it working on `samplerless_reads`. It seems that this is almost the only test using this feature.

But it is even more complicated than that. Because the test is not comparing the value in the image on the host side. It is just comparing the values coming from a sampler read and a samplerless reads. When we do not initialise the image, we just have garbage values there. The same comes from the sampler read and the samplerless one.

But on AMD when you read with a sampler, all NaN are encoded as `oxffc00000` whatever the real Nan value is (maybe `0x7fc00023` for example). Making the comparison fail.
Working on fixing the test for AMD, I finally found that it was due to this bug.

rjodinchr:
Note that #614 will add coverage for it in clvk.

kpet:
Ouch, that's a bit of a gap in the CTS. #614 will get us one test but the CTS should probably do more. I've created https://github.com/KhronosGroup/OpenCL-CTS/issues/1823.

kpet (approved):

PR #614, opened 2023-09-20
fix support for read of 3D images with unnormalised sampler

rjodinchr:
This PR depends on google/clspv#1234

We parse the new NormalizedSamplerMaskPushConstant non-semantic instruction from KhronosGroup/SPIRV-Headers#377.

Set the sampler mask in the push constant, but also create a new sampler with normalised coordinate if not already the case or already created.

### DIFF
--- a/src/kernel.cpp
+++ b/src/kernel.cpp
@@ -14,6 +14,8 @@
 
 #include <algorithm>
 
+#include "clspv/Sampler.h"
+
 #include "kernel.hpp"
 #include "memory.hpp"
 
@@ -42,6 +44,10 @@ cl_int cvk_kernel::init() {
 
     if (const auto* md = m_entry_point->image_metadata()) {
         m_image_metadata = md;
+    }
+
+    if (const auto* md = m_entry_point->sampler_metadata()) {
+        m_sampler_metadata = md;
     }
 
     // Init argument values
@@ -100,6 +106,45 @@ void cvk_kernel::set_image_metadata(cl_uint index, const void* image) {
     }
 }
 
+void cvk_kernel::set_sampler_metadata(cl_uint index, const void* sampler) {
+    if (!m_sampler_metadata) {
+        return;
+    }
+    auto md = m_sampler_metadata->find(index);
+    if (md != m_sampler_metadata->end()) {
+        auto apisampler = *reinterpret_cast<const cl_sampler*>(sampler);
+        auto offset = md->second;
+        auto sampler = icd_downcast(apisampler);
+        uint32_t sampler_mask = (sampler->normalized_coords()
+                                     ? clspv::CLK_NORMALIZED_COORDS_TRUE
+                                     : clspv::CLK_NORMALIZED_COORDS_FALSE) |
+                                (sampler->filter_mode() == CL_FILTER_NEAREST
+                                     ? clspv::CLK_FILTER_NEAREST
+                                     : clspv::CLK_FILTER_LINEAR);
+        switch (sampler->addressing_mode()) {
+        case CL_ADDRESS_NONE:
+            sampler_mask |= clspv::CLK_ADDRESS_NONE;
+            break;
+        case CL_ADDRESS_CLAMP:
+            sampler_mask |= clspv::CLK_ADDRESS_CLAMP;
+            break;
+        case CL_ADDRESS_REPEAT:
+            sampler_mask |= clspv::CLK_ADDRESS_REPEAT;
+            break;
+        case CL_ADDRESS_CLAMP_TO_EDGE:
+            sampler_mask |= clspv::CLK_ADDRESS_CLAMP_TO_EDGE;
+            break;
+        case CL_ADDRESS_MIRRORED_REPEAT:
+            sampler_mask |= clspv::CLK_ADDRESS_MIRRORED_REPEAT;
+            break;
+        default:
+            break;
+        }
+        m_argument_values->set_pod_data(offset, sizeof(sampler_mask),
+                                        &sampler_mask);
+    }
+}
+
 cl_int cvk_kernel::set_arg(cl_uint index, size_t size, const void* value) {
     std::lock_guard<std::mutex> lock(m_lock);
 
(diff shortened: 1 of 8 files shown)


rjodinchr:
Ref https://github.com/google/clspv/issues/1202

src/kernel.cpp
```diff
@@ -100,6 +106,26 @@ void cvk_kernel::set_image_metadata(cl_uint index, const void* image) {
...
+    if (md != m_sampler_metadata->end()) {
+        auto apisampler = *reinterpret_cast<const cl_sampler*>(sampler);
+        auto offset = md->second;
+        auto sampler = icd_downcast(apisampler);
+        uint32_t sampler_mask = (sampler->normalized_coords()
+                                     ? clspv::CLK_NORMALIZED_COORDS_TRUE
+                                     : clspv::CLK_NORMALIZED_COORDS_FALSE) |
+                                (sampler->filter_mode() == CL_FILTER_NEAREST
+                                     ? clspv::CLK_FILTER_NEAREST
+                                     : clspv::CLK_FILTER_LINEAR);
```
alan-baker, line 123: Seems like this is missing the addressing mode.

rjodinchr: Yes, I was a bit lazy here. As it is not used by clspv, I did not put it here.

src/kernel.hpp
```diff
@@ -237,7 +245,8 @@ struct cvk_kernel_argument_values {
         }
 
         if (m_entry_point->has_pod_arguments() ||
-            m_entry_point->has_image_metadata()) {
+            m_entry_point->has_image_metadata() ||
+            m_entry_point->has_sampler_metadata()) {
```
kpet, line 249: It's really time we refactor push constant management to make it more readable. In a lot of places, they're still named "POD". Best done as a standalone changes.

src/memory.cpp
```diff
@@ -158,7 +158,7 @@ cvk_sampler::create(cvk_context* context, bool normalized_coords,
     return sampler.release();
 }
 
-bool cvk_sampler::init() {
+bool cvk_sampler::init(bool norm) {
```
kpet, line 161: s/norm/force_normalized_coordinates/ (or better name). I think we should use a more descriptive name.

src/memory.hpp
```diff
@@ -428,13 +428,18 @@ struct cvk_sampler : public _cl_sampler, api_object<object_magic::sampler> {
...
+          m_properties(std::move(properties)), m_sampler(VK_NULL_HANDLE),
+          m_sampler_norm(VK_NULL_HANDLE) {}
 
     ~cvk_sampler() {
         if (m_sampler != VK_NULL_HANDLE) {
             auto vkdev = context()->device()->vulkan_device();
             vkDestroySampler(vkdev, m_sampler, nullptr);
         }
+        if (m_sampler_norm != VK_NULL_HANDLE) {
+            auto vkdev = context()->device()->vulkan_device();
```
kpet, line 440: Maybe that line could be common and outside the if's?

src/memory.hpp
```diff
@@ -453,17 +458,26 @@ struct cvk_sampler : public _cl_sampler, api_object<object_magic::sampler> {
     cl_addressing_mode addressing_mode() const { return m_addressing_mode; }
     cl_filter_mode filter_mode() const { return m_filter_mode; }
     VkSampler vulkan_sampler() const { return m_sampler; }
+    VkSampler vulkan_sampler_norm() {
+        if (m_sampler_norm == VK_NULL_HANDLE) {
+            if (!init(true)) {
+                return VK_NULL_HANDLE;
+            }
+        }
```
kpet, line 466: I don't really like this idiom as currently written. Having what looks like a getter reinitialise the sampler object breaks the principle of least surprise. Since we can't know at sampler creation time whether we need to force enable normalized coordinates we can either:
1. Always create both samplers. It's a bit wasteful but probably not a practical problem.
2. Give this function a more explicit name. Something along the lines of `get_or_create_vulkan_sampler_with_normalized_coords` (or better name).

src/program.cpp
```diff
@@ -1963,32 +1982,49 @@ cl_int cvk_entry_point::init() {
...
-                    std::min(data_type_offset, push_constant_range.offset);
-                if (data_type_offset + sizeof(uint32_t) >
+                    std::min(offset, push_constant_range.offset);
+                if (offset + sizeof(uint32_t) >
                     push_constant_range.offset + push_constant_range.size) {
-                    push_constant_range.size = data_type_offset +
-                                               sizeof(uint32_t) -
-                                               push_constant_range.offset;
+                    push_constant_range.size =
+                        offset + sizeof(uint32_t) - push_constant_range.offset;
```
kpet, line 2027: This code is now screaming "refactor me" :). To be done as a standalone change.

src/kernel.cpp
```diff
@@ -264,7 +294,17 @@ bool cvk_kernel_argument_values::setup_descriptor_sets() {
         }
         case kernel_argument_kind::sampler: {
             auto clsampler = static_cast<cvk_sampler*>(get_arg_value(arg));
-            auto sampler = clsampler->vulkan_sampler();
+            bool sampler_norm = false;
```
kpet, line 297:
```suggestion
            bool normalized_coord_sampler_required = false;
```

kpet (changes requested):
Fine with the approach. A couple of (suggested) name tweaks would make the code much easier to read and maintain.

src/kernel.cpp
rjodinchr: Done

src/memory.cpp
rjodinchr: Done

src/memory.hpp
rjodinchr: Done

src/memory.hpp
rjodinchr: I went with option 2.

src/kernel.cpp
rjodinchr: Done

rjodinchr:
This PR will need to have clspv and spirv-headers updated to go it

alan-baker (approved):
LGTM

kpet (approved):
LGTM. To be able to update clspv, we are going to need to get https://github.com/kpet/clvk/pull/609 in first.

PR #617, opened 2023-10-09
compile with --undefined-version when not using CLVK_UNIT_TESTING

rjodinchr:


rjodinchr:
Tests are failing because this option exists and is needed only with very recent versions of clang/lld

kpet:
It looks like we're going to need to detect whether the option is supported before using it. Alternatively, we could maybe explore using a separate `exports.map` file, ideally without duplicating the OpenCL symbols. Maybe we could generate it at build time from separate snippets if there isn't any support for inclusions etc?

rjodinchr:
I did not find a nice way to do that. For the time being I am just changing my cmake command for clvk to add it there, which make more sense as it is due to the compiler I am using right now.

PR #618, opened 2023-10-11
fix initialization of 2d and 3d image

rjodinchr:
After #612 the cts started to regress on swiftshader. It is because in the cts, the height and depth can be set to zero when the dimension is not used.

### DIFF
--- a/src/queue.cpp
+++ b/src/queue.cpp
@@ -1804,6 +1804,21 @@ cvk_command_image_init::build_batchable_inner(cvk_command_buffer& cmdbuf) {
         extent.height = m_image->height();
         extent.depth = m_image->depth();
 
+        switch (m_image->type()) {
+        case CL_MEM_OBJECT_IMAGE2D:
+        case CL_MEM_OBJECT_IMAGE2D_ARRAY:
+            extent.depth = 1;
+            break;
+        case CL_MEM_OBJECT_IMAGE1D_BUFFER:
+        case CL_MEM_OBJECT_IMAGE1D:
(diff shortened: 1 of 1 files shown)


rjodinchr:
The error with Android is with the infra:
```
/usr/bin/bash -xc env SCCACHE_IDLE_TIMEOUT=0 SCCACHE_DIR='/home/runner/work/clvk/clvk'/.sccache SCCACHE_CACHE_SIZE='2G' sccache --start-server
  + '[' -z '' ']'
  + return
  + case $- in
  + return
  + env SCCACHE_IDLE_TIMEOUT=0 SCCACHE_DIR=/home/runner/work/clvk/clvk/.sccache SCCACHE_CACHE_SIZE=2G sccache --start-server
  sccache: Starting the server...
  sccache: error: Timed out waiting for server startup
  Error: Restoring cache failed: Error: The process '/usr/bin/bash' failed with exit code 2
```

rjodinchr:
@kpet , this is blocking google infra, could you have a look please?
Thanks

rjodinchr:
> It's probably why the code was the way it was prior to [d2c0cd0](https://github.com/kpet/clvk/commit/d2c0cd0b17990f2ab4dad10c631ade6c695cf714). Why is this showing up only in Swiftshader runs?

I'm not sure. Maybe swiftshader is actually using `extend.depth` whatever the image type is? thus having a zero there would mess everything up.

kpet:
Not sure. The change looks good so I'll just merge this to unblock you.

rjodinchr:
thank you

PR #621, opened 2023-10-16
implement dynamic managment of batch parameters

rjodinchr:
Add an option to manage dynamicaly batch parameters. The aim is to get the best parameter for an application. Because many things can happen in the system, we are always updating to take any change of load in account.
This algorithm has been tested on chromeos with various workloads and produce great improvement performance wise.

This new option is disabled by default.

### DIFF
--- a/src/config.def
+++ b/src/config.def
@@ -47,6 +47,9 @@ OPTION(uint32_t, max_first_cmd_batch_size, 10000u)
 OPTION(uint32_t, max_cmd_group_size, UINT32_MAX)
 OPTION(uint32_t, max_first_cmd_group_size, UINT32_MAX)
 
+// experimental
+OPTION(bool, dynamic_batches, false)
+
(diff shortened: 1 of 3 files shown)


kpet (commented):
I really like the idea but, as discussed, I think the implementation could use a few improvements:
- I think we should encapsulate the control logic into a separate object such that they are easily pluggable and we have a way of grouping state that is specific to a given implementation of the control logic alongside that implementation. The first idea that comes to mind is to introduce a `cvk_command_queue_controller` object that'd be a friend of `cvk_command_queue` so that it can access and update its state. Command queue objects could optionally create a controller when dynamic control is enabled. 
- I think you should document the strategy behind your control algorithm and maybe comment the code a bit more. It's hard(er than it should) to build the big picture just from reading the code.

What do you think? Happy to discuss this further, F2F or here.

rjodinchr:
I still need to document a the strategy implemented here.

kpet (approved):

PR #622, opened 2023-10-18
remove fabs from native_builtins for intel and AMD

rjodinchr:
Both Intel and AMD GLSL fabs version flush denormalized numbers to zero.
This will prevent OpenCL compliance when clspv will be updated with google/clspv#1252 as functions like cbrt from libclc requires denormalized numbers not to be flushed to zero to be compliant.


kpet (approved):

rjodinchr:
Please wait before merging, this might not be needed in the end: https://github.com/google/clspv/pull/1256

rjodinchr:
This is needed in the end. Too many things can open leading to illegal code generated (inlining of the builtin for example)

kpet (approved):

Issue #625, opened 2023-10-24
Tidy up support for `CL_MEM_OBJECT_IMAGE1D_BUFFER`


kpet:
A few things to address after the initial PR:
- [ ] Rewrite conditions to exclude `CL_LUMINANCE` and `CL_INTENSITY` formats. See https://github.com/kpet/clvk/pull/609#discussion_r1359888301
- [ ] Handle differences between images backed by a Vulkan image or texel buffers in command objects instead of the API layer
- [ ] Require `VK_FORMAT_FEATURE_2_STORAGE_WRITE_WITHOUT_FORMAT_BIT_KHR` for writable texel buffers. Probably depends on https://github.com/kpet/clvk/issues/497
- [ ] Change implementation of `CL_DEVICE_MEM_BASE_ADDR_ALIGN` to take into account alignment requirements of texel buffers. See https://github.com/kpet/clvk/pull/609#issuecomment-1775271716
- [ ] Test 1D images created from a sub-buffer in the CTS

PR #629, opened 2023-10-30
add fixup for image1dbuffer with bgra format on Intel

rjodinchr:
also fix condition for CL_LUMINANCE and CL_INTENSITY (ref #625)

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -14,6 +14,7 @@
 
 #include "cl_headers.hpp"
 #include "icd.hpp"
+#include "image_format.hpp"
 #include "init.hpp"
 #include "kernel.hpp"
 #include "log.hpp"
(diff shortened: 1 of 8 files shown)


src/api.cpp
```diff
@@ -4820,56 +4674,50 @@ cl_int CLVK_API_CALL clGetSupportedImageFormats(cl_context context,
...
+    const cl_channel_type channel_data_types[] = {
+        CL_SNORM_INT8,       CL_SNORM_INT16,        CL_UNORM_INT8,
+        CL_UNORM_INT16,      CL_UNORM_SHORT_565,    CL_UNORM_SHORT_555,
+        CL_UNORM_INT_101010, CL_SIGNED_INT8,        CL_SIGNED_INT16,
+        CL_SIGNED_INT32,     CL_UNSIGNED_INT8,      CL_UNSIGNED_INT16,
+        CL_UNSIGNED_INT32,   CL_HALF_FLOAT,         CL_FLOAT,
+        CL_UNORM_INT24,      CL_UNORM_INT_101010_2,
+    };
+    for (auto const image_channel_order : channel_orders) {
+        for (auto const image_channel_data_type : channel_data_types) {
```
kpet, line 4692: I'm not sure I understand why you changed the iteration logic. With the current tables, these loops will iterate over 340 combinations, regardless of the number of CL <-> Vulkan mappings. Adding a new format with a new channel order or data type now requires a change in two places. Is this just because the format map is no longer visible? If that's the case we could probably introduce a getter for it.

src/device_properties.cpp
```diff
@@ -105,7 +105,12 @@ struct cvk_device_properties_intel : public cvk_device_properties {
         });
     }
     std::string get_compile_options() const override final {
-        return "-hack-mul-extended -hack-convert-to-float";
+        return "-hack-mul-extended -hack-convert-to-float "
+               "-hack-image1d-buffer-bgra";
```
kpet, line 109: Not directly related to this change but some of these limitations are presumably caused by specific drivers rather than tied to specific devices. We will probably want to capture/represent the distinction.

src/api.cpp
```diff
@@ -4590,195 +4591,47 @@ cl_int CLVK_API_CALL clGetImageInfo(cl_mem image, cl_image_info param_name,
...
+        features = properties.bufferFeatures;
+    } else {
+        // TODO support linear tiling
+        features = properties.optimalTilingFeatures;
+    }
+    if ((features & required_format_feature_flags) ==
+        required_format_feature_flags) {
+        if ((image_channel_order == CL_LUMINANCE ||
+             image_channel_order == CL_INTENSITY) &&
+            (image_type == CL_MEM_OBJECT_IMAGE1D_BUFFER)) {
```
kpet, line 4629: It still doesn't capture the reason but at least the intent is now very clear. That's a step up.

kpet (commented):
Thanks for the change! Good to start extracting all the image tables, etc.

src/api.cpp
rjodinchr: I don't know more than `IMAGE1D_BUFFER` and `CL_LUMINANCE` or `CL_INTENSITY` are not supported at the moment. I could add something like that in a comment, but I'm not sure that's adding anything.

src/device_properties.cpp
rjodinchr: I agree, that would need to be changed at some point.

src/api.cpp
rjodinchr: I have introduced a getter for the map.

src/api.cpp
kpet: Same here. I don't remember why. I'd apparently used a condition based on swizzles in my initial patch but neither of us could understand this. Don't sweat it, we'll figure it out as part of #625.

src/device_properties.cpp
kpet: I've created https://github.com/kpet/clvk/issues/633

kpet (changes requested):

src/api.cpp
```diff
@@ -4820,59 +4674,38 @@ cl_int CLVK_API_CALL clGetSupportedImageFormats(cl_context context,
...
+        image_format_support fmt_support;
+        cl_image_format clfmt = {
+            .image_channel_order = image_channel_order,
+            .image_channel_data_type = image_channel_data_type,
+        };
+        if (!cl_image_format_to_vulkan_format(clfmt, image_type, dev,
+                                              &fmt_support, &components_sampled,
+                                              &components_storage)) {
+            continue;
+        }
```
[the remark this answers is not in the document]
rjodinchr: If I remove this conversion, this PR makes no sense. The whole point of it is this conversion which allows to check for `is_bgra_format_not_supported_for_image1d_buffer`. That's why I did not want to iterate over the maps initially. But I agree having a getter is better. But I would prefer not to have access to the VkFormat there as people need to understand that the only mapping to consider is the one from the conversion function, not the map.

src/api.cpp
```diff
@@ -4820,59 +4674,38 @@ cl_int CLVK_API_CALL clGetSupportedImageFormats(cl_context context,
...
     // formats for which the Vulkan format is supported
-    for (auto const& clvkfmt : gFormatMaps) {
-        const cl_image_format& clfmt = clvkfmt.first;
-        auto const& fmt_support = clvkfmt.second;
+    for (auto mapping : get_format_maps()) {
+        auto image_channel_order = mapping.first.image_channel_order;
+        auto image_channel_data_type = mapping.first.image_channel_data_type;
+        VkComponentMapping components_sampled, components_storage;
+        image_format_support fmt_support;
+        cl_image_format clfmt = {
```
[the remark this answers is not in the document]
rjodinchr: I'll change that

src/api.cpp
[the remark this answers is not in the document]
rjodinchr: Instead of the getter, we could have a function returning the number of elements in the map, and a function to get the `cl_image_format` (only) for a specified element:
```
const unsigned get_format_maps_size();
const cl_image_format get_format_from_format_maps_at(unsigned i);
```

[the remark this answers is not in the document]
rjodinchr: Forget about that, FormatMaps is a map... not possible to do it like that.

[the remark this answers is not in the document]
kpet: Right, I'd lost track of the big picture :D. I'm sure we'll find good ways of refactoring this as more support for images makes its way into the code.

PR #630, opened 2023-10-30
Move the device timer conversion in cvk_device instead of commands

rjodinchr:


src/device.cpp
```diff
@@ -1196,11 +1196,15 @@ cl_int cvk_device::get_device_host_timer(cl_ulong* device_timestamp,
     return CL_SUCCESS;
 }
 
-cl_ulong cvk_device::device_timer_to_host(cl_ulong dev, cl_ulong sync_dev,
-                                          cl_ulong sync_host) const {
-    if (sync_host > sync_dev) {
-        return (sync_host - sync_dev) + dev;
+cl_ulong cvk_device::device_timer_to_host(cl_ulong dev) {
+    if (dev > m_sync_dev) {
```
kpet, line 1200:
```suggestion
    if (dev <= m_sync_dev) {
```
Isn't your condition backwards? You want a new pair of aligned timestamps only when you've wrapped around and the timestamp to convert to the host's time base is smaller than the device timestamp used as a reference.

src/device.cpp
```diff
@@ -1196,11 +1196,15 @@ cl_int cvk_device::get_device_host_timer(cl_ulong* device_timestamp,
...
+        if (get_device_host_timer(&m_sync_dev, &m_sync_host) != CL_SUCCESS) {
+            return dev;
```
kpet, line 1202: This means we just silently return (most likely) garbage data in case an error is encountered. We should propagate the error I think.

kpet (commented):
What's the motivation for this change? Performance improvements?

src/device.cpp
rjodinchr: I don't think it is backwards.
The idea is to get new sync points when the time we want is bigger than what we have converted so far. Because they might have been a slight divergence since we have the last sync point.

src/device.cpp
rjodinchr: Sure, I'll change that.

rjodinchr:
> What's the motivation for this change? Performance improvements?

From what I remember, this is needed to fix issues introduced by the new pattern from using timeline semaphore.
But as without timeline semaphore this is just a different implementation that works as well, I thought it would be better to submit it before in order to reduce the size of the [timeline semaphore PR](https://chromium.googlesource.com/chromiumos/overlays/chromiumos-overlay/+/refs/heads/main/media-libs/clvk/files/clvk-90-timeline-semaphores.patch).

src/device.cpp
kpet: So you'll end up requesting a new pair of timestamps for pretty much each command since the device timestamp will almost always be greater than the last one obtained when getting a host/device pair. The only time where you're not requesting a host/device pair is when the device timestamp wraps around. At which point I don't see much motivation for the change.

rjodinchr:
I have rerun the CTS without this patch (but with the timeline semaphore) to remind me why this patch is needed with timeline semaphore.

In the implementation I have, there is a race condition between the application thread and the executor thread. It can lead to the thread app looking for an event profiling info, because the event is complete, but the executor has not computed the profiling info.
But this event might be related to a cvk_command inside of a cvk_command_batch. As the sync points are taken care by the batch command, it was making things complicated. And this is the way I dealt with it, moving the sync point to the device instead of the batch command.

kpet:
As discussed, I really don't see the merit of this change on its own. In fact, it looks likely to be a measurable performance regression for some devices (to be revisited under #778) where getting synchronised timestamps is somewhat expensive. I understand that you've found it useful to enable further changes that we do want to get in so I'll take your word for it and merge it so we get forward progress.

PR #636, opened 2023-11-15
Update OpenCL dep

amaiorano:
This OpenCL commit renamed some macros:
https://github.com/KhronosGroup/OpenCL-Headers/commit/8de09e8f07e3272da06e62358a8116c27d235036

This breaks importing OpenCL into google internal, so fixing upstream.


kpet (approved):
Thanks!

PR #642, opened 2024-01-08
improve descriptor set reuse

### LINKED ISSUE #580: Improve descriptor set reuse
We currently allocate new descriptor sets for each kernel enqueue when kernel arguments change but we don't use features such as update after bind so we only really need one set of descriptor sets per concurrent instance of the kernel command building code. With the current approach to command queue locks, that should mean one per kernel per command queue.

rjodinchr:
Only allocate one descriptor set per entry-point per queue. And always reuse it as we are not doing update after bind.

Fix #580


rjodinchr:
I am running the full CTS against it

rjodinchr:
I have just realised that it is missing the per queue part that is needed.

rjodinchr:
This is definitively not the right way to go, abandonning that change

PR #643, opened 2024-01-09
add counter for allocated descriptor sets

rjodinchr:


kpet (commented):
Does it make sense to rebase that one to pick up the new Perfetto testing?

kpet (approved):

PR #644, opened 2024-01-09
allow to retry allocation of descriptor sets

rjodinchr:


src/queue.cpp
```diff
@@ -970,7 +972,25 @@ cvk_command_kernel::build_batchable_inner(cvk_command_buffer& command_buffer) {
...
+            m_queue->command_pool_unlock();
+            std::this_thread::sleep_for(std::chrono::microseconds(
+                config.descriptor_sets_allocate_retry_sleep_us));
+            m_queue->command_pool_lock();
+            result_valid = m_argument_values->setup_descriptor_sets();
+        } while (!result_valid && m_queue->have_batch_in_flight());
+        if (!result_valid) {
+            cvk_error_fn("could not setup descriptor sets");
+            return CL_OUT_OF_RESOURCES;
+        }
```
kpet, line 993: I did wonder whether we could encapsulate the retry within `setup_descriptor_sets` but this would mean keeping the command pool locked for the whole duration of the retry, preventing other threads from enqueuing kernels from entry points that still have descriptor sets available in their pool. I think the performance advantage is worth the lack of encapsulation. We could probably restructure things to avoid that trade-off but that'd be way beyond the scope of this PR. In short: fine with your approach :).

kpet (changes requested):
That's a great feature to have, thanks! I think we should add two tests as part of this PR:
- Check that enqueing more instances than the max number results in a failure with default config options (and check that retry is disabled by default in this test)
- Check that enqueing more instances than the max number with retry enabled succeeds.

I recommend you have a look at https://github.com/kpet/clvk/pull/656 that should make writing these tests easy (possibly requiring additional `CLVK_CONFIG_ASSERT` macros).

src/queue.cpp
rjodinchr: I have looked at it when making the PR. The issue is even worse, if the command pool is locked we cannot free descriptors set: `program.hpp: cvk_entry_point::free_descriptor_set` locks the command pool.

rjodinchr:
Not so easy to test.
With `llvmpipe`, the descriptor set allocation never fails (or you take down your machine). I have added a config option to force the failure.
I'm not sure how stable the tests will be. The good thing is that we can now use perfetto to trace whether we go in certain areas of the code (ie, you can see the new perfetto tracepoint `descriptor_sets_allocate_retry_sleep` in `api_tests.EnqueueTooManyCommandWithRetry-expectation.txt`)

rjodinchr:
Converting to draft as it is based on your PR

src/program.cpp
```diff
@@ -2216,8 +2224,13 @@ bool cvk_entry_point::allocate_descriptor_sets(VkDescriptorSet* ds) {
         vkAllocateDescriptorSets(m_device, &descriptorSetAllocateInfo, ds);
 
     if (res != VK_SUCCESS) {
-        cvk_error_fn("could not allocate descriptor sets: %s",
-                     vulkan_error_string(res));
+        if (config.descriptor_sets_allocate_retry_sleep_us == UINT32_MAX) {
+            cvk_error_fn("could not allocate descriptor sets: %s",
+                         vulkan_error_string(res));
+        } else {
+            cvk_info_fn("could not allocate descriptor sets: %s",
```
kpet, line 2231:
```suggestion
            cvk_warn_fn("could not allocate descriptor sets: %s",
```

Retries will have a performance impact so maybe we should flag them as warnings? Would that create too much noise in the applications you've used this with?

tests/api/enqueue.cpp
```diff
@@ -439,3 +439,147 @@ TEST_F(WithCommandQueue, FinishAfterFlush) {
         thread->join();
     }
 }
+
+#ifdef CLVK_UNIT_TESTING_ENABLED
+TEST_F(WithCommandQueue, EnqueueTooManyCommands) {
+
+    static const unsigned NUM_INSTANCES =
+        CLVK_CONFIG_GET(max_entry_points_instances);
```
kpet, line 447: I was initially thinking we'd override this config option to reduce the max number of instances: we don't need that many for this test. It doesn't take a lot of time though so fine as is.

tests/api/enqueue.cpp
```diff
@@ -439,3 +439,147 @@ TEST_F(WithCommandQueue, FinishAfterFlush) {
...
+    do {
+        err = clEnqueueNDRangeKernel(m_queue, kernel, 1, nullptr, &gws, &lws, 0,
+                                     nullptr, nullptr);
+        count++;
+    } while (err == CL_OUT_OF_RESOURCES);
+    cl_int event_cmd_execution_status;
+    GetEventInfo(event, CL_EVENT_COMMAND_EXECUTION_STATUS,
+                 &event_cmd_execution_status);
+    ASSERT_EQ(event_cmd_execution_status, CL_COMPLETE);
+    ASSERT_GT(count, 10);
```
kpet, line 500: Where does the 10 come from?

tests/api/enqueue.cpp
```diff
@@ -439,3 +439,147 @@ TEST_F(WithCommandQueue, FinishAfterFlush) {
...
+    cl_uint count = 0;
+    do {
+        err = clEnqueueNDRangeKernel(m_queue, kernel, 1, nullptr, &gws, &lws, 0,
+                                     nullptr, nullptr);
+        count++;
+    } while (err == CL_OUT_OF_RESOURCES);
+    cl_int event_cmd_execution_status;
+    GetEventInfo(event, CL_EVENT_COMMAND_EXECUTION_STATUS,
+                 &event_cmd_execution_status);
+    ASSERT_EQ(event_cmd_execution_status, CL_COMPLETE);
```
kpet, line 499: This fails on my machine. There's no synchronisation to guarantee that execution will have completed at that point.

tests/api/enqueue.cpp
```diff
@@ -439,3 +439,147 @@ TEST_F(WithCommandQueue, FinishAfterFlush) {
...
+    Flush();
+
+    SetKernelArg(kernel, 1, &i);
+    cl_uint err;
+    cl_uint count = 0;
+    do {
+        err = clEnqueueNDRangeKernel(m_queue, kernel, 1, nullptr, &gws, &lws, 0,
+                                     nullptr, nullptr);
+        count++;
+    } while (err == CL_OUT_OF_RESOURCES);
```
kpet, line 495: I'm not sure I understand the use of a loop or why the test is flushing. Can't we just enqueue `NUM_INSTANCES` kernel in a loop, asserting success and check that we get a failure if we enqueue one more?

tests/api/enqueue.cpp
```diff
@@ -439,3 +439,147 @@ TEST_F(WithCommandQueue, FinishAfterFlush) {
...
+    while (sampleTime() < ts_start + one_second && err == CL_OUT_OF_RESOURCES) {
+        err = clEnqueueNDRangeKernel(m_queue, kernel, 1, nullptr, &gws, &lws, 0,
+                                     nullptr, nullptr);
+    }
+    ASSERT_CL_SUCCESS(err);
+
+    for (; i < NUM_INSTANCES_RETRY; i++) {
+        SetKernelArg(kernel, 1, &i);
+        EnqueueNDRangeKernel(kernel, 1, nullptr, &gws, &lws);
+    }
```
kpet, line 564: Similar question here: I'm not sure I understand why you're doing the enqueues in three groups or why you're swallowing some enqueue errors. This doesn't seem to be the behaviour that we want/ Are we missing an automatic flush in the runtime when we run out of descriptor sets?

src/queue.cpp
kpet: Right, this is becoming a bit too tricky to reason about. We don't need to solve this here but I've created https://github.com/kpet/clvk/issues/658 so we don't forget about this issue.

.github/workflows/presubmit.yml
```diff
@@ -259,6 +259,8 @@ jobs:
         run: |
           ${{ env.testbindir }}/simple_test
           ./tests/perfetto/check-perfetto-trace.sh ${CLVK_PERFETTO_TRACE_DEST} ./tests/perfetto/simple_test-expectation.txt
+          ${{ env.testbindir}}/api_tests --gtest_filter=WithCommandQueue.EnqueueTooManyCommandsWithRetry
+          ./tests/perfetto/check-perfetto-trace.sh ${CLVK_PERFETTO_TRACE_DEST} ./tests/perfetto/api_tests.EnqueueTooManyCommandWithRetry-expectation.txt
```
kpet, line 263: It's nice to be able to check that we're hitting specific functions but I am concerned about how scalable/maintainable these tests will be if we start adding a few of them. It's not a problem right now but we need to keep an eye on this.

src/config.def
```diff
@@ -74,3 +75,10 @@ OPTION(bool, queue_profiling_use_timestamp_queries, false)
 OPTION(uint32_t, perfetto_trace_max_size, 1024u)
 OPTION(std::string, perfetto_trace_dest, "clvk.perfetto-trace")
 #endif // CLVK_PERFETTO_BACKEND_INPROCESS
+
+//
+// Test
+//
+#if CLVK_UNIT_TESTING_ENABLED
+OPTION(bool, force_descriptor_set_allocation_failure, false)
+#endif
```
kpet, line 84: Good idea! I hadn't thought about that but a nice side effect of my PR is that we can use config options to drive test behaviours in many scenarios that would have required dedicated test functions.

kpet (changes requested):
A few more tweaks needed.

src/program.cpp
rjodinchr: Yes, it would produce too much noise. The performance impact is controlled by the performance setting the retry time. Which is disabled by default, so to me it is fine keeping it as info.
Maybe we could just warn once on first retry?

tests/api/enqueue.cpp
rjodinchr: I'm not sure what I was thinking here, I'll simplify this test.

tests/api/enqueue.cpp
rjodinchr: I'm not sure what I was thinking here, I'll simplify this test.

tests/api/enqueue.cpp
rjodinchr: I'm not sure what I was thinking here, I'll simplify this test.

tests/api/enqueue.cpp
rjodinchr: This is complicated. You could consider that we are missing an automatic flush, but this is not an easy one to insert. Basically we are failing the descriptor set allocation when we are here:
```
  * frame #0: 0x00007ffff770637e libOpenCL.so.1`cvk_command_kernel::build_batchable_inner(this=0x0000555556086a00, command_buffer=0x0000555556086bb0) at queue.cpp:985:9
    frame #1: 0x00007ffff7706ec4 libOpenCL.so.1`cvk_command_batchable::build(this=0x0000555556086a00, command_buffer=0x0000555556086bb0) at queue.cpp:1169:16
    frame #2: 0x00007ffff770bbe4 libOpenCL.so.1`cvk_command_batch::add_command(this=0x0000555555bb1460, cmd=0x0000555556086a00) at queue.hpp:830:27
    frame #3: 0x00007ffff7700a90 libOpenCL.so.1`cvk_command_queue::enqueue_command(this=0x0000555555b98960, cmd=0x0000555556086a00, event=0x0000000000000000) at queue.cpp:150:32
    frame #4: 0x00007ffff7701854 libOpenCL.so.1`cvk_command_queue::enqueue_command_with_deps(this=0x0000555555b98960, cmd=0x0000555556086a00, num_dep_events=0, dep_events=0x0000000000000000, event=0x0000000000000000) at queue.cpp:209:12
    frame #5: 0x00007ffff76236dd libOpenCL.so.1`cvk_enqueue_ndrange_kernel(command_queue=0x0000555555b98960, kernel=0x0000555555ba9d20, dims=1, ndrange=0x00007fffffffd610, num_events_in_wait_list=0, event_wait_list=0x0000000000000000, event=0x0000000000000000) at api.cpp:4016:27
    frame #6: 0x00007ffff7623f1a libOpenCL.so.1`clEnqueueNDRangeKernel(command_queue=0x0000555555b98978, kernel=0x0000555555ba9d38, work_dim=1, global_work_offset=0x0000000000000000, global_work_size=0x00007fffffffd830, local_work_size=0x00007fffffffd828, num_events_in_wait_list=0, event_wait_list=0x0000000000000000, event=0x0000000000000000) at api.cpp:4075:12
 ```
 This makes it quite hard to flush at that point.
 Maybe we should not wait in build_batchable_inner, and do that in enqueue_command to be able to flush.
 I'll have a look to see if that is possible.

**Comment by rjodinchr** [comment]
Looking at the previous errors with the validation layers, it felt that the deletion of the command was needed, but it clearly broke macos and windows.

File: src/queue.hpp [hunk 10]
```diff
@@ -821,12 +823,16 @@ struct cvk_command_batch : public cvk_command {
             m_queue->command_pool_lock();
         }
 
+        cl_int ret = cmd->build(*m_command_buffer);
+        if (ret != CL_SUCCESS) {
+            m_queue->command_pool_unlock();
```
kpet, line 828: It looks like we could use a command pool lock holder type thing for functions like this one. Not sure how that'd have to interact with `cvk_command_buffer::begin` though.

src/program.cpp
```diff
@@ -2243,8 +2252,19 @@ bool cvk_entry_point::allocate_descriptor_sets(VkDescriptorSet* ds) {
...
+        } else if (m_first_allocation_failure) {
+            cvk_warn_fn(
+                "could not allocate descriptor sets: %s, retry in %u us",
+                vulkan_error_string(res),
+                config.enqueue_command_retry_sleep_us());
+        } else {
+            cvk_info_fn("could not allocate descriptor sets: %s",
+                        vulkan_error_string(res));
+        }
+        m_first_allocation_failure = false;
```
kpet, line 2267: It's quite a bit of complexity just to improve log messages :). Fine with it but I note the following:
- Maybe we should start thinking of a `cvk_warn_once` function that'd encapsulate this.
- The message about retry could get out of sync since the decision is made much higher up in the call stack. Ideally the message would be printed by the bit of code that handles the retry.

kpet (approved):
Thanks for all the changes!

PR #645, opened 2024-01-09
name executor thread

rjodinchr:


rjodinchr:
Not sure how it will behave on windows

kpet (commented):
I think we should push the details to `utils.[ch]pp` as done for other platform-specific functions. Maybe a `cvk_set_thread_name_if_supported(const std::native_handle_type&, const std::string&)`?

kpet (approved):
LGTM, thanks!

PR #646, opened 2024-01-09
add list of unsupported image_format in device_properties

rjodinchr:
Some Intel drivers have bug with CL_UNORM_SHORT_565. As it is not mandatory for OpenCL, just report it as unsupported for Intel.

### DIFF
--- a/src/device.hpp
+++ b/src/device.hpp
@@ -560,6 +560,11 @@ struct cvk_device : public _cl_device_id,
             ->is_bgra_format_not_supported_for_image1d_buffer();
     }
 
+    bool is_image_format_disabled(cl_image_format format) const {
+        return m_clvk_properties->get_unsupported_image_format().count(
+                   format) != 0;
+    }
+
 private:
     std::string version_desc() const {
         std::string ret = "CLVK on Vulkan v";
--- a/src/device_properties.cpp
+++ b/src/device_properties.cpp
@@ -111,6 +111,10 @@ struct cvk_device_properties_intel : public cvk_device_properties {
     is_bgra_format_not_supported_for_image1d_buffer() const override final {
         return true;
     }
+
+    image_format_set get_unsupported_image_format() const override final {
+        return image_format_set({{CL_RGB, CL_UNORM_SHORT_565}});
+    }
 };
 
 static bool isIntelDevice(const char* name, const uint32_t vendorID) {
--- a/src/device_properties.hpp
+++ b/src/device_properties.hpp
@@ -56,6 +56,17 @@ struct cvk_device_properties {
         return false;
     }
 
+    struct ClFormatSetCompare {
+        int operator()(const cl_image_format& lhs,
+                       const cl_image_format& rhs) const {
+            return lhs.image_channel_order > rhs.image_channel_order ||
+                   (lhs.image_channel_order == rhs.image_channel_order &&
+                    lhs.image_channel_data_type > rhs.image_channel_data_type);
+        }
+    };
+    using image_format_set = std::set<cl_image_format, ClFormatSetCompare>;
+    virtual image_format_set get_unsupported_image_format() const { return {}; }
+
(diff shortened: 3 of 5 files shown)


src/device.hpp
```diff
@@ -553,6 +553,11 @@ struct cvk_device : public _cl_device_id,
             ->is_bgra_format_not_supported_for_image1d_buffer();
     }
 
+    bool supports_image_format(cl_image_format format) const {
```
kpet, line 556: I think the name is confusing. This function does not return `true` if the format is supported and usable. It returns `true` if the format is not in a deny list. Maybe `is_image_format_disabled` (note that the caller is testing the negative) or better name along those lines?

Also I think we should fold the case currently handled by `is_bgra_format_not_supported_for_image1d_buffer` into this new property query. Fine if you want to do this as a separate PR (we can create an issue for it then).

src/device_properties.cpp
```diff
@@ -124,7 +128,7 @@ struct cvk_device_properties_amd : public cvk_device_properties {
     }
     cl_uint get_max_first_cmd_batch_size() const override final { return 10; }
     cl_uint get_max_cmd_group_size() const override final { return 1; }
-    const std::set<std::string> get_native_builtins() const override final {
+    std::set<std::string> get_native_builtins() const override final {
```
kpet, line 131: Technically a separate change but we could make the sets static and return a const reference to them instead of performance copies. I'm not too concerned about the performance impact given where it's called though.

src/device_properties.hpp
```diff
@@ -56,6 +56,17 @@ struct cvk_device_properties {
...
 
+    struct ClFormatSetCompare {
+        int operator()(const cl_image_format& lhs,
+                       const cl_image_format& rhs) const {
+            return lhs.image_channel_order > rhs.image_channel_order ||
+                   (lhs.image_channel_order == rhs.image_channel_order &&
+                    lhs.image_channel_data_type > rhs.image_channel_data_type);
+        }
+    };
+    using image_format_set = std::set<cl_image_format, ClFormatSetCompare>;
```
kpet, line 67: Does the set need to be ordered?
- If not, can you reuse `ClFormatMapEqual`? (Maybe after a rename to remove `Map`)
- If yes, I suggest moving the functor to `image_format.hpp` with a rename to remove `Set`.

kpet (changes requested):

src/device_properties.cpp
rjodinchr: I'll remove that and make another PR for it later.

src/device_properties.hpp
rjodinchr: The set does not need to be ordered, but reusing `clFromatMapEqual` leads to crashes on ChromeOS, so I'd prefer to stick with what is in this PR.
I can move the type to `image_format.hpp` but it feels weird to me as it creates a dependency of `device_properties.hpp` on `image_format.hpp` that is not really needed.

src/device.hpp
rjodinchr: I'll rename this function.
But I don't think `is_bgra_format_not_supported_for_image1d_buffer` should be dealt with inside it.  Those are not at the same level. One is about the CL format, the other is about the Vulkan format.
Everything is taking care of in `cl_image_format_to_vulkan_format` which is the link between the two.

kpet: Yeah, looking again at how it's used we're doing a bit more than I remembered and remapping the format to another one so we would need a bit more work to merge everything into a single mechanism. Better to wait until we see more cases to attempt to do something.

src/device_properties.hpp
kpet: Hmm, this sounds suspicious :). The dependency might not be too bad. I'll take a stab at this.

kpet (approved):
Let's get this in since it works and is required for CTS.

PR #647, opened 2024-01-09
only report read-write images if we have theright spirv capability

rjodinchr:
We need at least StorageImageReadWithoutFormat to support read write images.

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -730,6 +730,7 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
             val_uint = 0;
             copy_ptr = &val_uint;
             size_ret = sizeof(val_uint);
+            break;
         }
         [[fallthrough]];
     case CL_DEVICE_MAX_WRITE_IMAGE_ARGS:
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -731,7 +731,7 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
             copy_ptr = &val_uint;
             size_ret = sizeof(val_uint);
         }
-        // fallthrough
+        [[fallthrough]];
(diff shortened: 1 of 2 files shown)


src/api.cpp
```diff
@@ -726,6 +726,12 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
         size_ret = sizeof(val_uint);
         break;
     case CL_DEVICE_MAX_READ_WRITE_IMAGE_ARGS:
+        if (!device->supports_read_write_images()) {
+            val_uint = 0;
+            copy_ptr = &val_uint;
+            size_ret = sizeof(val_uint);
```
kpet, line 732:
```suggestion
            size_ret = sizeof(val_uint);
            break;
```

Is this missing a `break`? Unless I'm missing something, `val_uint` will always be overwritten below.

src/api.cpp
```diff
@@ -726,6 +726,12 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
...
+        }
+        // fallthrough
```
kpet, line 734:
```suggestion
        [[fallthrough]];
```

We're now building as C++17 so might as well :)

src/device.hpp
```diff
@@ -298,6 +298,11 @@ struct cvk_device : public _cl_device_id,
         return devices_support_images() ? CL_TRUE : CL_FALSE;
     }
 
+    bool supports_read_write_images() const {
+        return supports_capability(
+            spv::CapabilityStorageImageReadWithoutFormat);
```
kpet, line 303: I think we should also require `CapabilityStorageImageWriteWithoutFormat`.

kpet (changes requested):

src/device.hpp
rjodinchr: I agree, I have added it

kpet (approved):

PR #648, opened 2024-01-09
fix return code of api function when spirv-il is not supported

rjodinchr:


kpet (commented):
It seems there are other entry points for which we should return this according to Appendix H but we currently don't (e.g. `clCreateProgramWithIL`). Are they not covered by the CTS?

Also, while I agree this approach is good enough in practice for now (single device per context and IL support is not a per device property) the API layer ought to iterate over devices to report the error. It's fine by me if you don't want to do this now but I think we should create an issue that I could add to a big to-do list for multiple device support.

rjodinchr:
The other functions are going through `parse_user_spec_constants` at some point which takes care of returning the correct error code if clvk has been compiled with `ENABLE_SPIRV_IL=0`.

kpet (approved):
Right and there's going to be work to adapt that code too for multiple devices. Let's not try to boil the ocean :).

PR #649, opened 2024-01-13
name build thread

rjodinchr:


rjodinchr:
I don't understand the issue. Can you rerun the tests please?

rjodinchr:
I think github is having some issues.

kpet:
I've retriggered the failing job a few times and it's always failing. I can't make sense of it but this may be a genuine issue on macOS. If that's the case, I don't understand why we didn't see it on the PR that named executor threads. Maybe you could try disabling thread naming on macOS in the utility function?

rjodinchr:
Thread naming is already disabled on macOS

rjodinchr:
The other PR is having issues with macOS. I think this is an issue with github and macOS.

PR #650, opened 2024-01-13
add test using perfetto in CI

### LINKED ISSUE #442: [CI] add test using perfetto

rjodinchr:
fix #442


rjodinchr:
I don't understand the error here. Feels like an issue with github trying to execute the `env` part of the `presubmit.yml`.
Can you reschedule it please?

.github/workflows/presubmit.yml
```diff
@@ -112,6 +113,17 @@ jobs:
         if: ${{ matrix.android-abi != '' }}
         run: |
           echo "android-cmake-args=-DCMAKE_TOOLCHAIN_FILE=${{ github.workspace }}/ndk/android-ndk-r25c/build/cmake/android.toolchain.cmake -DANDROID_ABI=${{ matrix.android-abi }} -DVulkan_LIBRARY=${{ github.workspace }}/ndk/android-ndk-r25c/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/lib/aarch64-linux-android/33/libvulkan.so" >> $GITHUB_ENV
+      - name: Build Perfetto
+        if: ${{ matrix.perfetto }}
+        shell: bash
+        env:
+          PERFETTO_DIR: '${{ github.workspace }}/external/perfetto'
```
kpet, line 120:
```suggestion
          PERFETTO_DIR: ${{ github.workspace }}/external/perfetto
```

This might fix the issues you're seeing.

rjodinchr:
Only the macOS issue left!

At least the perfetto part looks ready for review

kpet (changes requested):
Generally looks good, thanks! I think the expected output should be provided as a parameter to the checking script and the expectation file named after the corresponding test so it's easy to understand the relationship between the two and add further tests.

kpet (approved):
LGTM, thanks!

PR #651, opened 2024-01-18
rework thread naming to avoid requiring native_handle_type

rjodinchr:
We are having issue in google infra with native_handle_type which is
not supported.

### DIFF
--- a/src/utils.hpp
+++ b/src/utils.hpp
@@ -29,8 +29,7 @@
 
 char* cvk_mkdtemp(std::string& tmpl);
 int cvk_exec(const std::string& cmd, std::string* output = nullptr);
-void cvk_set_thread_name_if_supported(const std::thread::native_handle_type&,
-                                      const std::string&);
+void cvk_set_current_thread_name_if_supported(const std::string&);
 
 #define CVK_VK_CHECK_INTERNAL(logfn, res, msg)                                 \
     do {                                                                       \
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -1515,7 +1515,7 @@ bool cvk_program::check_capabilities(const cvk_device* device) const {
 }
 
 void cvk_program::do_build() {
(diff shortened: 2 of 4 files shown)


src/utils.hpp
```diff
@@ -29,8 +29,12 @@
 
 char* cvk_mkdtemp(std::string& tmpl);
 int cvk_exec(const std::string& cmd, std::string* output = nullptr);
+#if ENABLE_THREAD_NAMING
 void cvk_set_thread_name_if_supported(const std::thread::native_handle_type&,
                                       const std::string&);
+#else
+#define cvk_set_thread_name_if_supported(...)
```
ben-clayton, line 36: Defines in headers is not very friendly, nor are lowercase defines.
Compilers might also warn about semicolons after the "call" that come without a statement.
Recommend using an empty inline function instead.

rjodinchr: The empty inline function will keep the `m_thread->native_handle` when the function is called in the code, which is really what's breaking google infra.

rjodinchr:
Another way to fix google infra issues would be to use `pthread_self` instead of `m_thread->native_handle()` (and then change `cvk_set_thread_name_if_supported` prototype).

@kpet, would that work for you?

src/utils.hpp
rjodinchr: Possible. I personally don't like the idea because it is not developer friendly.  But if @kpet likes it, I'll do that.

kpet (changes requested):
Right, `std::thread::native_handle_type` is not guaranteed to be present. I don't think we should add a user-configurable option just for this. If we want to keep the functionality in your environment, we could change the naming function so it only applies to the current thread: `cvk_set_current_thread_name_if_supported` maybe? You could then use `pthread_self`. We could detect whether the required functions are supported using CMake potentially (bonus points for this but not IMO required).

kpet (approved):

rjodinchr:
Thanks 😀

PR #654, opened 2024-01-22
fix subgroup support & add partial support of cl_intel_subgroups

rjodinchr:
This PR fix an issue on Intel device where the subgroup size used by the compiler is not the default one exposed by the vulkan driver through `m_subgroup_properties.subgroupSize`. It leads to error when running CTS `test_api sub_group_dispatch` as the expected number of subgroup does not match the one seen during execution.

To fix that, this PR uses
`VkPipelineShaderStageRequiredSubgroupSizeCreateInfo` to force the subgroup size used by the compiler.

Doing so, it breaks TFLite because it assumes that the subgroup size is always 16. Which used to work by luck because it was the size used by the compiler for the kernel using subgroups on application I have tried.
But now the default size exposed by vulkan is 32, thus kernels are compiled with subgroup size of 32.

To avoid breaking TFLite, this PR implements a subpart of `cl_intel_subgroups` which is enough for TFLite to use this extension and specify the subgroup size inside the kernel using `__attribute__((intel_reqd_sub_group_size))`.

This PR also adds 2 config variable. One to set a prefered subgroup size to use if nothing is specified. Another to set a force subgroup size overriding everything. This aims at allowing easy debug if/when we will be facing other issues with subgroup size.

### DIFF
--- a/README.md
+++ b/README.md
@@ -436,7 +436,7 @@ variables. Here's a quick guide:
   conformance on devices that do not support linear filtering with all the image formats
   required for conformance.
 
-* `CLVK_PREFERED_SUBGROUP_SIZE` specifies the subgroup size to use if nothing
+* `CLVK_PREFERRED_SUBGROUP_SIZE` specifies the subgroup size to use if nothing
   is specified in the kernel. When not set use the default value reported by
   the Vulkan driver.
 
--- a/README.md
+++ b/README.md
@@ -436,6 +436,13 @@ variables. Here's a quick guide:
   conformance on devices that do not support linear filtering with all the image formats
   required for conformance.
 
+* `CLVK_PREFERED_SUBGROUP_SIZE` specifies the subgroup size to use if nothing
+  is specified in the kernel. When not set use the default value reported by
+  the Vulkan driver.
+
+* `CLVK_FORCE_SUBGROUP_SIZE` specifies the subgroup size to use, overriding
+  everything.
+
 # Limitations
 
 * Only one device per CL context
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -430,7 +430,7 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
     cl_device_device_enqueue_capabilities val_dev_enqueue_caps;
     cl_device_pci_bus_info_khr val_pci_bus_info;
     cl_device_atomic_capabilities val_atomic_capabilities;
-    std::vector<size_t> subgroup_sizes;
+    std::vector<size_t> val_subgroup_sizes;
 
     auto device = icd_downcast(dev);
 
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -906,7 +906,7 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
         size_ret = sizeof(val_pci_bus_info);
         break;
     case CL_DEVICE_SUB_GROUP_SIZES_INTEL:
-        if (device->supports_controled_subgroups()) {
+        if (device->supports_subgroup_size_selection()) {
             uint32_t size = device->min_sub_group_size();
             while (size <= device->max_sub_group_size()) {
                 subgroup_sizes.push_back((size_t)size);
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -909,11 +909,11 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
         if (device->supports_subgroup_size_selection()) {
             uint32_t size = device->min_sub_group_size();
             while (size <= device->max_sub_group_size()) {
-                subgroup_sizes.push_back((size_t)size);
+                val_subgroup_sizes.push_back((size_t)size);
                 size *= 2;
             }
-            copy_ptr = subgroup_sizes.data();
-            size_ret = sizeof(size_t) * subgroup_sizes.size();
+            copy_ptr = val_subgroup_sizes.data();
+            size_ret = sizeof(size_t) * val_subgroup_sizes.size();
             break;
         }
         [[fallthrough]];
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -430,6 +430,7 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
     cl_device_device_enqueue_capabilities val_dev_enqueue_caps;
     cl_device_pci_bus_info_khr val_pci_bus_info;
     cl_device_atomic_capabilities val_atomic_capabilities;
+    std::vector<size_t> subgroup_sizes;
 
     auto device = icd_downcast(dev);
 
@@ -904,6 +905,18 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
         copy_ptr = &val_pci_bus_info;
         size_ret = sizeof(val_pci_bus_info);
         break;
+    case CL_DEVICE_SUB_GROUP_SIZES_INTEL:
+        if (device->supports_controled_subgroups()) {
+            uint32_t size = device->min_sub_group_size();
+            while (size <= device->max_sub_group_size()) {
+                subgroup_sizes.push_back((size_t)size);
+                size *= 2;
+            }
+            copy_ptr = subgroup_sizes.data();
+            size_ret = sizeof(size_t) * subgroup_sizes.size();
+            break;
+        }
+        [[fallthrough]];
     default:
         ret = CL_INVALID_VALUE;
         break;
--- a/src/config.def
+++ b/src/config.def
@@ -39,6 +39,9 @@ OPTION(std::string, llvmspirv_bin, DEFAULT_LLVMSPIRV_BINARY_PATH)
 #endif // !CLSPV_ONLINE_COMPILER
 #endif // COMPILER_AVAILABLE
 
+OPTION(uint32_t, force_subgroup_size, 0u) // 0 meaning not forced
+OPTION(uint32_t, prefered_subgroup_size, 0u) // 0 meaning no preference
+
 //
 // Command execution
 //
--- a/src/config.def
+++ b/src/config.def
@@ -40,7 +40,7 @@ OPTION(std::string, llvmspirv_bin, DEFAULT_LLVMSPIRV_BINARY_PATH)
 #endif // COMPILER_AVAILABLE
 
 OPTION(uint32_t, force_subgroup_size, 0u) // 0 meaning not forced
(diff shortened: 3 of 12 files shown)


README.md
```diff
@@ -436,6 +436,13 @@ variables. Here's a quick guide:
   conformance on devices that do not support linear filtering with all the image formats
   required for conformance.
 
+* `CLVK_PREFERED_SUBGROUP_SIZE` specifies the subgroup size to use if nothing
```
kpet, line 439:
```suggestion
* `CLVK_PREFERRED_SUBGROUP_SIZE` specifies the subgroup size to use if nothing
```

src/api.cpp
```diff
@@ -430,6 +430,7 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
     cl_device_device_enqueue_capabilities val_dev_enqueue_caps;
     cl_device_pci_bus_info_khr val_pci_bus_info;
     cl_device_atomic_capabilities val_atomic_capabilities;
+    std::vector<size_t> subgroup_sizes;
```
kpet, line 433:
```suggestion
    std::vector<size_t> val_subgroup_sizes;
```

For consistency with other definitions.

src/api.cpp
```diff
@@ -904,6 +905,18 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
         copy_ptr = &val_pci_bus_info;
         size_ret = sizeof(val_pci_bus_info);
         break;
+    case CL_DEVICE_SUB_GROUP_SIZES_INTEL:
+        if (device->supports_controled_subgroups()) {
```
kpet, line 909:
```suggestion
        if (device->supports_subgroup_size_selection()) {
```
Or maybe `supports_subgroup_size_control` so it's a bit more explicit than "controled_subgroups".

src/api.cpp
```diff
@@ -904,6 +905,18 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
...
+            uint32_t size = device->min_sub_group_size();
+            while (size <= device->max_sub_group_size()) {
+                subgroup_sizes.push_back((size_t)size);
+                size *= 2;
+            }
```
kpet, line 914: Are all sizes in between guaranteed to be supported?

src/config.def
```diff
@@ -39,6 +39,9 @@ OPTION(std::string, llvmspirv_bin, DEFAULT_LLVMSPIRV_BINARY_PATH)
 #endif // !CLSPV_ONLINE_COMPILER
 #endif // COMPILER_AVAILABLE
 
+OPTION(uint32_t, force_subgroup_size, 0u) // 0 meaning not forced
+OPTION(uint32_t, prefered_subgroup_size, 0u) // 0 meaning no preference
```
kpet, line 43:
```suggestion
OPTION(uint32_t, preferred_subgroup_size, 0u) // 0 meaning no preference
```

src/device.hpp
```diff
@@ -283,8 +283,27 @@ struct cvk_device : public _cl_device_id,
     }
 
     cl_uint sub_group_size() const {
+        if (supports_controled_subgroups()) {
+            cl_uint force_subgroup_size = config.force_subgroup_size();
+            if (config.force_subgroup_size.set &&
+                force_subgroup_size >= min_sub_group_size() &&
+                force_subgroup_size <= max_sub_group_size()) {
```
kpet, line 290: We should emit a message when a user provides an invalid subgroup size.

src/device.hpp
```diff
@@ -283,8 +283,27 @@ struct cvk_device : public _cl_device_id,
     }
 
     cl_uint sub_group_size() const {
+        if (supports_controled_subgroups()) {
+            cl_uint force_subgroup_size = config.force_subgroup_size();
```
kpet, line 287: Since we're repeating the use of `config.` I'm not sure this variable adds much.

src/device.hpp
```diff
@@ -313,6 +332,15 @@ struct cvk_device : public _cl_device_id,
 
     bool supports_subgroups() const { return m_has_subgroups_support; }
 
+    bool supports_controled_subgroups() const {
+        return (m_properties.apiVersion >= VK_MAKE_VERSION(1, 3, 0) ||
+                is_vulkan_extension_enabled(
+                    VK_EXT_SUBGROUP_SIZE_CONTROL_EXTENSION_NAME)) &&
```
kpet, line 338: Maybe you could set a boolean like the other `m_has_*` to avoid repeating the conditions. Maybe we should make this automatic at some point.

src/device.hpp
```diff
@@ -283,8 +283,27 @@ struct cvk_device : public _cl_device_id,
...
+        if (supports_controled_subgroups()) {
+            cl_uint force_subgroup_size = config.force_subgroup_size();
+            if (config.force_subgroup_size.set &&
+                force_subgroup_size >= min_sub_group_size() &&
+                force_subgroup_size <= max_sub_group_size()) {
+                return force_subgroup_size;
+            }
+            if (m_prefered_subgroup_size != 0 &&
+                m_prefered_subgroup_size >= min_sub_group_size() &&
+                m_prefered_subgroup_size <= max_sub_group_size()) {
```
kpet, line 295: Same here, we should emit a message when the size is not valid.

src/program.cpp
```diff
@@ -2159,13 +2162,29 @@ cvk_entry_point::create_pipeline(const cvk_spec_constant_map& spec_constants) {
         specConstantData.data(),
     };
 
+    void* pipelineShaderStageCreateInfoPNext = nullptr;
+    VkPipelineShaderStageRequiredSubgroupSizeCreateInfo
+        reqdSubgroupSizeCreateInfo;
+    if (m_device->supports_controled_subgroups()) {
+        auto reqdSubgroupSize = m_program->required_sub_group_size(m_name);
+        if (reqdSubgroupSize > m_device->max_sub_group_size() ||
+            reqdSubgroupSize < m_device->min_sub_group_size()) {
```
kpet, line 2171: Shouldn't we fail in that case? We should not magically change the subgroup size kernel writers assumed would be used.

src/program.hpp
```diff
@@ -418,23 +419,27 @@ struct cvk_program;
...
-    cvk_entry_point(VkDevice dev, cvk_program* program,
+    cvk_entry_point(cvk_device* dev, cvk_program* program,
                     const std::string& name);
 
     ~cvk_entry_point() {
         for (auto pipeline : m_pipelines) {
             cvk_info("destroying pipeline %p for kernel %s", pipeline.second,
                      m_name.c_str());
-            vkDestroyPipeline(m_device, pipeline.second, nullptr);
+            vkDestroyPipeline(m_device->vulkan_device(), pipeline.second,
```
kpet, line 429: You could define a `vkdev` variable to shorthen those lines. It's a common idiom throughout the code base.

src/program.hpp
```diff
@@ -759,6 +765,43 @@ struct cvk_program : public _cl_program, api_object<object_magic::program> {
         return m_binary.required_work_group_size(kernel);
     }
 
+    uint32_t required_sub_group_size(std::string& kernel) const {
```
kpet, line 768: Surely we should find a better way of doing this :). Is the information encoded in the SPIR-V module at all? If not we should teach clspv about the attribute and likely emit a **SubgroupSize** Execution Mode that we could either parse at binary loading time.

kpet (changes requested):
Mostly cosmetic comments apart from how the required size attribute is handled. I think we should also add a few tests to cover at least:
- The new config options
- The device query (CTS won't give us any coverage here)
-  The use of the required size attribute

src/api.cpp
rjodinchr: That's what I understand from the specification, but it is not stated as such.

src/device.hpp
rjodinchr: I am not sure to understand what condition we are repeating.

src/program.hpp
rjodinchr: I am not a fan of that idea. It means having to go through SPIRV-Registry SPIRV-Headers and (maybe?) SPIRV-Tools. All that to let clspv do the parsing instead of clvk? I have trouble understanding what's better that way.

rjodinchr:
I still need to add some tests

src/device.hpp
kpet: The check for Vulkan 1.3 or the extension.

tests/api/subgroup_size.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+static const char* program_source = R"(
+#pragma OPENCL EXTENSION cl_khr_subgroups : enable
+%s__kernel void test(__global int *output)
+{
+    output[0] = get_num_sub_groups();
+}
+)";
+
+#ifdef CLVK_UNIT_TESTING_ENABLED
```
kpet, line 25: Might as well include the program source to avoid a warning in non-unit builds.

tests/api/subgroup_size.cpp
```diff
@@ -0,0 +1,74 @@
...
+                  &max_work_group_size, nullptr);
+
+    auto buffer = CreateBuffer(0, sizeof(cl_int));
+
+    auto run = [this](const char* kernel_prefix, size_t subgroup_size,
+                      size_t max_work_group_size, cl_mem buffer) {
+        cl_int expected_num_sub_groups = max_work_group_size / subgroup_size;
+        size_t work_size = expected_num_sub_groups * subgroup_size;
+        char source[512];
+        sprintf(source, program_source, kernel_prefix);
```
kpet, line 45:
```suggestion
        sprintf(source, sizeof(source), program_source, kernel_prefix);
```

tests/api/subgroup_size.cpp
```diff
@@ -0,0 +1,74 @@
...
+               kernel_prefix, subgroup_size, max_work_group_size, *result,
+               expected_num_sub_groups);
+#endif
+        ASSERT_EQ(expected_num_sub_groups, *result);
+        EnqueueUnmapMemObject(buffer, result);
+    };
+
+    for (auto subgroup_size : subgroup_sizes) {
+        char attribute[64];
+        sprintf(attribute, "__attribute__((intel_reqd_sub_group_size(%lu))) ",
```
kpet, line 64:
```suggestion
        snprintf(attribute, sizeof(attribute), "__attribute__((intel_reqd_sub_group_size(%lu))) ",
```

kpet (changes requested):
Nice test! A couple of finishing touches and this will be good to go. It's a shame we can't use `std::format` for that test.

kpet (approved):
Thanks!

PR #660, opened 2024-01-28
fix typo in readme

rjodinchr:


kpet (approved):
Can't wait to get a better system to document options, one where the documentation is generated from the option definitions.

PR #661, opened 2024-01-28
do not fail if filter linear is not supported. use nearest instead

rjodinchr:


src/memory.cpp
```diff
@@ -184,8 +184,9 @@ bool cvk_sampler::init(bool force_normalized_coordinates) {
     }
 
     if (!config.supports_filter_linear && m_filter_mode == CL_FILTER_LINEAR) {
-        cvk_error_fn("linear filter is not supported");
-        return false;
+        m_filter_mode = CL_FILTER_NEAREST;
+        cvk_info_fn(
```
kpet, line 188:
```suggestion
        cvk_warn_fn(
```

kpet (approved):
As discussed on Discord, I had a suspicion we would need to do that :). I've turned the info into a warning. Do shout if you think that's a problem.

PR #662, opened 2024-01-28
Tidy up management of per-device disabled image formats

kpet:
- Use an unordered_set and existing functors from image_format.hpp
- Rename the device properties function to match the users
- Use function scope static objects to store the set of disabled formats

Change-Id: I9e3b5a66bae9abd29ce8e41e74ea9e2ad7c8df0b


rjodinchr (approved):
looking good on ChromeOS. Thanks!

PR #663, opened 2024-01-29
set CL_DEVICE_LATEST_CONFORMANCE_VERSION_PASSED

rjodinchr:


kpet (approved):
Nice! Thank you very much for relentlessly driving toward that goal :).

Issue #631, opened 2023-11-03
Support config management through files


Rekt3421:
Currently we manage our configurations in Config.cpp. We wish to make a file where we can define the configuration for say using on different platforms.

Rekt3421:
@rjodinchr

kpet:
Closed by #664

PR #664, opened 2024-02-01
Add config.conf and logic to handle config files

Rekt3421:
Adds configs from .conf file, I decided to go against using individual sections to append the name further since we can do so by keeping everything in a unordered map. Env vars are given priority. Addresses issue #631


rjodinchr:
I think we need to think about the path of that file. Where should it be? Should we consider multiple possibilities? like `/etc/something` as well as `<current_directory>` for example.

Rekt3421:
@rjodinchr  I think you are right either we should make a file under /etc/something for system specific configs perhaps? and check the current directory as well. Maybe we can populate a file based on customized configs based on the platform? I am not quite sure myself. WDYT?

src/config.cpp
```diff
@@ -59,31 +59,126 @@ void parse_uint32(void* value_ptr, const char* txt) {
...
+            case config_option_type::string:
+                parse_string(optval, txt);
+                break;
+            case config_option_type::boolean:
+                parse_boolean(optval, txt);
+                break;
+            case config_option_type::uint32:
+                parse_uint32(optval, txt);
+                break;
+            }
```
kpet, line 181: Parsing config options from a file should be done in a dedicated function called from `init_config` (see TODO) rather than mixing the logic with environment variable parsing. This split makes it easier to reason about how the various sources for configuration are layered on top of each other and to change the order if we want to.

src/config.cpp
```diff
@@ -59,31 +59,126 @@ void parse_uint32(void* value_ptr, const char* txt) {
...
-            break;
-        case config_option_type::uint32:
-            parse_uint32(optval, txt);
-            break;
+            if (fileConfigValues.find(opt.name) == fileConfigValues.end() ||
+                fileConfigValues[opt.name].length() == 0) {
+                continue;
+            }
+            auto curr_val = fileConfigValues[opt.name];
+            config_option_type optType = infer_type(fileConfigValues[opt.name]);
```
kpet, line 156: Not sure I understand why you need that `infer_type` function at all. What's wrong with using `opt.type`?

src/config.hpp
```diff
@@ -55,3 +60,5 @@ struct config_struct {
 extern const config_struct config;
 
 extern void init_config();
+
+const std::string config_file = "config.toml";
```
kpet, line 64: I think we'll want to look for the config file in multiple location, probably something like:
- A central system location (e.g. `/etc/clvk.conf`
- The user's home directory (e.g. `~/.config/clvk.conf`). 
- The current directory

Doing this in a portable way is the challenge but we can start by adding logic to look for the file in a list of locations, we'll definitely need that.

Also, I think we probably don't want to use the `.toml` extension. We're only using a subset of toml and that's an implementation detail. I'd suggest `clvk.conf`.

src/config.hpp
```diff
@@ -20,7 +20,12 @@
 #include <cstdint>
 #include <cstdio>
 #include <cstring>
+#include <filesystem>
+#include <fstream>
+#include <iostream>
 #include <string>
+#include <unordered_map>
+#include <vector>
```
kpet, line 28: All these headers are not needed in `config.hpp`  AFAICT. If they're just needed in `config.cpp`, I suggest you include them there.

src/config.toml
```diff
@@ -0,0 +1,39 @@
...
+percentage_of_available_memory_reported = 100 
+max_mem_alloc_size_mb = 1024 
+spirv_validation = 2 
+
+# instrumentation
+queue_profiling_use_timestamp_queries = false
+
+# conditional: Perfetto Backend
+perfetto_trace_max_size = 1024
+perfetto_trace_dest = "clvk.perfetto-trace"
```
kpet, line 39: I see the value of including a sample config file but it should be generated from `config.def` so we don't have to define default values in two places.

kpet (changes requested):
Thanks for this contribution!

I'm wondering whether we should also add the ability to select a config file from the environment using a variable (e.g. `CLVK_CONFIG_FILE=/path/to/clvk.conf`.

We also need to think through how we could test the ability to load config files automatically. Testing all cases would be hard but I think we should at least have _some_ testing.

src/config.hpp
Rekt3421: done, thanks! Was not sure if adding them to the .cpp files was okay.

src/config.cpp
Rekt3421: done thanks!

src/config.cpp
Rekt3421: removed, thanks!

src/config.hpp
Rekt3421: done in this order
1. env var
2. /etc/clvk.conf
3. ~/.config/clvk.conf
4. current dir

src/config.cpp
```diff
@@ -88,10 +206,26 @@ void parse_env() {
...
+    }
+    read_config_file(file_config_values, config_def);
+}
+
 } // namespace
 
 void init_config() {
-    // TODO Parse config file
-    // Parse environment
+    // make a default config file from config.def
```
rjodinchr, line 227: Why is that needed?

src/config.hpp
rjodinchr: > done in this order
> 
> 1. env var
> 2. /etc/clvk.conf
> 3. ~/.config/clvk.conf
> 4. current dir

I'm not sure about this order.
I would have the `current dir` in second position and `/etc/clvk.conf` as the 4th.

rjodinchr:
have a look at https://github.com/kpet/clvk/blob/main/src/log.hpp, and then change all the prints please.

src/config.cpp
Rekt3421: kevin asked to have a default .conf file generated from the .def file

src/config.hpp
Rekt3421: done!

src/config.cpp
rjodinchr: I am not sure he meant that exactly. It feels weird to generate it at run time.

src/config.hpp
rjodinchr: Shouldn't those include be removed from here?

src/config.hpp
```diff
@@ -55,3 +54,10 @@ struct config_struct {
 extern const config_struct config;
 
 extern void init_config();
+
+namespace configs {
+extern void read_config_file(std::unordered_map<std::string, std::string>& umap,
+                             std::ifstream& config_stream);
+// Tries to read configs from pre-defined places and returns the path used.
```
Rekt3421, line 61: please let me know if this is not the way you would like the logic to flow i.e. returning the .conf file we actually ended up using.

tests/config/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+add_gtest_executable(config_test
+    main.cpp
+    ${CLVK_PROJECT_SOURCE_DIR}/src/config.cpp
+    ${CLVK_PROJECT_SOURCE_DIR}/src/log.cpp
```
Rekt3421, line 18: Not sure if this is the right way to do this or not.

src/config.cpp
```diff
@@ -13,12 +13,19 @@
...
+#include <fstream>
+#include <iostream>
+#include <stdlib.h>
+#include <unordered_map>
+#include <vector>
 
 const config_struct config;
 
-namespace {
+namespace configs {
```
Rekt3421, line 28: made a new name space for testing purposes. If this is okay I will file another issue to do more test coverage for these funcs.

tests/config/main.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include <gtest/gtest.h>
+
+#ifdef CLVK_UNIT_TESTING_ENABLED
+#include "config.hpp"
+
+// Test for finding file through env.
+// Note : make sure the other paths dont have config files.
+TEST(ConfigTest, FileFromEnvVar) {
+
+    std::string conf_file = "/tmp/tmp-test-config.conf";
```
rjodinchr, line 26: This is not portable across operating systems.
Consider using `std::filesystem::temp_directory_path`

tests/config/CMakeLists.txt
rjodinchr: I don't think we really want it that way.
I think we would want to do it like other tests using the `CLVK_UNIT_TESTING` meaning: include `unit.hpp` only, not `config.hpp` and use the function you have there. If you do not have everything add them in `unit.hpp`, but I think `clvk_get_config` should be enough for your purpose.

src/config.cpp
rjodinchr: If you switch to using `unit.hpp` instead of `config.hpp` in the test, it should not be needed I guess

rjodinchr (commented):
Please have a look at `<clvk>/.github/workflows/presubmit.yml` to add your test, to have it run by the github actions.

src/config.cpp
Rekt3421: looks like I cant still access the anon namespace funcs outside of the file. Lets keep it i guess?

src/config.hpp
```diff
@@ -17,10 +17,9 @@
 #include "cl_headers.hpp"
 
 #include <algorithm>
-#include <cstdint>
```
rjodinchr, line 20: `cstdint` should stay there as this header contains `uint32_t`

src/config.cpp
```diff
@@ -13,8 +13,16 @@
 // limitations under the License.
 
 #include "config.hpp"
+#include "log.hpp"
+#include "unit.hpp"
```
Rekt3421, line 17: not needed.

.github/workflows/presubmit.yml
```diff
@@ -196,6 +196,9 @@ jobs:
             cp '${{ env.builddir }}'/src/OpenCL.dll '${{ env.builddir }}'
           fi
           echo "testbindir=${{ env.builddir }}" >> $GITHUB_ENV
+      - name: Config tests
+        if: ${{ matrix.compiler-available && matrix.android-abi == '' }}
```
rjodinchr, line 200: We don't need the compiler-available or not to be on android for that test. Only the unit testing to be enabled.

CMakeLists.txt
```diff
@@ -197,6 +197,9 @@ endif ()
 message(STATUS "Vulkan_INCLUDE_DIRS = '${Vulkan_INCLUDE_DIRS}'")
 message(STATUS "Vulkan_LIBRARIES = '${Vulkan_LIBRARIES}'")
 
+set(CONFIG_FILE_PATH "${PROJECT_SOURCE_DIR}/src/config.def")
```
rjodinchr, line 200: Is that part still needed?

src/config.cpp
```diff
@@ -87,11 +194,9 @@ void parse_env() {
         }
     }
 }
-
```
rjodinchr, line 90: Avoid changing parts that are not needed for that PR

src/config.hpp
```diff
@@ -55,3 +55,6 @@ struct config_struct {
 extern const config_struct config;
 
 extern void init_config();
+
+extern void read_config_file(std::unordered_map<std::string, std::string>& umap,
+                             std::ifstream& config_stream);
```
rjodinchr, line 60: missing EOL

src/unit.cpp
```diff
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 #include "device.hpp"
-#include "log.hpp"
```
rjodinchr, line 16: Avoid changing parts that are not needed for that PR

tests/config/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+add_gtest_executable(config_test
+    main.cpp
+)
```
rjodinchr, line 17: missing EOL

tests/config/main.cpp
```diff
@@ -0,0 +1,59 @@
...
+    EXPECT_EQ(clvk_get_config()->cache_dir.value, "testing");
+}
+
+int main(int argc, char** argv) {
+
+    ::testing::InitGoogleTest(&argc, argv);
+    return RUN_ALL_TESTS();
+}
+
+#endif
```
rjodinchr, line 59: missing EOL

rjodinchr (commented):
I think we also want to test that when we have 2 files, the one with the highest priority is overriding things in the low priority.

src/config.cpp
```diff
@@ -59,6 +67,106 @@ void parse_uint32(void* value_ptr, const char* txt) {
...
+}
+
+void parse_config_file() {
+    std::unordered_map<std::string, std::string> file_config_values;
+    std::string conf_file = "clvk.conf";
+    std::ifstream config_stream;
+
+    std::vector<std::string> config_file_paths;
+    config_file_paths.push_back("/etc/clvk.conf");
+    config_file_paths.push_back("~/.config/clvk.conf");
```
rjodinchr, line 127: `~` won't work with windows. I supposed there is something is `std::filesystem` to have the a portable way to express the user directory.

tests/config/main.cpp
```diff
@@ -0,0 +1,42 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifdef _WIN32
+#include <stdlib.h>
+#define set_env(name, value) _putenv_s(name, value)
+#else
+#include <cstdlib>
+#define set_env(name, value) setenv(name, value, 1)
+#endif
```
rjodinchr, line 21: This part can be clean I guess

tests/config/run_test.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
```
rjodinchr, line 2:
```suggestion

set -xe

```

tests/config/run_test.sh
```diff
@@ -0,0 +1,23 @@
...
+# Create the first file with the base path
+touch $base_path/conf_test.conf
+echo "cache_dir=failed" >> $base_path/conf_test.conf
+
+# Create the second file with the base path
+touch $base_path/clvk.conf
+echo "cache_dir=testing" >> $base_path/clvk.conf
+
+export CLVK_CONFIG_FILE=$base_path/clvk.conf
```
rjodinchr, line 20:
```suggestion
export CLVK_CONFIG_FILE="${base_path}/clvk.conf"
```

tests/config/run_test.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# Run test
+$base_path/config_test
```
rjodinchr, line 23: Please clean the created file in a function to execute on exit. Here is an example: https://github.com/kpet/clvk/blob/main/tests/perfetto/check-perfetto-trace.sh#L17

tests/config/main.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include <gtest/gtest.h>
+
+#ifdef CLVK_UNIT_TESTING_ENABLED
+#include "unit.hpp"
+
+// Test for making sure the configs from env var are read
+// and overwrite configs from other paths.
+TEST(ConfigTest, FileFromEnvVar) {
+    clGetPlatformIDs(1, nullptr, nullptr);
+    EXPECT_EQ(clvk_get_config()->cache_dir.value, "testing");
```
rjodinchr, line 34: You also want to check for something set in the current directory file that does not get overwritten.

tests/config/run_test.sh
rjodinchr, line 23:
```suggestion
"${base_path}/config_test"
```

tests/config/run_test.sh
```diff
@@ -0,0 +1,23 @@
...
+# Assign the base path from the argument
+base_path=$1
+
+# Create the first file with the base path
+touch $base_path/conf_test.conf
+echo "cache_dir=failed" >> $base_path/conf_test.conf
+
+# Create the second file with the base path
+touch $base_path/clvk.conf
+echo "cache_dir=testing" >> $base_path/clvk.conf
```
rjodinchr, line 18:
```suggestion
echo "cache_dir=testing" > "${base_path}/clvk.conf"
```

tests/config/main.cpp
```diff
@@ -0,0 +1,42 @@
...
+#ifdef _WIN32
+#include <stdlib.h>
+#define set_env(name, value) _putenv_s(name, value)
+#else
+#include <cstdlib>
+#define set_env(name, value) setenv(name, value, 1)
+#endif
+
+#include <filesystem>
+#include <fstream>
```
rjodinchr, line 24:
```suggestion
```

src/config.cpp
Rekt3421: leaving this as it is since theres nothing in ```std::filesystem```

tests/config/run_test.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+##set -xe
```
rjodinchr, line 3:
```suggestion
set -xe
```
It will help understand what is happening in the CI.

tests/config/run_test.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# Create the second file with the base path
+echo "cache_dir=failed" > "${base_path}/clvk.conf"
+echo "compiler_temp_dir=not/overwritten/" > "${base_path}/clvk.conf"
+
+
+export CLVK_CONFIG_FILE="${base_path}/conf_test.conf"
+
+# Run test
+"${base_path}/config_test"
```
rjodinchr, line 31: missing EOL

tests/config/run_test.sh
Rekt3421: just uploaded a new one. Thanks!

src/config.cpp
```diff
@@ -13,8 +13,20 @@
...
 
 #include <cassert>
+#include <cstdlib>
+#include <filesystem>
+#include <fstream>
+#include <iostream>
+#include <unordered_map>
+#include <vector>
+
+#ifdef _WIN32
```
rjodinchr, line 27: I think you did not end up using Windows specific code. If that is correct, remove it.

.github/workflows/presubmit.yml
```diff
@@ -196,6 +196,9 @@ jobs:
             cp '${{ env.builddir }}'/src/OpenCL.dll '${{ env.builddir }}'
           fi
           echo "testbindir=${{ env.builddir }}" >> $GITHUB_ENV
+      - name: Config tests.
+        if: ${{ matrix.unit }}
+        run: bash ./tests/config/run_test.sh ${{ env.testbindir }}
```
rjodinchr, line 201:
```suggestion
      - name: Config tests.
        shell: bash
        if: ${{ matrix.unit }}
        run: ./tests/config/run_test.sh ${{ env.testbindir }}
```

src/config.hpp
```diff
@@ -55,3 +55,6 @@ struct config_struct {
 extern const config_struct config;
 
 extern void init_config();
+
```
rjodinchr, line 58: I think you do not need it anymore, right?

src/config.cpp
```diff
@@ -13,8 +13,16 @@
...
 
 #include "config.hpp"
+#include "log.hpp"
+#include "unit.hpp"
 
 #include <cassert>
+#include <cstdlib>
+#include <filesystem>
+#include <fstream>
+#include <iostream>
```
rjodinchr, line 23: Do you need both still?

tests/config/run_test.sh
```diff
@@ -0,0 +1,31 @@
...
```
rjodinchr, line 31:
```suggestion
# Run test
CLVK_CONFIG_FILE="${base_path}/conf_test.conf" \
"${base_path}/config_test"
```
Because one may want to run it locally, let's avoid exporting.

tests/config/run_test.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# Assign the base path from the argument
+base_path=$1
+
+# Create the first file with the base path
+echo "cache_dir=testing" > "${base_path}/conf_test.conf"
+
+# Create the second file with the base path
+echo "cache_dir=failed" > "${base_path}/clvk.conf"
+echo "compiler_temp_dir=not/overwritten/" > "${base_path}/clvk.conf"
```
rjodinchr, line 25: are you sure that this will be the current directory?
Maybe do something like:
```
TEMP_DIR="$(mktemp -d)"
pushd "${TEMP_DIR}"

...
[ create the files & execute the test ]
...

popd
 ```

Comment (rjodinchr) [reply tests/config/run_test.sh:25]: Note that it would also mean, you would not need base_path I believe

Comment (Rekt3421) [reply tests/config/run_test.sh:25]: I think I will still need the binary path for the config_test binary

File: tests/config/run_test.sh [hunk 36]
```diff
@@ -0,0 +1,28 @@
...
+  exit 1
+fi
+
+TEMP_DIR="$(mktemp -d)"
+pushd "${TEMP_DIR}"
+
+function clean() {
+    rm -r "${TEMP_DIR}"
+}
+#trap clean EXIT
```
rjodinchr, line 16:
```suggestion
trap clean EXIT
```

tests/config/run_test.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# Assign the base path from the argument
+binary_path=$1
+
+cp "${binary_path}/clvk.conf" "${TEMP_DIR}"
+cp "${binary_path}/conf_test.conf" "${TEMP_DIR}"
+
+# Run test
+CLVK_CONFIG_FILE="${TEMP_DIR}/conf_test.conf" \
```
rjodinchr, line 25:
```suggestion
# Run test
CLVK_CONFIG_FILE="${binary_path}/conf_test.conf" \
```
No real need for the copy right?

src/config.hpp
```diff
@@ -18,7 +18,6 @@
 
 #include <algorithm>
 #include <cstdint>
-#include <cstdio>
```
rjodinchr, line 21: Do not remove please

rjodinchr (changes requested):

rjodinchr (approved):

src/config.cpp
```diff
@@ -59,6 +66,106 @@ void parse_uint32(void* value_ptr, const char* txt) {
...
+            continue;
+        } else {
+            // Parse key-value pairs
+            size_t pos = line.find('=');
+            if (pos != std::string::npos) {
+                std::string key = trim(line.substr(0, pos));
+                std::string value = trim(line.substr(pos + 1));
+                // Store values (if any)
+                if (value != "") {
+                    umap[key] = value;
```
rjodinchr, line 111:
```suggestion
                    umap[key] = value;
                    cvk_info_fn("%s = %s", key.c_str(), value.c_str());
```

src/config.cpp
```diff
@@ -59,6 +66,106 @@ void parse_uint32(void* value_ptr, const char* txt) {
...
+    }
+    for (auto& curr_path : config_file_paths) {
+        if (!std::filesystem::exists(curr_path)) {
+            continue;
+        }
+        config_stream.open(curr_path);
+        if (!config_stream.is_open()) {
+            cvk_error("Error opening config file - %s", curr_path.c_str());
+        }
+        read_config_file(file_config_values, config_stream);
```
rjodinchr, line 144:
```suggestion
        cvk_info_fn("%s", curr_path.c_str());
        read_config_file(file_config_values, config_stream);
```

tests/config/run_test.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
kpet, line 1:
```suggestion
#!/usr/bin/env bash
```

Tends to be a bit more portable

tests/config/run_test.sh
```diff
@@ -0,0 +1,27 @@
...
+TEMP_DIR="$(mktemp -d)"
+pushd "${TEMP_DIR}"
+
+function clean() {
+    rm -r "${TEMP_DIR}"
+}
+trap clean EXIT
+
+# Assign the base path from the argument
+binary_path=$1
```
kpet, line 19:
```suggestion
binary_path="$1"
```

src/config.cpp
```diff
@@ -59,6 +66,106 @@ void parse_uint32(void* value_ptr, const char* txt) {
...
+    }
+}
+
+void read_config_file(std::unordered_map<std::string, std::string>& umap,
+                      std::ifstream& config_stream) {
+
+    std::string line;
+    while (std::getline(config_stream, line)) {
+        // Ignore comments and empty lines
+        if (line.empty() || line[0] == '#') {
```
kpet, line 96: This means lines that have spaces before the '#' won't be treated as comments. I think that's acceptable but I can see it tripping someone.

src/config.cpp
```diff
@@ -59,6 +66,106 @@ void parse_uint32(void* value_ptr, const char* txt) {
...
+            continue;
+        }
+        line = trim(line);
+        // Check for section headers
+        if (line[0] == '[' && line[line.size() - 1] == ']') {
+            continue;
```
kpet, line 102: Why do we have any support for section headers? These are not semantic in any way and in effect are just an alternative comment syntax :). If you have plans for them in the future, fine to keep this in, otherwise I'd suggest removing this (more untested code!).

src/config.cpp
```diff
@@ -59,6 +66,106 @@ void parse_uint32(void* value_ptr, const char* txt) {
...
+        } else {
+            // Parse key-value pairs
+            size_t pos = line.find('=');
+            if (pos != std::string::npos) {
```
kpet, line 106: Should we report to the user that the line is malformed when we can't find a '='? I think it's worth at least a warning.

src/config.cpp
kpet: Maybe better for the debug log level? This could be very noisy. Maybe it makes sense to introduce a 'config' log group otherwise?

src/config.cpp
```diff
@@ -59,6 +66,106 @@ void parse_uint32(void* value_ptr, const char* txt) {
...
+    // Extract the trimmed and unquoted substring
+    std::string trimmed = str.substr(first, (last - first + 1));
+
+    // Check if the trimmed string still starts and ends with quotes
+    if (trimmed.front() == '"' && trimmed.back() == '"') {
+        return trimmed.substr(1, trimmed.size() - 2); // Remove the quotes
+    } else {
+        return trimmed;
+    }
+}
```
kpet, line 88: I'm not sure I understand why we need to handle quotes at all. Can you explain?

tests/config/clvk.conf
```diff
@@ -0,0 +1,2 @@
+cache_dir=failed
+compiler_temp_dir=not/overwritten/
```
kpet, line 2: We should have an option of each type in the test file so all parse functions are exercised.

tests/config/run_test.sh
```diff
@@ -0,0 +1,27 @@
...
+
+cp "${binary_path}/clvk.conf" "${TEMP_DIR}"
+
+# Run test
+CLVK_CONFIG_FILE="${binary_path}/conf_test.conf" \
```
kpet, line 24: We probably want at least one env override to test that environment variables take precedence over config files.

.github/workflows/presubmit.yml
```diff
@@ -196,6 +196,10 @@ jobs:
             cp '${{ env.builddir }}'/src/OpenCL.dll '${{ env.builddir }}'
           fi
           echo "testbindir=${{ env.builddir }}" >> $GITHUB_ENV
+      - name: Config tests.
```
kpet, line 199:
```suggestion
      - name: Config tests
```

src/config.cpp
```diff
@@ -59,6 +66,106 @@ void parse_uint32(void* value_ptr, const char* txt) {
...
+    std::string conf_file = "clvk.conf";
+    std::ifstream config_stream;
+
+    std::vector<std::string> config_file_paths;
+    config_file_paths.push_back("/etc/clvk.conf");
+    config_file_paths.push_back("~/.config/clvk.conf");
+    config_file_paths.push_back(
+        (std::filesystem::current_path() / conf_file).string());
+    // First check if env var has file
+    std::string conv_file_env_var = "CLVK_CONFIG_FILE";
```
kpet, line 130: This environment variable should be mentioned in the documentation.

kpet (changes requested):
Nice work! This is missing a few finishing touches and documentation. Ideally we'd rework the section on environment variables into a more generic description of the config system, explaining that there is support for files and environment variables. If you're unsure of how to go about it, then just document `CLVK_CONFIG_FILE` and we can do the rest later.

I'm sure we will need to spend more time on config file path management and system integration but this PR is good enough as a first step IMO.

src/config.cpp
rjodinchr: I don't see it being that noisy, but I don't mind, as long as we have them it's fine with me. I agree a log group would make sense.

src/config.cpp
Rekt3421: ill move the trim func first to eliminate any whitespace.

src/config.cpp
Rekt3421: good point. I thought some people may try to put quotes around the confs. Removing this functionality.

tests/config/run_test.sh
Rekt3421: I think we are already doing so. As clvk.conf and conf_test.conf are both modifying the cache_dir configuration. Let me know if I am misunderstanding.

src/config.cpp
Rekt3421: Just removed it for now.

src/config.cpp
rjodinchr: It feels like this functionality has not been removed, am I missing something?

src/config.cpp
```diff
@@ -59,6 +66,109 @@ void parse_uint32(void* value_ptr, const char* txt) {
     cfgval->set = true;
 }
 
+// Helper function to trim whitespace and remove quotations
+std::string trim(const std::string& str) {
+    size_t first = str.find_first_not_of(" \t\n"); // Also include quotes
```
rjodinchr, line 71: The comment does not match the content

src/config.cpp
```diff
@@ -59,6 +66,109 @@ void parse_uint32(void* value_ptr, const char* txt) {
...
+    }
+    for (auto& curr_path : config_file_paths) {
+        if (!std::filesystem::exists(curr_path)) {
+            continue;
+        }
+        config_stream.open(curr_path);
+        if (!config_stream.is_open()) {
+            cvk_error("Error opening config file - %s", curr_path.c_str());
+        }
+        cvk_info_group(loggroup::cfg, "Parsing config file '%s'",
```
rjodinchr, line 145:
```suggestion
        cvk_info_group_fn(loggroup::cfg, "Parsing config file '%s'",
```

src/log.cpp
```diff
@@ -48,6 +48,7 @@ static uint64_t init_logging_groups() {
         {"api", loggroup::api},
         {"event", loggroup::event},
         {"validation", loggroup::validation},
+        {"cfg", loggroup::cfg},
```
rjodinchr, line 51:
```suggestion
        {"config", loggroup::config},
```

tests/config/run_test.sh
```diff
@@ -0,0 +1,38 @@
...
+trap clean EXIT
+
+# Assign the base path from the argument
+binary_path="$1"
+
+cp "${binary_path}/clvk.conf" "${TEMP_DIR}"
+
+# Assing an env var for testing
+default_color_val=$(echo CLVK_LOG_COLOUR)
+export CLVK_LOG_COLOUR=1
```
rjodinchr, line 25: Do not export, just define it as `CLVK_CONFIG_FILE` when running the test.

src/config.cpp
```diff
@@ -59,6 +66,109 @@ void parse_uint32(void* value_ptr, const char* txt) {
...
+            std::string key = trim(line.substr(0, pos));
+            std::string value = trim(line.substr(pos + 1));
+            // Store values (if any)
+            if (value != "") {
+                umap[key] = value;
+                cvk_info_group_fn(loggroup::cfg, "%s = %s", key.c_str(),
+                                  value.c_str());
+            }
+        } else {
+            cvk_warn_group(loggroup::cfg, "%s , %s",
```
rjodinchr, line 113:
```suggestion
            cvk_warn_group_fn(loggroup::cfg, "%s , %s",
```

rjodinchr (changes requested):

src/log.cpp
kpet: +1

Rekt3421: so there is another variable in this context called config hence why I used cfg  change it.

src/config.cpp
rjodinchr: The comment mentions "quotes" while they are not part of the argument of `find_first_not_of`.

src/config.cpp
```diff
@@ -59,6 +66,103 @@ void parse_uint32(void* value_ptr, const char* txt) {
...
+std::string trim(const std::string& str) {
+    size_t first = str.find_first_not_of(" \t\n"); // Also include quotes
+    size_t last = str.find_last_not_of(" \t\n");
+
+    // Check for valid range
+    if (first == std::string::npos || last == std::string::npos) {
+        return str; // Empty or only whitespace/quotes
+    }
+
+    // Extract the trimmed and unquoted substring
```
rjodinchr, line 79: The substring has not been unquoted. Please update the comment

tests/config/run_test.sh
rjodinchr: This is not resolved, I still see the export. Just define `CLVK_LOG_COLOUR=1` as the test does for `CLVK_CONFIG_FILE`. That way, the env variable is not exported, it is just set for the execution of `config_test` and does not need to be restored after the run.

src/log.cpp
rjodinchr: Ah...
Here is a proposition: let's put the 2 enums (`loggroup` & `loglevel`) in a namespace `clvk_log` (I wanted to call it just `log` but it is already defined by the logarithmic function).
And because of the namespace let's call `loggroup` `group` and `loglevel` `level`?
Here is a patch of what I did that work on my machine.
[clvk-664.patch](https://github.com/kpet/clvk/files/14598498/clvk-664.patch)

rjodinchr (changes requested):

src/log.cpp
kpet: Ok, let's just keep cfg for simplicity then.

src/config.cpp
```diff
@@ -97,10 +106,14 @@ void read_config_file(std::unordered_map<std::string, std::string>& umap,
         if (pos != std::string::npos) {
             std::string key = trim(line.substr(0, pos));
             std::string value = trim(line.substr(pos + 1));
+            // Skip if an env var with the same name exists.
+            if (getenv(get_clvk_env_name(key).c_str()) != nullptr) {
```
rjodinchr, line 110: This is not checking that the parsing of that env went well. If `parse_env` did not end up really setting the value, I don't think we should skip here.

I think it would be easier to just use `parse_env` twice. Once in each `init_config` functions.

rjodinchr (approved):

README.md
```diff
@@ -407,6 +407,11 @@ variables. Here's a quick guide:
   folder to store compiled program data used in a single run. This folder shall
   have write permission (default: current directory).
 
+* `CLVK_CONFIG_FILE` specifies the path to a configuration file. You can see a
+  sample of the file under test/config/conf_test.conf. The file specified by env
+  var takes precedence over all other config files but does not take precedence
+  over env variable defined values.
```
kpet, line 413: s/env/environment/g

src/config.cpp
```diff
@@ -59,14 +66,116 @@ void parse_uint32(void* value_ptr, const char* txt) {
...
+    std::string conf_file = "clvk.conf";
+    std::ifstream config_stream;
+
+    std::vector<std::string> config_file_paths;
+    config_file_paths.push_back("/etc/clvk.conf");
+    config_file_paths.push_back("~/.config/clvk.conf");
+    config_file_paths.push_back(
+        (std::filesystem::current_path() / conf_file).string());
+    // First check if env var has file
+    std::string conv_file_env_var = "CLVK_CONFIG_FILE";
```
kpet, line 134: Why did you introduce a function to compose the name of environment variables used for configuration without using it?

src/config.cpp
```diff
@@ -59,14 +66,116 @@ void parse_uint32(void* value_ptr, const char* txt) {
...
+            cvk_error("Error opening config file - %s", curr_path.c_str());
+        }
+        cvk_info_group_fn(loggroup::cfg, "Parsing config file '%s'",
+                          curr_path.c_str());
+        read_config_file(file_config_values, config_stream);
+    }
+
+    for (auto& opt : gConfigOptions) {
+        if (file_config_values.find(opt.name) == file_config_values.end() ||
+            file_config_values[opt.name].length() == 0) {
```
kpet, line 155: I think we don't need to check that the length is zero since option values only get stored in the map if they are not empty strings (see `read_config_file`).

src/config.cpp
```diff
@@ -90,8 +199,9 @@ void parse_env() {
 
 } // namespace
 
+void init_config_from_env_only() { parse_env(); }
```
kpet, line 202: Why are we now parsing the environment config twice?

kpet (approved):
Let's call this good enough. I'll tidy up on top.

src/config.cpp
rjodinchr: https://github.com/kpet/clvk/pull/679/files#r1527486001

Issue #578, opened 2023-07-16
Fix/improve printf support


kpet:
- [ ] Support batching kernels that use printf
- [ ] Remove serialisation between kernel regions when printf is used
- [ ] Introduce unit test callback to give tests access to printf's output stream and remove all stdout capture logic (https://github.com/kpet/clvk/pull/565#discussion_r1264651226)
- [ ] Fix concurrent accesses to the printf buffer (https://github.com/kpet/clvk/pull/565#discussion_r1264652192). Add tests.

Rekt3421:
Hey Kevin can you please check off point 3? I will start working on other parts now.

rjodinchr:
To fix the issue with concurrent accesses, we mentioned during our last discussion that we might just not perform any reset.
I don't think this would be working because, when we print on clvk side, we go from offset 0, to the end offset.
The issue is with both `0` (the start offset) and the end offset (first word of the printf buffer).

So 2 possibilities:
- continuing the work that @Rekt3421 started (having one printf buffer per kernel instead of having one per queue)
- changing the printf interface in clspv to use another buffer containing 2 words (a per-kernel buffer) to indicate the start&end offset

@kpet what do you think?

PR #668, opened 2024-03-05
This change introduces using cl_arm_printf with callbacks for using printf inside the kernel

Rekt3421:
This changes is a part of a series which will address #578 . This change addresses point 3 out of 4.


tests/api/printf.cpp
```diff
@@ -40,6 +40,15 @@ static void releaseStdout(int fd) {
     close(fd);
 }
 
+#define RETURN_ON_FAILURE(err, ret)                                            \
```
Rekt3421, line 43: This is similar to the main one. Was having some trouble using ASSERT_CL_SUCCESS in some cases so just copied and reused this.

Rekt3421:
There is one slight issue though how do I verify the contents of the printf buffer? I could not access the string from within the callback ( I am assuming perhaps this is due to some synchronization issue?)

tests/api/printf.cpp
```diff
@@ -104,64 +115,131 @@ static std::string getStdoutFileName(temp_folder_deletion& temp) {
...
+                     void* user_data) {
+    captured_output.append(buffer, len);
+}
+
+void get_command_queue_printf(cl_command_queue& queue, cl_context new_ctx,
+                              cl_device_id device) {
+    cl_int err;
+    queue = clCreateCommandQueue(new_ctx, device, 0, &err);
+    ASSERT_CL_SUCCESS(err);
+}
```
rjodinchr, line 174: This function feels unnecessary

rjodinchr (changes requested):
You should rely on the framework used by `api_tests`. Everything is in `testcl.hpp`.
It will allow to easily manage release/destruction as well as testing the error code (removing the need for `RETURN_ON_FAILURE` that you added)

You will need to add things there to create your context I believe.

In the end, we should not have any function starting with `cl` in `printf.cpp`.

tests/api/printf.cpp
```diff
@@ -104,64 +115,131 @@ static std::string getStdoutFileName(temp_folder_deletion& temp) {
...
+    auto device = get_device_test(platform);
+    cl_context_properties properties[] = {
+        /* Enable a printf callback function for this context. */
+        CL_PRINTF_CALLBACK_ARM, (cl_context_properties)printf_callback,
+
+        /* Request a minimum printf buffer size of 4MiB for devices in the
+        context that support this extension. */
+        CL_PRINTF_BUFFERSIZE_ARM, (cl_context_properties)0x100000,
+
+        CL_CONTEXT_PLATFORM, (cl_context_properties)gPlatform, 0};
```
rjodinchr, line 188: I understand that this is part of the sample example [here](https://registry.khronos.org/OpenCL/extensions/arm/cl_arm_printf.html), but I don't think it is needed.

PR #669, opened 2024-03-06
set queue global priority when available

rjodinchr:


README.md
```diff
@@ -443,6 +443,14 @@ variables. Here's a quick guide:
 * `CLVK_FORCE_SUBGROUP_SIZE` specifies the subgroup size to use, overriding
   everything.
 
+* `CLVK_QUEUE_GLOBAL_PRIORITY` specifies the queue global priority to use if it
+  is supported by the driver:
+
+  * `0`: `VK_QUEUE_GLOBAL_PRIORITY_LOW_KHR` (default)
```
rjodinchr, line 449: if you prefer to have something else as default, I'm open to it, but I'll set it to low in ChromeOS, as we want compute workload to have the lowest priority to avoid slowing down the system for non-critical features.

src/device.cpp
```diff
@@ -716,10 +722,36 @@ bool cvk_device::create_vulkan_queues_and_device(uint32_t num_queues,
     cvk_info("Creating Vulkan device and queues");
     // Give all queues the same priority
     std::vector<float> queuePriorities(num_queues, 1.0f);
+    void* pNext = nullptr;
+    VkDeviceQueueGlobalPriorityCreateInfoKHR globalPriorityCreateInfo;
+    cvk_info_fn("queue global priority: %u",
+                m_features_queue_global_priority.globalPriorityQuery);
+    if (m_features_queue_global_priority.globalPriorityQuery) {
```
kpet, line 729: This does not guarantee that specific priorities are supported so queue creation might fail if a user passes a priority that the Vulkan implementation does not support. This might be fine in practice. If we want to do better we should use `VkQueueFamilyGlobalPriorityPropertiesKHR` to get the list of supported priorities and maybe fall back to one that's supported with a warning when a user requests an unsupported priority. Fine if you want to do this in this PR, fine if you don't.

kpet (changes requested):

src/device.cpp
rjodinchr: I have tried to use `VkQueueFamilyGlobalPriorityPropertiesKHR`, but all the drivers I have tested return no supported priority, while they do support at least medium. And sometimes even more.
I can add a warning, but I'm not fond of it as it would always be printed in ChromeOS while it would not mean anything.

src/device.cpp
```diff
@@ -345,6 +348,9 @@ void cvk_device::init_features(VkInstance instance) {
             VER_EXT_FEAT(VK_MAKE_VERSION(1, 2, 0),
                          VK_KHR_BUFFER_DEVICE_ADDRESS_EXTENSION_NAME,
                          m_features_buffer_device_address),
+            VER_EXT_FEAT(VK_MAKE_VERSION(1, 1, 0),
```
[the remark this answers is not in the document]
rjodinchr: I don't know why I've done that. Reading back the extension, I agree that it has not been promoted to any core version.

PR #671, opened 2024-03-08
use cl_arm_printf extension for testing

Rekt3421:


src/printf.hpp
```diff
@@ -25,7 +26,9 @@ struct printf_descriptor {
 };
 
 using printf_descriptor_map_t = std::unordered_map<uint32_t, printf_descriptor>;
+typedef std::function<void(const char*, size_t)> printf_callback_func;
```
rjodinchr, line 29: This is not the prototype defined by the `cl_arm_printf` [extension](https://registry.khronos.org/OpenCL/extensions/arm/cl_arm_printf.html#_modifications_to_the_opencl_api_specification):
```
(void)(*callback)(const char *buffer, size_t len, size_t complete, void *user_data)
```

src/queue.cpp
```diff
@@ -297,21 +299,36 @@ cl_int cvk_command_group::execute_cmds() {
...
         TRACE_END();
     }
     return global_status;
 }
 
+bool cvk_command_queue::get_printf_callback(printf_callback_func& callback) {
+    auto& properties = m_context->properties();
+    for (unsigned i = 0; i < properties.size(); i += 2) {
+        if (properties[i] == CL_PRINTF_CALLBACK_ARM) {
+            void (*func_ptr)(const char*, long unsigned int) =
```
rjodinchr, line 322: I think it should be possible to use the `printf_callback_func` type here somehow.

src/queue.hpp
```diff
@@ -172,11 +172,21 @@ struct cvk_command_queue : public _cl_command_queue,
         return m_command_pool.free_command_buffer(cmdbuf);
     }
 
+    size_t get_printf_buffer_size() {
+        auto& properties = m_context->properties();
+        for (unsigned i = 0; i < properties.size(); i += 2) {
```
rjodinchr, line 177: This is kind of a duplicate of `cvk_command_queue::get_printf_callback`. Maybe merge that in a function in `cvk_context`, something like `get_property` or `get_property_index`.

src/queue.hpp
```diff
@@ -757,6 +768,12 @@ struct cvk_command_kernel final : public cvk_command_batchable {
         : cvk_command_batchable(CL_COMMAND_NDRANGE_KERNEL, q), m_kernel(kernel),
           m_dimensions(dims), m_ndrange(ndrange), m_pipeline(VK_NULL_HANDLE),
           m_argument_values(nullptr) {}
+    cvk_command_kernel(cvk_command_queue* q, cvk_kernel* kernel, uint32_t dims,
```
rjodinchr, line 771: I think this is not used in the end, right? If yes please remove

tests/api/testcl.hpp
```diff
@@ -757,3 +757,20 @@ class WithProfiledCommandQueue : public WithCommandQueue {
 protected:
     void SetUp() override { SetUpQueue(CL_QUEUE_PROFILING_ENABLE); }
 };
+
+class WithPrintfEnabled : public WithCommandQueue {
+protected:
+    void SetUp() override{};
+    void TearDown() override {
+        cl_int err = clReleaseContext(m_context);
+        ASSERT_CL_SUCCESS(err);
```
rjodinchr, line 766: We will need to call `WithContext::TearDown()` and `WithCommandQueue::TearDown()` depending on whether the creation succeeded.

tests/api/testcl.hpp
```diff
@@ -757,3 +757,20 @@ class WithProfiledCommandQueue : public WithCommandQueue {
...
+    }
+
+    void SetupPrintfCallback(const cl_context_properties* props) {
+        cl_int err;
+        m_context = clCreateContext(props, 1, &gDevice, nullptr, nullptr, &err);
```
rjodinchr, line 771: Let's keep `clCreateContext` in `WithContext`. Add a `SetUpContext(const cl_context_properties*)` method in `WithContext`.

tests/api/testcl.hpp
```diff
@@ -757,3 +757,20 @@ class WithProfiledCommandQueue : public WithCommandQueue {
...
+        ASSERT_CL_SUCCESS(err);
+        m_queue = clCreateCommandQueue(m_context, gDevice, 0, &err);
```
rjodinchr, line 773: Move `WithContext::SetUp()` out of `WithCommandQueue::SetUpQueue` and use it here.
Don't forget to add calls to `WithContext::SetUp()` where `SetUpQueue` is already called.

rjodinchr (changes requested):
Looking good overall

Clean the unneeded changes, and resolve my comments and it should be good.
`src/program.cpp` is also using `config.printf_buffer_size`, please update it.

rjodinchr:
I think we also want to add `cl_arm_printf` to `cvk_device::build_extension_ils_list`.

src/printf.cpp
```diff
@@ -238,12 +240,17 @@ void process_printf(char*& data, const printf_descriptor_map_t& descs,
         next_part = part_end;
         arg_idx++;
     }
-
-    printf("%s", printf_out.str().c_str());
+    auto output = printf_out.str().c_str();
+    if (printf_cb != nullptr) {
+        printf_cb(output, buffer_size, true, nullptr);
```
rjodinchr, line 245: `complete` is not always `true`. Please update with the right condition.

src/queue.cpp
```diff
@@ -1063,7 +1063,16 @@ cl_int cvk_command_kernel::do_post_action() {
...
         }
-        return cvk_printf(buffer, m_kernel->program()->printf_descriptors());
+        auto cb_index =
+            m_queue->context()->get_property_index(CL_PRINTF_CALLBACK_ARM);
+        if (cb_index == -1) {
+            cvk_error_fn("failed to get printf callback function");
+            return CL_INVALID_PROPERTY;
+        }
+        auto all_props = m_queue->context()->properties();
+        auto cb_func = (printf_callback_func*)all_props[cb_index];
```
rjodinchr, line 1073: Can we do that once in the queue constructor to avoid doing it at each execution?

src/queue.hpp
```diff
@@ -174,9 +174,18 @@ struct cvk_command_queue : public _cl_command_queue,
...
-                context(), 0, config.printf_buffer_size, nullptr, &status);
+            cl_int status = CL_SUCCESS;
+            auto buff_size_prop_index =
+                m_context->get_property_index(CL_PRINTF_BUFFERSIZE_ARM);
+            if (buff_size_prop_index == -1) {
+                cvk_error_fn("Could not get printf buffer size");
+                status = CL_INVALID_BUFFER_SIZE;
+            }
+            auto all_props = m_context->properties();
+            auto buff_size = all_props[buff_size_prop_index];
```
rjodinchr, line 185: This is a duplicate of what is done in `src/program.cc`. Please do that in the context constructor and keep the buffer size to return it without having anything to compute.

tests/api/testcl.hpp
```diff
@@ -208,6 +208,12 @@ class WithContext : public ::testing::Test {
         ASSERT_CL_SUCCESS(err);
     }
 
+    void SetUpContext(const cl_context_properties* props) {
```
rjodinchr, line 211: Please use it in `WithContext::SetUp` to avoid duplication

tests/api/testcl.hpp
```diff
@@ -505,7 +511,6 @@ class WithCommandQueue : public WithContext {
 #ifndef COMPILER_AVAILABLE
         GTEST_SKIP();
 #endif
-        WithContext::SetUp();
```
rjodinchr, line 508: We need to add it in `WithCommandQueue::SetUp` otherwise it will break the tests.

tests/api/testcl.hpp
```diff
@@ -755,5 +760,28 @@ class WithCommandQueue : public WithContext {
...
+protected:
+    void SetUp() override{};
+    void TearDown() override {
+        cl_int err = clReleaseContext(m_context);
+        ASSERT_CL_SUCCESS(err);
+        if (m_queue != nullptr) {
+            WithCommandQueue::TearDown();
+        }
+        if (m_context != nullptr) {
+            WithContext::TearDown();
```
rjodinchr, line 779: It will have already been called by `WithCommandQueue::TearDown` if `m_queue != nullptr`.

tests/api/printf.cpp
```diff
@@ -227,4 +227,34 @@ TEST_F(WithCommandQueue, PrintfMissingLengthModifier) {
     ASSERT_STREQ(printf_buffer, message);
 }
 
+void printf_callback(const char* buffer, size_t len, size_t complete,
+                     void* user_data) {
+    ASSERT_EQ(strlen(buffer), len);
+    printf("Received data: %s (length: %zu)\n", buffer, len);
+}
```
rjodinchr, line 234: This is not really testing anything. If we return an empty buffer, it would pass.
We cannot rely on visual check

rjodinchr (changes requested):

tests/api/printf.cpp
Rekt3421: what would be an appropriate way to validate the contents of the buffer?

tests/api/testcl.hpp
```diff
@@ -760,5 +761,23 @@ class WithCommandQueue : public WithContext {
...
+        WithContext::SetUp();
+        SetUpQueue(CL_QUEUE_PROFILING_ENABLE);
+    }
+};
+
+class WithPrintfEnabled : public WithCommandQueue {
+protected:
+    void SetUp() override{};
+    void TearDown() override {
+        if (m_queue != nullptr) {
```
rjodinchr, line 774: Should not be needed, if there is a test to do, it should be inside `WithCommandQueue::TearDown`.

tests/api/testcl.hpp
```diff
@@ -507,15 +504,18 @@ class WithCommandQueue : public WithContext {
...
 #ifndef COMPILER_AVAILABLE
         GTEST_SKIP();
 #endif
-        WithContext::SetUp();
         auto queue = CreateCommandQueue(device(), properties);
         m_queue = queue.release();
     }
 
-    void SetUp() override { SetUpQueue(0); }
+    void SetUp() override {
```
rjodinchr, line 515: Let's create a `WithCommandQueue::SetUpWithQueueProperties` to avoid having to duplicate what is done here in `WithProfiledQueue::SetUp`.

tests/api/testcl.hpp
```diff
@@ -203,13 +203,14 @@ static void GetDeviceAndHostTimer(cl_device_id device, cl_ulong* device_ts,
...
     cl_context m_context;
+    const cl_context_properties* m_props = nullptr;
 
     cl_platform_id platform() const { return gPlatform; }
 
     void SetUp() override {
         cl_int err;
         m_context =
-            clCreateContext(nullptr, 1, &gDevice, nullptr, nullptr, &err);
+            clCreateContext(m_props, 1, &gDevice, nullptr, nullptr, &err);
```
rjodinchr, line 213: Let's have a `WithContext::SetUpWithContextProperties` and remove `m_props`.

tests/api/printf.cpp
```diff
@@ -227,4 +234,35 @@ TEST_F(WithCommandQueue, PrintfMissingLengthModifier) {
...
+
+    sprintf(source, "kernel void test_printf() { printf(\"%s\");}", message);
+    /* Create a cl_context with a printf_callback and user specified buffer
+     * size. */
+    cl_context_properties properties[] = {
+        /* Enable a printf callback function for this context. */
+        CL_PRINTF_CALLBACK_ARM,
+        (cl_context_properties)printf_callback,
+        CL_PRINTF_BUFFERSIZE_ARM,
+        strlen(message),
```
rjodinchr, line 254: No need to have the buffer than small, let's make it a bit bigger than the message that will be passed.

tests/api/printf.cpp
```diff
@@ -227,4 +234,35 @@ TEST_F(WithCommandQueue, PrintfMissingLengthModifier) {
...
 }
 
+void printf_callback(const char* buffer, size_t len, size_t complete,
+                     void* user_data) {
+    ASSERT_EQ(strlen(buffer), len);
+}
+
+TEST_F(WithPrintfEnabled, PrintSimple) {
+    const char message[] = "Hello World!";
+    char source[512];
```
rjodinchr, line 244: Maybe make a `#define` or something to have `"Hello World!" in a single spot and be able in `printf_callback` to compare the buffer to the expected string.

src/printf.cpp
```diff
@@ -238,12 +239,17 @@ void process_printf(char*& data, const printf_descriptor_map_t& descs,
...
         arg_idx++;
     }
-
-    printf("%s", printf_out.str().c_str());
+    auto output = printf_out.str().c_str();
+    if (printf_cb != nullptr) {
+        printf_cb(output, buffer_size, data >= data_end, nullptr);
+    } else {
+        printf("%s", output);
+    }
```
rjodinchr, line 247:
```suggestion
    auto output = printf_out.str();
    if (printf_cb != nullptr) {
        auto len = output.size();
        printf_cb(output.c_str(), len, data+len >= data_end, nullptr);
    } else {
        printf("%s", output.c_str());
    }
```
`printf_out.str()` returns a string. That will be destroyed if no handle is kept on it. Thus the current implementation has a `const char *` (through `c_str()`) on a buffer that is no longer allocated.

`len` and `complete` in `printf_cb` also seems to be incorrect in the current implementation

rjodinchr (changes requested):
The issue is with the `output` variable in `process_printf`, I proposed a fix below.

Also please clean the PR so that you only modify what is really needed (do not format unmodified lines)

tests/api/printf.cpp
rjodinchr: Comparing it to the expectation

src/printf.cpp
Rekt3421: thanks for that Romaric!

src/context.hpp
```diff
@@ -41,6 +41,14 @@ struct cvk_context : public _cl_context,
...
             m_properties.push_back(*props);
         }
+        // Compute and save buffer size.
+        auto buff_size_prop_index =
+            get_property_index(CL_PRINTF_BUFFERSIZE_ARM);
+        if (buff_size_prop_index != -1) {
+            m_buffer_size = m_properties[buff_size_prop_index];
+        } else {
+            m_buffer_size = config.printf_buffer_size;
+        }
```
Rekt3421, line 51: technically I can default to the config.printf_buffer_size by default and get rid of this else.

src/queue.cpp
```diff
@@ -63,6 +63,13 @@ cvk_command_queue::cvk_command_queue(
 
     TRACE_CNT(batch_in_flight_counter, 0);
     TRACE_CNT(group_in_flight_counter, 0);
+    auto cb_index = m_context->get_property_index(CL_PRINTF_CALLBACK_ARM);
+    if (cb_index != -1) {
+        auto all_props = m_context->properties();
+        m_cb_func = (printf_callback_func*)all_props[cb_index];
+    } else {
```
Rekt3421, line 70: Can probably get rid of this else as well.

tests/api/testcl.hpp
```diff
@@ -213,6 +213,13 @@ class WithContext : public ::testing::Test {
         ASSERT_CL_SUCCESS(err);
     }
 
+    void SetUpWithContextProperties(const cl_context_properties* _props) {
```
rjodinchr, line 216: Should be called by `WithContext::SetUp`

rjodinchr: `m_context` should also be released if not null, before creating a new context.
Maybe `m_context` should be initialized to null in the first place.

tests/api/testcl.hpp
```diff
@@ -510,12 +517,19 @@ class WithCommandQueue : public WithContext {
...
         m_queue = queue.release();
     }
 
-    void SetUp() override { SetUpQueue(0); }
+    void SetUp() override {
+        WithContext::SetUp();
+        SetUpQueue(0);
+    }
+
+    void SetUpWithQueueProperties(const cl_context_properties* _prop) {
```
rjodinchr, line 529: `SetUpWithQueueProperties` indicates that `_prop` is a queue properties, but it is used for the context properties, it does not make sense.
The idea here is to have a `WithCommandQueue::WithSetUpWithQueueProperties` that can be used in `WithProfiledCommandQueue::SetUp` instead of have to call `WithCommandQueue::SetUpQueue` explicitely.

rjodinchr: We will also need a `WithCommandQueue::SetUpWithContextProperties`

tests/api/testcl.hpp
```diff
@@ -760,5 +774,18 @@ class WithCommandQueue : public WithContext {
...
+        SetUpQueue(CL_QUEUE_PROFILING_ENABLE);
+    }
+};
+
+class WithPrintfEnabled : public WithCommandQueue {
+protected:
+    void SetUp() override{};
+    void TearDown() override { WithCommandQueue::TearDown(); }
+
+    void SetupPrintfCallback(const cl_context_properties* _props) {
```
rjodinchr, line 788: Please rename `WithPrintfEnabled::SetUpWithContextProperties`.

tests/api/printf.cpp
```diff
@@ -34,6 +34,9 @@ static std::string stdoutFileName;
 #define BUFFER_SIZE 1024
 static char stdoutBuffer[BUFFER_SIZE];
 
+std::vector<char> GLOBAL_BUFFER;
```
rjodinchr, line 37: clvk does not have any global variables written in full uppercase anywhere. Please use a format already use somewhere else in the repo. I would say `globalBuffer`?

tests/api/printf.cpp
```diff
@@ -153,29 +172,29 @@ TEST_F(WithCommandQueue, TooLongPrintf) {
...
-
-    // We only get the first 2 prints because the buffer is too small to get the
-    // last one.
-    const char* message = "get_global_id(0) = 0\nget_global_id(1) = 0\n";
-    ASSERT_STREQ(printf_buffer, message);
+    // Reset the buffer if complete, otherwise keep the remaining part
+    ASSERT_STREQ(GLOBAL_BUFFER.data(), message);
+    BUFFER_FILL_LEVEL = 0;
+    GLOBAL_BUFFER.clear();
+    GLOBAL_BUFFER.shrink_to_fit();
```
rjodinchr, line 179: Should not be needed here. Everything could be done at the beginning. We can leave the buffer as it is at the end

tests/api/printf.cpp
```diff
@@ -128,17 +143,21 @@ TEST_F(WithCommandQueue, SimplePrintf) {
...
+    cl_context_properties properties[] = {
+        /* Enable a printf callback function for this context. */
+        CL_PRINTF_CALLBACK_ARM,
+        (cl_context_properties)printf_callback,
+        CL_PRINTF_BUFFERSIZE_ARM,
+        buff_size,
+
+    };
+    SetupPrintfCallback(properties);
+    GLOBAL_BUFFER.resize(buff_size, '\0');
```
rjodinchr, line 157: Let's have a function in this file to do all the setup, to avoid duplication

tests/api/printf.cpp
```diff
@@ -220,11 +244,43 @@ TEST_F(WithCommandQueue, PrintfMissingLengthModifier) {
...
-    auto printf_buffer = getStdoutContent();
-    ASSERT_NE(printf_buffer, nullptr);
+    ASSERT_STREQ(GLOBAL_BUFFER.data(), message);
+    BUFFER_FILL_LEVEL = 0;
+    GLOBAL_BUFFER.clear();
+    GLOBAL_BUFFER.shrink_to_fit();
+}
 
-    ASSERT_STREQ(printf_buffer, message);
+TEST_F(WithPrintfEnabled, PrintSimple) {
```
rjodinchr, line 253: I don't think we need to add this test.

tests/api/printf.cpp
```diff
@@ -34,6 +34,9 @@ static std::string stdoutFileName;
 #define BUFFER_SIZE 1024
 static char stdoutBuffer[BUFFER_SIZE];
 
+std::vector<char> GLOBAL_BUFFER;
+size_t BUFFER_FILL_LEVEL = 0;
```
rjodinchr, line 38: I think we could get rid of this global variable.

Let's do it like that:
- Resize the global buffer to `0` at the beginning of each test.
- in printf_callback resize it to be able to copy the content

rjodinchr (changes requested):

tests/api/testcl.hpp
Rekt3421: I do not quite understand why `WithContext::SetUp` should call this function, can you please explain?

rjodinchr: We just want to avoid duplicating the code.
```
void SetUp() override {
  SetUpWithContextProperties(nullptr);
}
```

src/context.hpp
```diff
@@ -41,6 +43,17 @@ struct cvk_context : public _cl_context,
...
+        auto buff_size_prop_index =
+            get_property_index(CL_PRINTF_BUFFERSIZE_ARM);
+        if (buff_size_prop_index != -1) {
+            m_buffer_size = m_properties[buff_size_prop_index];
+        }
+        // If buffersize is set in config then default to that.
+        if constexpr (std::is_member_object_pointer_v<
+                          decltype(&config_struct::printf_buffer_size)>) {
+            m_buffer_size = config.printf_buffer_size;
+        }
```
rjodinchr, line 56: When I was saying printf buffer size is set, I was referering to https://github.com/kpet/clvk/blob/main/src/config.hpp#L42

No need for `type_traits`. Just check for `config.printf_buffer_size.set`. In fact `config.printf_buffer_size` is a shortcut for `config.printf_buffer_size()` or `config.printf_buffer_size.value`.

And make sure to always set `m_buffer_size`. With the current implementation, it's not obvious that we would not end up with an uninitialized value.

src/printf.cpp
```diff
@@ -256,9 +264,8 @@ cl_int cvk_printf(cvk_mem* printf_buffer,
     const size_t bytes_written = read_inc_buff<uint32_t>(data) * 4;
     const size_t limit = std::min(bytes_written, data_size);
     auto* data_end = data + limit;
-
```
rjodinchr, line 259: no need to remove than line

src/program.cpp
```diff
@@ -15,7 +15,6 @@
 #include <algorithm>
 #include <cstring>
 #include <filesystem>
-#include <iostream>
```
rjodinchr, line 18: why did you remove that include?

tests/api/printf.cpp
```diff
@@ -17,7 +17,7 @@
 #include "testcl.hpp"
 #include "unit.hpp"
 #include "utils.hpp"
-
+#include <cstring>
```
rjodinchr, line 20: please keep an empty line between local and standard includes

tests/api/printf.cpp
```diff
@@ -104,6 +106,36 @@ static std::string getStdoutFileName(temp_folder_deletion& temp) {
...
+
+    if (complete || (space_available < len && space_available >= 1)) {
+        globalBuffer.emplace_back('\0');
+    }
+}
+
+std::vector<cl_context_properties>& setup_arm_printf_test(long int buff_size) {
+    globalBuffer.clear();
+    globalBuffer.shrink_to_fit();
+    globalBuffer.reserve(buff_size);
```
rjodinchr, line 129: Why do you want to reserve `buff_size`? This is just complicated things I feel. `globalBuffer` has nothing to do with the `cl_arm_printf` extension, it does not have to use a specific size reserve.

tests/api/printf.cpp
```diff
@@ -104,6 +106,36 @@ static std::string getStdoutFileName(temp_folder_deletion& temp) {
     return (prefix / suffix).string();
 }
 
+void printf_callback(const char* buffer, size_t len, size_t complete,
```
rjodinchr, line 109: This function is too complicated I feel.
Let's make `globalBuffer` a `std::string` it should make it much simplier.

tests/api/printf.cpp
```diff
@@ -190,26 +204,17 @@ TEST_F(WithCommandQueue, TooLongPrintf2) {
...
 
-TEST_F(WithCommandQueue, PrintfMissingLengthModifier) {
-    temp_folder_deletion temp;
-    stdoutFileName = getStdoutFileName(temp);
-
-    int fd;
-    ASSERT_TRUE(getStdout(fd));
-
-    const char message[] = "1,2,3,4";
+TEST_F(WithPrintfEnabled, PrintfMissingLengthModifier) {
```
rjodinchr, line 211: Why did you modify this test? could it stay with `WithCommandQueue`?

tests/api/printf.cpp
```diff
@@ -153,29 +180,16 @@ TEST_F(WithCommandQueue, TooLongPrintf) {
...
-    auto cfg1 =
-        CLVK_CONFIG_SCOPED_OVERRIDE(printf_buffer_size, uint32_t, 36, true);
-
-    temp_folder_deletion temp;
-    stdoutFileName = getStdoutFileName(temp);
-
-    int fd;
-    ASSERT_TRUE(getStdout(fd));
+TEST_F(WithPrintfEnabled, TooLongPrintf2) {
+    auto props = setup_arm_printf_test(46);
```
rjodinchr, line 188: Why `46`? it used to be `36` with a comment explaining why, but now it's another value that we don't know where it comes from.

tests/api/printf.cpp
```diff
@@ -128,17 +160,12 @@ TEST_F(WithCommandQueue, SimplePrintf) {
...
-    auto cfg1 =
-        CLVK_CONFIG_SCOPED_OVERRIDE(printf_buffer_size, uint32_t, 28, true);
-
-    temp_folder_deletion temp;
-    stdoutFileName = getStdoutFileName(temp);
-
-    int fd;
-    ASSERT_TRUE(getStdout(fd));
+TEST_F(WithPrintfEnabled, TooLongPrintf) {
+    auto props = setup_arm_printf_test(44);
```
rjodinchr, line 164: Why `44`? it used to be `28` with a comment explaining why, but now it's another value that we don't know where it comes from.

tests/api/testcl.hpp
```diff
@@ -510,12 +516,24 @@ class WithCommandQueue : public WithContext {
...
-        WithContext::SetUp();
         auto queue = CreateCommandQueue(device(), properties);
         m_queue = queue.release();
     }
 
-    void SetUp() override { SetUpQueue(0); }
+    void SetUp() override {
+        WithContext::SetUp();
+        SetUpQueue(0);
+    }
```
rjodinchr, line 526:
```suggestion
    void SetUp() override {
        SetUpWithContextProperties(nullptr);
    }
```

tests/api/printf.cpp
```diff
@@ -128,17 +160,12 @@ TEST_F(WithCommandQueue, SimplePrintf) {
...
+    SetUpWithContextProperties(props.data());
```
rjodinchr, line 165: `SetUpWithContextProperties` is always called after `setup_arm_printf_test`. We should move it inside to avoid duplication.
Also we need to call `WithPrintfEnabled::TearDown()` before calling `SetUpWithContextProperties` to make sure that everything has been deleted before creating the new context. You can check that part using valgrind, to make sure there is no memory leak.

rjodinchr (changes requested):

src/program.cpp
Rekt3421: because I think i put this in accidentally in a past cl but lets leave it there.

tests/api/printf.cpp
Rekt3421: I tried that previously (to move the `SetUpWithContextProperties` inside the func(I am assuming that is what you mean) but that was failing as it does not have the scope for the `WithPrintfEnabled`

rjodinchr:
@kpet Could you re-generate the swiftshader artifact please?

kpet:
@rjodinchr I've kicked off the job (https://github.com/kpet/clvk/actions/runs/10080985898). Can you re-trigger the CI once it completes or will I need to do it too?

rjodinchr:
> @rjodinchr I've kicked off the job (https://github.com/kpet/clvk/actions/runs/10080985898). Can you re-trigger the CI once it completes or will I need to do it too?

We do not have the ability to re-trigger the CI, but @Rekt3421 could just push some kind of empty commit to force a trigger...

Rekt3421:
thanks Kevin I will do an empty push once the job completes

src/context.hpp
```diff
@@ -73,11 +105,28 @@ struct cvk_context : public _cl_context,
...
+    int get_property_index(const int prop) {
+        for (unsigned i = 0; i < m_properties.size(); i += 2) {
+            if (m_properties[i] == prop) {
+                return i + 1;
+            }
+        }
+        return -1;
+    }
+
+    int get_prop_size() { return m_properties.size(); }
```
rjodinchr, line 117: Not used, please remove

rjodinchr (approved):
I think you can mark this PR ready for review

src/context.hpp
```diff
@@ -73,11 +99,27 @@ struct cvk_context : public _cl_context,
...
+    void* get_printf_userdata() { return m_printf_userdata; }
+
 private:
     cvk_device* m_device;
     std::mutex m_callbacks_lock;
     std::vector<cvk_context_callback> m_destuctor_callbacks;
     std::vector<cl_context_properties> m_properties;
+    size_t m_printf_buffersize;
+    cvk_printf_callback_t m_printf_callback;
+    void* m_printf_userdata;
```
rjodinchr, line 122:
```suggestion
    void* m_userdata;
```
And change all the places where it is used.

src/unit.hpp
```diff
@@ -14,6 +14,8 @@
 
 #pragma once
 
+#define CLVK_PRINTF_USERDATA 0x5100
```
rjodinchr, line 17: Not needed anymore

rjodinchr (approved):

src/context.hpp
```diff
@@ -73,11 +99,27 @@ struct cvk_context : public _cl_context,
         return size <= m_device->max_mem_alloc_size();
     }
 
+    int get_property_index(const int prop) {
```
kpet, line 102:
```suggestion
    int get_property_value_index(const int prop) {
```

src/context.hpp
```diff
@@ -28,7 +33,8 @@ struct cvk_context : public _cl_context,
                      refcounted,
                      object_magic_header<object_magic::context> {
 
-    cvk_context(cvk_device* device, const cl_context_properties* props)
+    cvk_context(cvk_device* device, const cl_context_properties* props,
+                void* user_data)
```
kpet, line 37: Suggest initialising `m_user_data` here. I can't think of a reason to gate the behaviour on the presence of the printf properties.

tests/api/printf.cpp
```diff
@@ -128,17 +134,19 @@ TEST_F(WithCommandQueue, SimplePrintf) {
...
-
-    temp_folder_deletion temp;
-    stdoutFileName = getStdoutFileName(temp);
-
-    int fd;
-    ASSERT_TRUE(getStdout(fd));
+    cl_context_properties properties[4] = {
+        CL_PRINTF_CALLBACK_ARM, (cl_context_properties)printf_callback,
+        CL_PRINTF_BUFFERSIZE_ARM, (cl_context_properties)28};
+    WithCommandQueue::SetUpWithContextProperties(
```
kpet, line 144: Doing the set up from within the test is a bit weird. I'm not sure that's recommended practise with Googletest. We could have a dedicated fixture where printf is always configured or maybe even do this in the default one.

kpet (approved):
This has been open for too long. Merging as is but will probably land a few tweaks on top.

tests/api/printf.cpp
rjodinchr: It can easily be done for the callback, but the complex part is that the test needs to define the buffer size.

PR #676, opened 2024-03-13
fix WithCommandQueue.EnqueueTooManyCommands

rjodinchr:
When using hardware that override max_first_cmd_group/batch_size in device_properties, this test is not passing.

As those properties are set much before the scope of the test and are not reevaluate later, we cannot override them in the test.

Fix the test by adding a not config to prevent early flush when testing.


kpet (approved):
Thanks!

PR #679, opened 2024-03-17
Tidy up and document support for config files

kpet:
Change-Id: I359515fb1decf5cced112587c4d4c58e46414448

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -237,6 +237,8 @@ void clvk_global_state::init_executors() {
 void clvk_global_state::term_executors() { delete m_thread_pool; }
 
 clvk_global_state::clvk_global_state() {
+    // Init the configuration using environment variables before logging so
+    // we can enable full logging support before parsing configuration files.
     init_config_from_env_only();
     init_logging();
     init_config();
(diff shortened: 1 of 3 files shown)


src/init.cpp
```diff
@@ -237,9 +237,8 @@ void clvk_global_state::init_executors() {
 void clvk_global_state::term_executors() { delete m_thread_pool; }
 
 clvk_global_state::clvk_global_state() {
-    init_config_from_env_only();
-    init_logging();
     init_config();
+    init_logging();
```
rjodinchr, line 241: I understand that parsing the env twice is a workaround more than a clean implementation.

The goal is to be able to have logs when running `parse_config_file`. So we need the logging to be initialized before it. And to have any logs, we need to parse the env.
Then we need to parse it one more time to override things that could have been set by `parse_config_file`.

kpet: Makes sense, thanks. I did bump into this when I first implemented config handling and solved it by not using logging at all during config parsing. I'll revert my changes and add a comment for now.

rjodinchr (approved):

PR #680, opened 2024-03-18
fix clGetMemObjectInfo of CL_MEM_HOST_PTR after fix in CTS

rjodinchr:
https://github.com/KhronosGroup/OpenCL-CTS/pull/1850

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -2068,7 +2068,10 @@ cl_int CLVK_API_CALL clGetMemObjectInfo(cl_mem mem, cl_mem_info param_name,
         ret_size = sizeof(val_sizet);
         break;
     case CL_MEM_HOST_PTR:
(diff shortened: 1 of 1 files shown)


src/api.cpp
```diff
@@ -2068,7 +2068,10 @@ cl_int CLVK_API_CALL clGetMemObjectInfo(cl_mem mem, cl_mem_info param_name,
         ret_size = sizeof(val_sizet);
         break;
     case CL_MEM_HOST_PTR:
-        val_ptr = memobj->host_ptr();
+        val_ptr = nullptr;
+        if (memobj->flags() & CL_MEM_USE_HOST_PTR) {
```
kpet, line 2072:
```suggestion
        if (memobj->has_any_flag(CL_MEM_USE_HOST_PTR)) {
```

kpet (approved):

PR #681, opened 2024-03-19
update nvidia device properties (remove fabs)

rjodinchr:
update of llvm & builtins libraries in clspv change the implementation of nextafter and remainder that now requires fabs to use the "emulated" implementation.

https://github.com/google/clspv/pull/1326


rjodinchr:
The test failing is a flake. Here is a fix for it: https://github.com/kpet/clvk/pull/682

kpet (approved):

PR #685, opened 2024-03-25
fix end_current_command_batch following enqueue_retry implementation

rjodinchr:
Since we are able to fail an enqueue to retry later on, we have a scenario where `m_command_buffer` has been allocated during `add_command` (in `cvk_command_batch`) but the `build` has failed. Then, in `enqueue_command_with_retry` we can end up calling `end_current_command_batch` which would try to enqueue an empty command buffer.

This scenario leads to different issues later on in the flow with the command buffer trying to be executed while not having any command in it.

### DIFF
--- a/src/queue.cpp
+++ b/src/queue.cpp
@@ -279,7 +279,7 @@ cl_int cvk_command_queue::enqueue_command_with_deps(
 }
 
 cl_int cvk_command_queue::end_current_command_batch() {
-    if (m_command_batch) {
+    if (m_command_batch && m_command_batch->batch_size() > 0) {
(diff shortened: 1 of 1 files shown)


rjodinchr:
I think it is not possible to have a test for that right now.

My initial issue is coming from this [line](https://chromium.googlesource.com/chromiumos/overlays/chromiumos-overlay/+/main/media-libs/clvk/files/clvk-90-timeline-semaphores.patch#713) in the timeline semaphore patch.

It is needed to chain events that are in the same timeline (NDRange having a different timeline than the batch object).

kpet (approved):
Thanks for having a look! Let's proceed optimistically then as discussed. Worst case we'll revert the change or add a fix on top.

PR #688, opened 2024-04-05
Avoid creating a thread to build program when not needed

rjodinchr:
If no callback is given, do not build in a thread. The goal is to make perfetto more readable by reducing the number of thread used only for a very small instant.
It may also reduce build time, but honestly this is less than 1%.

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -2303,21 +2303,8 @@ clBuildProgram(cl_program prog, cl_uint num_devices,
     // clCreateProgramWithSource or clCreateProgramWithBinary or
     // clCreateProgramWithILKHR.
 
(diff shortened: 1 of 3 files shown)


kpet (approved):
Not using a separate thread for synchronous compiler operations is a good change. I'm fine with merging this change as is but I'm not a big fan of having to compare the build status against `CL_BUILD_NONE` to decide to return `CL_INVALID_OPERATION`. This breaks encapsulation. I think we can make error handling much simpler if we change `cvk_program::build` to return a `cl_int`. In the case of build failures you'd just need to look at the `build_operation` to pick `CL_BUILD_PROGRAM_FAILURE` or `CL_LINK_PROGRAM_FAILURE`.

kpet (approved):
Nice, thanks!

PR #690, opened 2024-04-19
change file config order

rjodinchr:
It makes more sense to start from the most global location, which means starting with `/etc/clvk.conf`.


kpet (approved):
Thanks for making the change!

PR #696, opened 2024-05-06
add option to force to build in a separate thread

rjodinchr:
Also name the build thread only if built in a separate thread.

### DIFF
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -1515,8 +1515,6 @@ bool cvk_program::check_capabilities(const cvk_device* device) const {
 }
 
 void cvk_program::do_build() {
-    cvk_set_current_thread_name_if_supported("clvk-build");
-
     // Destroy entry points from previous build
     m_entry_points.clear();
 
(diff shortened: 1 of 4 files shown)


rjodinchr:
I'm experiencing issues with clang starting to segfault in the frontend. Building in a separate thread fixes the issue. It feels like a stack corruption, but I did not figure out exactly where it is coming from.

I am not able to have a simple reproducer that I can share yet.

rjodinchr:
This is not needed anymore in ChromeOS. It seems that clang has been fixed.
I would be in favour of still merging it to be able to easily switch in case anyone experiences that issue, but I'm also fine just closing it.

src/program.cpp
```diff
@@ -1682,12 +1680,24 @@ cl_int cvk_program::build(build_operation operation, cl_uint num_devices,
     m_operation_callback_data = data;
 
     cl_int ret = CL_SUCCESS;
-    if (cb) {
+    bool build_in_separate_thread = config.build_in_separate_thread() || cb;
+    bool check_build_status = !cb;
```
kpet, line 1684:
```suggestion
    bool wait_for_completion = !cb;
```
Feels a bit clearer maybe.

kpet (approved):
Let's get it in as a debugging tool.

kpet (approved):

PR #697, opened 2024-05-20
tests/api/images: force unmap/release to be perform after check

rjodinchr:
Depending on values in device_properties, some platform can perform the unmap when the function is called, not when the app calls clFinish.
Force this behavior by adding a userevent to control when the unmap can be performed.

It might also fix flakyness in github CI. Not too sure.

### DIFF
--- a/tests/api/images.cpp
+++ b/tests/api/images.cpp
@@ -727,7 +727,12 @@ TEST_F(WithCommandQueue, 1DBufferImageUnmapAfterRelease) {
         buffer,                       // buffer
     };
 
+    // clvk controller can decide to submit command directly after being
+    // enqueued. As we need to make sure that the unmap command will not be
+    // executed before we call Finish, we add this event to garantee this
+    // property.
     auto event = CreateUserEvent();
+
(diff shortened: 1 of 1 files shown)


tests/api/images.cpp
```diff
@@ -727,7 +727,12 @@ TEST_F(WithCommandQueue, 1DBufferImageUnmapAfterRelease) {
         buffer,                       // buffer
     };
 
+    // clvk controller can decide to submit command directly after being
+    // enqueued. As we need to make sure that the unmap command will not be
+    // executed before we call Finish, we add this event to garantee this
```
kpet, line 732:
```suggestion
    // executed before we call Finish, we add this event to guarantee this
```

tests/api/images.cpp
```diff
@@ -771,7 +776,12 @@ TEST_F(WithCommandQueue, 1DBufferImageReleaseAfterUnmap) {
...
```
kpet, line 781:
```suggestion
    // executed before we call Finish, we add this event to guarantee this
```

kpet (approved):

PR #703, opened 2024-06-12
add config option to expose device extensions

rjodinchr:
The goal is to be able to enable extensions that are not fully supported by clvk, but enough for some applications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -500,6 +500,9 @@ using the name of the corresponding environment variable.
   applications to work with the downside of not cleanly terminating clvk. Use
   with caution.
 
+* `CLVK_DEVICE_EXTENSIONS` specifies extensions to be added to the list of
+  exposed extensions. It expects a whitespace separated list of extensions.
+
 # Limitations
 
 * Only one device per CL context
--- a/src/device.cpp
+++ b/src/device.cpp
@@ -674,6 +674,26 @@ void cvk_device::build_extension_ils_list() {
         m_has_subgroup_size_selection = true;
     }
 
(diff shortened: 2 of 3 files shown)


src/device.cpp
```diff
@@ -679,6 +679,7 @@ void cvk_device::build_extension_ils_list() {
         m_extension_string += ext.name;
         m_extension_string += " ";
     }
+    m_extension_string += config.device_extensions();
```
kpet, line 682: As discussed, this approach will not work with applications that use `CL_DEVICE_EXTENSIONS_WITH_VERSION`. I suggest you add the extensions to the `m_extension` vector so both mechanisms are supported. I think it's acceptable to start by just hardcoding the version to 0.0.0.

kpet (changes requested):

README.md
```diff
@@ -500,6 +500,10 @@ using the name of the corresponding environment variable.
   applications to work with the downside of not cleanly terminating clvk. Use
   with caution.
 
+* `CLVK_DEVICE_EXTENSIONS` specifies extensions to be added to the list of
+  exposed extensions by `clGetDeviceInfo` in `CL_DEVICE_EXTENSIONS`. It
```
kpet, line 504:
```suggestion
  exposed extensions. It
```
This now no longer reflects what the change does. Maybe keeping it abstract is easier?

kpet (changes requested):
The doc needs a tweak, looks good otherwise.

kpet (approved):

PR #707, opened 2024-06-27
Pr/update deps

rjodinchr:


rjodinchr:
@kpet 🎉

kpet (approved):
Thanks for doing this! That's a great step forward. We'll have to figure out how we update the LLVM archive used in the CI but let's put a stake in the ground.

Issue #710, opened 2024-07-31
Add checks to avoid running out of order execution queues


Rekt3421:
at the moment three conformance test issues are failing as clvk simply ignore out of order execution (as it is not supported) the tests are negative tests and are meant to fail the tests are :
```
negative_create_command_queue
negative_create_command_queue_with_properties
negative_create_command_queue_with_properties_khr
```

kpet:
Now fixed by #713

PR #713, opened 2024-08-12
Give appropriate errors for out of order execution related props

Rekt3421:
this addresses issues for the failing tests in #710 for the following tests
```
negative_create_command_queue
negative_create_command_queue_with_properties
negative_create_command_queue_with_properties_khr
```


src/api.cpp
```diff
@@ -1466,11 +1476,26 @@ cvk_create_command_queue(cl_context context, cl_device_id device,
         return nullptr;
     }
 
+    cl_int err = CL_SUCCESS;
+    std::string oooe_env_var = "CLVK_IGNORE_OUT_OF_ORDER_EXECUTION";
```
rjodinchr, line 1480: Please use clvk configuration (`<clvk>/src/config.def`). This should be a boolean there. It will allow to use the env var as well as the configuration file.

rjodinchr (changes requested):

src/config.def
```diff
@@ -50,6 +50,7 @@ OPTION(uint32_t, max_cmd_batch_size, 10000u)
 OPTION(uint32_t, max_first_cmd_batch_size, 10000u)
 OPTION(uint32_t, max_cmd_group_size, UINT32_MAX)
 OPTION(uint32_t, max_first_cmd_group_size, UINT32_MAX)
+OPTION(uint32_t, ignore_out_of_order_execution, 0u) // 0 meaning dont ignore
```
rjodinchr, line 53: Make it a `bool` please

Rekt3421: done

README.md
```diff
@@ -345,6 +345,11 @@ using the name of the corresponding environment variable.
 
 * `CLVK_CONFIG_FILE` specifies the path to an additional configuration file.
 
+* `CLVK_IGNORE_OUT_OF_ORDER_EXECUTION` can be used to ignore out of order execution checks
+
+   * 0: does not ignore OOOE checks and validates if hardware can support it
```
rjodinchr, line 350:
```suggestion
   * 0: does not ignore OOOE checks and validates if hardware can support it (default)
```

src/api.cpp
```diff
@@ -1466,11 +1476,24 @@ cvk_create_command_queue(cl_context context, cl_device_id device,
         return nullptr;
     }
 
+    cl_int err = CL_SUCCESS;
+    if (!config.ignore_out_of_order_execution.set ||
+        config.ignore_out_of_order_execution.value == 0) {
```
rjodinchr, line 1481:
```suggestion
    if (!(config.ignore_out_of_order_execution.set &&
        config.ignore_out_of_order_execution())) {
```

src/config.def
```diff
@@ -50,6 +50,7 @@ OPTION(uint32_t, max_cmd_batch_size, 10000u)
 OPTION(uint32_t, max_first_cmd_batch_size, 10000u)
 OPTION(uint32_t, max_cmd_group_size, UINT32_MAX)
 OPTION(uint32_t, max_first_cmd_group_size, UINT32_MAX)
+OPTION(uint32_t, ignore_out_of_order_execution, false) // false meaning dont ignore
```
rjodinchr, line 53:
```suggestion
OPTION(bool, ignore_out_of_order_execution, false) // false meaning dont ignore
```

Rekt3421: ah I forgot to sync my local branch with the bool change. Thanks!

rjodinchr (approved):

src/api.cpp
```diff
@@ -163,6 +163,16 @@ struct api_query_string : public std::string {
     size_t size_with_null() const { return size() + 1; }
 };
 
+bool out_of_order_device_support(const cl_device_id devices, cl_int err) {
```
rjodinchr, line 166: What's the idea with `cl_int err` here? If the goal is to be able to have the caller being able to have the value, it needs to be a reference.

rjodinchr (changes requested):

src/api.cpp
```diff
@@ -1467,6 +1467,14 @@ cvk_create_command_queue(cl_context context, cl_device_id device,
         return nullptr;
     }
 
+    if (!(config.ignore_out_of_order_execution.set &&
```
rjodinchr, line 1470: Why do we want to check whether it is set? I feel like just checking the value should be enough, but I might be missing something here.

rjodinchr (approved):

src/api.cpp
```diff
@@ -1467,6 +1467,15 @@ cvk_create_command_queue(cl_context context, cl_device_id device,
...
 
+    if (config.ignore_out_of_order_execution()) {
+        {
+            // We do not support out of order command queues so this must fail
+            if (properties & CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE) {
+                *errcode_ret = CL_INVALID_QUEUE_PROPERTIES;
+                return nullptr;
+            }
+        }
+    }
```
rjodinchr, line 1478:
```suggestion
    if (config.ignore_out_of_order_execution()) {
          // We do not support out of order command queues so this must fail
          if (properties & CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE) {
              *errcode_ret = CL_INVALID_QUEUE_PROPERTIES;
              return nullptr;
          }
    }
```

rjodinchr (changes requested):

rjodinchr (approved):

rjodinchr (approved):

README.md
```diff
@@ -369,6 +369,11 @@ using the name of the corresponding environment variable.
 
 * `CLVK_CONFIG_FILE` specifies the path to an additional configuration file.
 
+* `CLVK_IGNORE_OUT_OF_ORDER_EXECUTION` can be used to ignore out of order execution checks
+
+   * 0: does not ignore OOOE checks and validates if hardware can support it (default)
+   * 1: ignores the aforementioned checks
```
kpet, line 375:
```suggestion
   * 0: creating an out-of-order queue results in a failure (default)
   * 1: creating an out-of-order queue is supported but it will function as an in-order queue
```

kpet (approved):

Issue #714, opened 2024-08-14
Fix issues with printf tests.


Rekt3421:
Right now the majority of conformance tests for printf related to strings are failing, there are about 11 tests and 10 fail. It mostly has to do with how we process the strings in printf.cpp and also how kernel descriptors are setup here is an example of issues related to printf.cpp

```
1)testing printf("%s","")
terminate called after throwing an instance of 'std::out_of_range'
```

and here is an example of issues to due to kernel descriptors not setup properly 
```
12)testing printf("%s","foo\0foo")
No data read from analysis buffer
verifyOutputBuffer failed with kernel: 
__kernel void test9(void)
{
   printf("%s\n","foo\0foo");
}

 expected: foo
 got:     
```

rjodinchr:
This issue can be closed

PR #715, opened 2024-08-14
Address short comings with printf for strings

Rekt3421:
This change fixes short comings encountered in 9 tests of the OpenCL-CTS as mentioned in issue #714 we still need to fix kernel descriptors.


Rekt3421:
Next step is to add some unit tests

src/printf.cpp
```diff
@@ -166,77 +166,121 @@ void process_printf(char*& data, const printf_descriptor_map_t& descs,
...
-            // should be printed verbatim
-            printf_out << part_fmt;
-            break;
+    if (next_part > format_string.size()) {
+        // if there is no format string followed by a string to print just print
+        // the whole string itself.
+        if (descs.size() == 1) {
+            printf_out << format_string;
         }
+    } else {
```
rjodinchr, line 175:
```suggestion
    if (next_part > format_string.size()) {
        // if there is no format just print the whole string itself.
        assert (descs.size() == 1);
        printf_out << format_string;
        printf("%s", printf_out.str().c_str());
        return;
    }
```

src/printf.cpp
```diff
@@ -167,21 +167,54 @@ void process_printf(char*& data, const printf_descriptor_map_t& descs,
...
+    }
     printf_out << format_string.substr(0, next_part);
 
     // Decompose the remaining format string into individual strings with
     // one format specifier each, handle each one individually
     size_t arg_idx = 0;
     while (next_part < format_string.size() - 1) {
-        // Get the part of the format string before the next format specifier
+        // Get the part of the format string before the next format
+        // specifier
```
rjodinchr, line 183:
```suggestion
        // Get the part of the format string before the next format specifier
```

src/printf.cpp
```diff
@@ -210,19 +243,24 @@ void process_printf(char*& data, const printf_descriptor_map_t& descs,
...
             // Vector argument
             if (el_size == 0) {
-                // 'ele_size == 0' means that no length modifier has been used.
-                // According to the spec, this is an undefined behavior. Let's
-                // use the size coming from clspv and the vec_len to figure out
-                // the element size then.
+                // 'ele_size == 0' means that no length modifier has been
+                // used. According to the spec, this is an undefined
+                // behavior. Let's use the size coming from clspv and the
+                // vec_len to figure out the element size then.
```
rjodinchr, line 263:
```suggestion
                // 'ele_size == 0' means that no length modifier has been used.
                // According to the spec, this is an undefined behavior. Let's
                // use the size coming from clspv and the vec_len to figure out
                // the element size then.
```

rjodinchr (approved):
Make sure you only modify what really needs to change.

tests/api/printf.cpp
```diff
@@ -32,6 +32,20 @@ TEST_F(WithCommandQueueAndPrintf, SimplePrintf) {
     ASSERT_STREQ(m_printf_output.c_str(), message);
 }
 
+TEST_F(WithCommandQueueAndPrintf, SimpleFormatedPrintf) {
+    const char message[] = "";
+    char source[512];
+    sprintf(source, "kernel void test_printf() { printf(\"%%s\", \"\"); }");
```
rjodinchr, line 38: I don't get the point of using `sprintf` if they are no arguments after the format.

rjodinchr (approved):

rjodinchr:
This is fixing CTS `printf string`, but `printf format_string` is regressing (segfault with `"\"%%\""`).

We should maybe try to address everything (`string` & `format_string`) together.

Rekt3421:
current status 3 sub-subtests are still failing. One which is due to some issue with printf itself. 

I.e.  in strings subtest there is a subtest
```
printf(%%foo%%) 
```
is printing out 
```
%%foo%
```
which is a bit confusing since the two %% should be consumed as one. 

also in the same subtests for 
```
   printf("%s\n","foo\0foo");                                                           
``` 
the kernel descriptors are not properly populated for some reason.

a vector sub-subtest is also failing 
for which I do not have an explanation yet.

src/printf.cpp
```diff
@@ -167,21 +184,66 @@ void process_printf(char*& data, const printf_descriptor_map_t& descs,
...
+    }
     printf_out << format_string.substr(0, next_part);
 
     // Decompose the remaining format string into individual strings with
     // one format specifier each, handle each one individually
     size_t arg_idx = 0;
+    std::stack<char> quotes;
+    if (format_string.size() > 0) {
+        auto curr_quote = get_quote(format_string.substr(
+            0, next_part)); // checking if curr selection is inside a " or '.
```
Rekt3421, line 198: Perhaps I should change this comment to "checking if curr selection has a " or '

src/printf.cpp
```diff
@@ -13,12 +13,19 @@
...
+#include <stack>
 
 #include "printf.hpp"
 
 // Extract the conversion specifier from a format string
 char get_fmt_conversion(std::string_view fmt) {
     auto conversionSpecPos = fmt.find_first_of("diouxXfFeEgGaAcsp");
+    // Check if a valid conversion specifier was found
+    if (conversionSpecPos == std::string_view::npos) {
+        conversionSpecPos = -1;
```
rjodinchr, line 25: Why setting `conversionSpecPos` to `-1`?

src/printf.cpp
```diff
@@ -242,8 +318,8 @@ void process_printf(char*& data, const printf_descriptor_map_t& descs,
 
     auto output = printf_out.str();
     if (printf_cb != nullptr) {
-        auto len = output.size();
-        printf_cb(output.c_str(), len, data >= data_end, printf_userdata);
+        printf_cb(output.c_str(), output.size(), data >= data_end,
+                  printf_userdata);
```
rjodinchr, line 322: What does it fix?

Rekt3421: nothing I just though it was more compact. Ill put it back to the way it was.

src/printf.cpp
Rekt3421: I was previously passing this as a ref to also get the position of the specifier, realized I don't need to, thanks for point this out;

rjodinchr (changes requested):

src/printf.cpp
```diff
@@ -165,14 +165,20 @@ void process_printf(char*& data, const printf_descriptor_map_t& descs,
...
 
-    // Firstly print the part of the format string up to the first '%'
+    // Firstly print the part of the format string up to the first '%' if any
+    // otherwise print the whole string as is and move the data pointer to the
+    // end.
     size_t next_part = format_string.find_first_of('%');
-    printf_out << format_string.substr(0, next_part);
-
+    if (next_part == std::string::npos) {
+        printf_out << format_string;
```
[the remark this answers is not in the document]
Rekt3421: this is what I get on my end if I do not have this change 

```
0)testing printf("","foo")
verifyOutputBuffer failed with kernel: 
__kernel void test9(void)
{
   printf("\n","foo");
}

 expected: 
 got:      
fo
```

src/printf.cpp
```diff
@@ -210,8 +226,11 @@ void process_printf(char*& data, const printf_descriptor_map_t& descs,
             // Special case for %s
             if (get_fmt_conversion(part_fmt) == 's') {
                 uint32_t string_id = read_buff<uint32_t>(data);
-                printf_out << print_part(
-                    part_fmt, descs.at(string_id).format_string.c_str(), size);
+                if (string_id < descs.size()) {
```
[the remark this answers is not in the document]
Rekt3421: if I remove this, I get the following for string test_printf
```
1)testing printf("%s","")
terminate called after throwing an instance of 'std::out_of_range'
  what():  unordered_map::at
Aborted
```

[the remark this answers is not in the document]
Rekt3421: okay I have removed it. Still seeing this failure but probably due to some issue on my end.

src/printf.cpp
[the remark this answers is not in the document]
Rekt3421: done. thanks

Rekt3421:
@kpet to get this PR working correctly we need to update clspv with the fixes landed by David

src/printf.cpp
```diff
@@ -181,7 +187,17 @@ void process_printf(char*& data, const printf_descriptor_map_t& descs,
         // Handle special cases
         if (part_end == part_start + 1) {
             printf_out << "%";
-            next_part = part_end + 1;
+            size_t next_format = format_string.find_first_of('%', part_end + 1);
```
rjodinchr, line 190: Could you revert that part to what I propose?

rjodinchr (approved):

kpet (approved):

PR #717, opened 2024-08-28
Tweak implementation of cl_arm_printf and introduce dedicated test fixture

kpet:
Change-Id: Ib27b60fb043e47227f519f34e04316d04f47b55d

### DIFF
--- a/src/context.hpp
+++ b/src/context.hpp
@@ -35,7 +35,7 @@ struct cvk_context : public _cl_context,
 
     cvk_context(cvk_device* device, const cl_context_properties* props,
                 void* user_data)
-        : m_device(device) {
+        : m_device(device), m_user_data(user_data) {
 
         if (props) {
             while (*props) {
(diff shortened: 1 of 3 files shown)


src/context.hpp
```diff
@@ -49,23 +49,21 @@ struct cvk_context : public _cl_context,
         }
         // Get printf buffer size from extension.
         auto buff_size_prop_index =
-            get_property_index(CL_PRINTF_BUFFERSIZE_ARM);
+            get_property_value_index(CL_PRINTF_BUFFERSIZE_ARM);
         if (buff_size_prop_index != -1 && !config.printf_buffer_size.set) {
             m_printf_buffersize = m_properties[buff_size_prop_index];
         } else {
-            m_printf_buffersize = config.printf_buffer_size;
+            m_printf_buffersize = 0;
```
rjodinchr, line 56: I think it `m_printf_buffersize` really needs to be initialize with `config.printf_buffer_size()`.

kpet: I changed the getter so the buffer size is taken from the config if it wasn't provided via the extension. This works for test overrides. It adds a bit of complexity but I think it's worth it for the benefits on the test side.

rjodinchr (approved):

PR #720, opened 2024-09-02
init image at creation time

### LINKED ISSUE #719: Image init data buffer
At the moment clvk defers the initialisation of image content given at creation (either by `CL_MEM_COPY_HOST_PTR` or `CL_MEM_USE_HOST_PTR`).

It leads to a staging buffer being allocated until the data is actually copied to the image memory.

I have an application where lots of images with init data are created and it ends up hitting an out-of-memory.

Could we have a mode where the copy of the data to the image memory is done at creation time, not at first use?

rjodinchr:
fix #719

### DIFF
--- a/src/context.cpp
+++ b/src/context.cpp
@@ -1,0 +1,34 @@
+// Copyright 2024 The clvk authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "context.hpp"
+#include "queue.hpp"
+
+cvk_command_queue* cvk_context::get_or_create_image_init_command_queue() {
+    std::unique_lock<std::mutex> lock(m_queue_image_init_lock);
+    if (m_queue_image_init != nullptr) {
+        return m_queue_image_init;
+    }
+    std::vector<cl_queue_properties> properties_array;
+    m_queue_image_init =
+        new cvk_command_queue(this, m_device, 0, std::move(properties_array));
+    cl_int ret = m_queue_image_init->init();
+    if (ret != CL_SUCCESS) {
+        return nullptr;
+    }
+    m_queue_image_init->detach_from_context();
+    return m_queue_image_init;
+}
+
+void cvk_context::free_image_init_command_queue() { delete m_queue_image_init; }
--- a/src/context.hpp
+++ b/src/context.hpp
@@ -29,6 +29,8 @@ struct cvk_context_callback {
     void* data;
 };
 
+struct cvk_command_queue;
+
 struct cvk_context : public _cl_context,
                      refcounted,
                      object_magic_header<object_magic::context> {
@@ -73,6 +75,7 @@ struct cvk_context : public _cl_context,
             auto cb = *cbi;
             cb.pointer(this, cb.data);
         }
+        free_image_init_command_queue();
     }
 
     const std::vector<cl_context_properties>& properties() const {
@@ -116,6 +119,9 @@ struct cvk_context : public _cl_context,
     cvk_printf_callback_t get_printf_callback() { return m_printf_callback; }
     void* get_printf_userdata() { return m_user_data; }
 
+    cvk_command_queue* get_or_create_image_init_command_queue();
+    void free_image_init_command_queue();
+
 private:
     cvk_device* m_device;
     std::mutex m_callbacks_lock;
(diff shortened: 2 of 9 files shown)


src/queue.cpp
```diff
@@ -101,6 +101,10 @@ cl_int cvk_command_queue::satisfy_data_dependencies(cvk_command* cmd) {
             continue;
         }
         CVK_ASSERT(mem->is_image_type());
+        if (config.init_image_at_creation()) {
+            tracker.set_state(cvk_mem_init_state::completed);
```
kpet, line 105: Shouldn't this be done in `cvc_image::init_vulkan_image`, after completing the execution of the init command?

src/context.cpp
```diff
@@ -0,0 +1,32 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "context.hpp"
+#include "queue.hpp"
+
+cvk_command_queue* cvk_context::get_or_create_image_init_command_queue() {
```
kpet, line 18: Looks like you need some kind of locking or protections. AFACIT after a quick read, it seems two threads could be creating this queue at the same time.

src/context.hpp
```diff
@@ -120,6 +126,8 @@ struct cvk_context : public _cl_context,
     size_t m_printf_buffersize;
     cvk_printf_callback_t m_printf_callback;
     void* m_user_data;
+
+    cvk_command_queue* m_queue_image_init = nullptr;
```
kpet, line 130: `std:unique_ptr`? You wouldn't need the free function.

kpet (changes requested):
Thanks! Fine with this being an option. Please add a unit test.

src/context.cpp
rjodinchr: I agree, I've added a lock

src/queue.cpp
rjodinchr: Yes, that's better, changed made

src/context.hpp
rjodinchr: I agree, that's what I tried to do initially. But it is not simple. clvk's headers contain too much code everywhere leading to build errors (you can see that `context.hpp` does not include `queue.hpp` in this PR, it's only declaring `struct cvk_command_queue`, thus making the destructor inaccessible for `std::unique_ptr`).

I've tried to move a few things around, but it quickly started to be too much making this PR too complicated for what it is trying to achieve.

I can open an issue to have a rework task for it to do once this land if you want.

kpet: Right, we do need to untangle those headers a bit. I've already bumped into similar issues in the past. Please create an issue with as many specifics as possible. Thanks!

src/queue.hpp
```diff
@@ -175,6 +182,11 @@ struct cvk_command_queue : public _cl_command_queue,
     }
 
     cvk_buffer* get_or_create_printf_buffer() {
+        if (m_init_image_queue) {
+            cvk_error_fn("Could not get printf buffer from init-image queue");
+            CVK_ASSERT(false);
+            return nullptr;
+        }
```
kpet, line 189: Why is this required?

src/queue.cpp
```diff
@@ -63,6 +64,10 @@ cvk_command_queue::cvk_command_queue(
 
     TRACE_CNT(batch_in_flight_counter, 0);
     TRACE_CNT(group_in_flight_counter, 0);
+
+    if (init_image_queue) {
+        m_context.reset(nullptr);
+    }
```
kpet, line 70: Why is this required?

rjodinchr: `cvk_context` is refcounted. It will not be freed until every object releases their hold on it.
`cvk_command_queue` has a ref on `cvk_context` and that new queue for image initialisation is freed in `cvk_context` destructor with that PR.
Thus to allow everything to get freed properly, the new queue for image initialisation releases the context at creation.

src/queue.hpp
rjodinchr: because the queue for image initialisation does not have a context, which is needed for that function.
Image initialisation should not need printf.

src/context.hpp
rjodinchr: https://github.com/kpet/clvk/issues/735

src/queue.cpp
kpet: That makes sense, thanks. Hmm, rather than introducing a new type of queue effectively, with state in every single command queue, we could introduce a `cvk_command_queue::detach_from_context` function doing the above. That function would only be called in `get_or_create_image_init_command_queue` and you could just assert that `m_context` is non-NULL in `get_or_create_printf_buffer` (although this should really never be called if the command queue is only used for image init; part of me thinks this assert could be omitted, the resulting crash would be very obvious and we wouldn't run the risk of adding new functions that use `m_context` without a corresponding assert). WDYT?

rjodinchr: I'm fine with that.
Done

kpet (approved):
Thanks!

Issue #723, opened 2024-10-08
get_image_width on 1Dbuffer is returning the wrong value


dneto0:
The conformance tests for get_image_width are failing for 1Dbuffer, when the the texel buffer byte size is larger than the `image_width * sizeof(texel)`

It's returning the `buffer_byte_size / sizeof(texel)` instead of the image width (the value of `cl_image_desc.image_width` property  when the image was created).

Here's a failing  test run (with some extra debugging output; and slightly changing the extra widths... but the problem exists on the existing code)

> ./test_conformance/images/kernel_image_methods/test_kernel_image_methods  debug_trace 1Dbuffer

```
Compute Device Name = Quadro P1000, Compute Device Vendor = NVIDIA Corporation, Compute Device Version = OpenCL 3.0 CLVK on Vulkan v1.3.242 driver 2246950976, CL C Version = OpenCL C 1.2 CLVK on Vulkan v1.3.242 driver 2246950976
Device latest conformance version passed: v2023-12-12-00
Supports single precision denormals: NO
sizeof( void*) = 8  (host)
sizeof( void*) = 4  (device)
1Dbuffer...
Running 1D image buffer read-only tests...
[CL_BGRA CL_SIGNED_INT8           4]
   at size 2647297 (row pitch 10589408) out of 134217728
 - Creating 1D image 2647297 ...
ERROR: Returned width did not validate (expected 2647297, got 2647352)
FAILED: [CL_BGRA CL_SIGNED_INT8           4]

[CL_RGBA CL_UNSIGNED_INT16        4]
   at size 2647297 (row pitch 21179000) out of 134217728
 - Creating 1D image 2647297 ...
ERROR: Returned width did not validate (expected 2647297, got 2647375)
FAILED: [CL_RGBA CL_UNSIGNED_INT16        4]
...
```

The full log is attached
[image_methods.txt](https://github.com/user-attachments/files/17294411/image_methods.txt)

The failure also happens with SwiftShader.

dneto0:
This bug was exposed after https://github.com/KhronosGroup/OpenCL-CTS/pull/2073 was merged.

PR #724, opened 2024-10-08
Fix get_image_width() kernel method for 1D texel buffers

dneto0:
Return the image_width specified by the cl_image_desc.

The underlying buffer might be bigger than strictly necessary, because the rowPitch might have padding larger than a single pixel.

Therefore, when creating the underlying buffer view, use a view size of element_size() * width() instead of WHOLE_SIZE.

Fixed: #723


dneto0:
@rjodinchr @kpet PTAL

rjodinchr (approved):

kpet (approved):
Thanks!

PR #725, opened 2024-10-08
Handle concurrent printf

Rekt3421:
This change moves the printf buffer from the queue to the kernel to resolve the issue of concurrent printf not printing properly

### DIFF
--- a/src/queue.cpp
+++ b/src/queue.cpp
@@ -667,7 +667,7 @@ cl_int cvk_command_kernel::update_global_push_constants(
         CVK_ASSERT(pc->size == 8);
         CVK_ASSERT(program->uses_printf());
 
-        auto buffer = m_queue->get_printf_buffer();
+        auto buffer = get_printf_buffer();
         if (buffer == nullptr) {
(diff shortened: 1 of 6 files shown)


src/queue.cpp
```diff
@@ -1041,8 +1041,8 @@ cvk_command_kernel::build_batchable_inner(cvk_command_buffer& command_buffer) {
     // Setup printf buffer descriptor if needed
     if (m_kernel->program()->uses_printf()) {
         // Create and initialize the printf buffer
-        auto buffer = m_queue->get_or_create_printf_buffer();
-        auto err = m_queue->reset_printf_buffer();
+        auto buffer = get_or_create_printf_buffer();
+        auto err = reset_printf_buffer();
```
rjodinchr, line 1045: I don't think we will ever reuse that buffer.
Those 2 functions could be merged into a single `create_printf_buffer`

rjodinchr (commented):
I think testing that in the CI is too complicated as it depends on the performance of the hardware.
You can remove that test for now.

PR #731, opened 2024-10-22
714-2 Fix vector 16 printf

Rekt3421:
This change fixes the functionality related to printf of vectors of size 16

### DIFF
--- a/src/printf.cpp
+++ b/src/printf.cpp
@@ -57,7 +57,7 @@ std::string get_vector_fmt(std::string fmt, int& vector_size, int& element_size,
 
     size_t vec_length_pos_start = ++pos;
     size_t vec_length_pos_end =
-        fmt.find_first_not_of("23468", vec_length_pos_start);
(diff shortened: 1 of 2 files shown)


rjodinchr (approved):

src/printf.cpp
```diff
@@ -56,8 +56,13 @@ std::string get_vector_fmt(std::string fmt, int& vector_size, int& element_size,
     fmt = fmt_specifier;
 
     size_t vec_length_pos_start = ++pos;
-    size_t vec_length_pos_end =
-        fmt.find_first_not_of("123468", vec_length_pos_start);
+    // Assume vec len is at max 8 in which case we need just one digit.
+    size_t vec_length_pos_end = pos + 1;
+    // Check if len is 16 and update accordingly. This is only possible iff
```
kpet, line 61:
```suggestion
    // Check if len is 16 and update accordingly. This is only possible if
```

kpet (approved):

PR #732, opened 2024-10-25
fix integer dot product support for device without int8 support

rjodinchr:


rjodinchr:
cc @dj2

rjodinchr:
The swiftshader artifact needs to be re-generated.

rjodinchr:
@kpet, this is blocking Google infra, could you have a look please?

kpet (approved):

PR #733, opened 2024-11-04
cmake: Disable clspv spirv build

rjodinchr:
This is causing issues with android build

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -121,6 +121,7 @@ if (CLVK_COMPILER_AVAILABLE)
 endif()
 set(CLSPV_SOURCE_DIR ${PROJECT_SOURCE_DIR}/external/clspv
     CACHE STRING "Clspv source directory")
+set(CLSPV_BUILD_SPIRV_DIS OFF)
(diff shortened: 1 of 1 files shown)


kpet:
Not sure I can reconcile this change with https://github.com/google/clspv/commit/38494814dac9cc7f0ffe9bbcec2bd727a9920c41. What's the issue? Does this change mean that clspv test can no longer be run from a clvk build tree?

rjodinchr:
There was no reason not to build `clspv-reflection` when building with Android. That's why https://github.com/google/clspv/commit/38494814dac9cc7f0ffe9bbcec2bd727a9920c41 was merged.

But it causes issues  when trying to build (only for Android, not sure why exactly), as some SPIRV targets are declared twice. Once from the `<clvk>/external/SPIRV-Tools` and another time from `<clvk>/external/clspv/third_party/SPIRV-Tools`.

As `clvk` is already building SPIRV-Tools (and defining all SPIRV targets needed for `clspv`), we do not need to do it on the `clspv` side.

The goal of this PR is to make sure everything will still work when we will update clspv here.

kpet (approved):
Thanks for the explanation. This change should not be needed but, at the same time, I can't think of an issue it'd create.

PR #739, opened 2024-11-14
CMake: allow to use a out-of-tree SPIRV-Tools sources

rjodinchr:


rjodinchr:
The error in `macos-latest` is a github infra issue.

kpet (approved):

PR #740, opened 2024-11-15
Marked "DaVinci Resolve" as Unsupported.

ru-ka:
I've tested DaVinci Resolve and [OpenCL-Benchmark](https://github.com/ProjectPhysX/OpenCL-Benchmark) within a machine with clvk correctly installed (since OpenCL-Benchmark works).

Then, to double check, I've tested the same softwares with the mesa Clover implementation.

The devices generated by clvk are correctly recognized but attempting to do any OpenCL accelerated operation (so basically everything in the software except tweaking the settings) crashes the app instantly.

### DIFF
--- a/docs/supported-applications.md
+++ b/docs/supported-applications.md
@@ -37,7 +37,7 @@
 | [Geekbench 5](https://www.geekbench.com/) | Partial support |
 | [Libreoffice](https://www.libreoffice.org/) | Untested |
 | [hashcat](https://github.com/hashcat/hashcat) | :x: Kernel build fails |
-| [DaVinci Resolve](https://www.blackmagicdesign.com/uk/products/davinciresolve) | Untested |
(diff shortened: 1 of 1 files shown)


kpet (approved):
Thank you for your contribution! In order for us to start looking into the issues it would be really helpful if you could re-run with the following environment variables set:
-  `CLVK_LOG=4`
- `CLVK_LOG_DEST=file:davinci.log`
- `CLVK_KEEP_TEMPORARIES=1`

These enable full logging of clvk's behaviour to the `davinci.log` file and the capture of the OpenCL program sources passed by the application. If you could then create an issue and attach the log file and any `clvk-*` folders (you might need to package them into a zip file) created, this would enable clvk contributors who cannot run DaVinci Resolve themselves to start working on solving the issues.

Thanks again!

P.S.: It seems that OpenCL-Benchmark did work well. Feel free to create another PR to add it to the list of tested apps ;)

PR #741, opened 2024-11-19
github actions: do not build libclc manually for android

rjodinchr:


rjodinchr:
Note that if it can't get in because of the runner we have in the CI, this is only for Android. All other targets are building everything as everyone is doing.

rjodinchr:
linking is failing for android, I would be in favour of just living the GitHub actions as they are nowadays. What do you think?

kpet:
Looks like we're running out of disk space. I'm trying something on top of your changes: https://github.com/kpet/clvk/tree/libclc-android

kpet (approved):

Issue #743, opened 2024-11-20
Trouble compiling on Pi 4


krakenrf:
Hi, the docs say this can be built on a Raspberry Pi, but I've been trying to compile this on a Pi 4 running Ubuntu 24.10 for the last few days without luck. Running on a 64GB SD card which I hope is big enough.

I've managed to get a little further each time with some fixes:

1) First I couldn't run fetch_sources as Git simply did not want to download llvm as I think it's too big. Using --shallow helped.

2) I kept getting an error `cc: error: unrecognized command-line option` during make. It seems that the compiler it was trying to use was gcc by default. I forced it to use clang by adding `-DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++` to cmake.

3) I have the Pi 4 2GB and 8GB models. At first I tried the 2GB model, and the compilation just kept crashing and causing the entire board to restart. Switching to the 8GB model helped me compile to a higher percentage, but it still crashed either with a reset, or just the terminal window closing. I then increased the swap space to 16GB, and again that got me further, but I still end up with a crash, or just with the terminal window closing itself without completion.

4) I tried `cmake ../ -DCLVK_BUILD_TESTS=OFF -DLLVM_INCLUDE_BENCHMARKS=OFF -DLLVM_INCLUDE_TESTS=OFF -DLLVM_ENABLE_BINDINGS=OFF -DLLVM_ENABLE_UNWIND_TABLES=OFF-DLLVM_BUILD_TOOLS=OFF -DCLSPV_BUILD_SPIRV_DIS=OFF -DCLSPV_BUILD_TESTS=OFF -DCLVK_BUILD_TESTS=OFF -DCLVK_BUILD_SPIRV_TOOLS=OFF -DCLVK_ENABLE_SPIRV_IL=OFF -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++` to try and cut down on some of the things that need to compile, but at some point during an overnight compile the terminal window just closes itself, and the compilation never completes.

5) Running `make -j1` to force one core seems to help to get it a little further, but ultimately I still end up with a reset or terminal window close.

The culprit appears to be that it never gets past the llvm compilation. One guess I have is that the crashes are from running out of memory, but I've never seen the ram bar full during compilation with htop up, though the crashes always occur when I leave it running overnight so I may be missing it.

I am Interested to hear if anyone has anything else I should try, or if anyone else had it successfully installed on a Pi 4 recently?

[...]

PR #745, opened 2024-11-26
do not copy config files to build directory

rjodinchr:
Add the path to the config files in the sources instead.

Ref #743


kpet (approved):
Thanks!

PR #751, opened 2024-12-06
Init the logging parts of the config before calling init_logging

rjodinchr:
This is needed when logging properties are set in a configuration file.


rjodinchr:
I will try to implement what we talked about this week:

Initialise a subset of the config before `init_logging`, and the rest after. That would make for 2 `init_config` functions, not sure of their name right now.

kpet (approved):
LGTM, thanks!

PR #753, opened 2024-12-27
fix cvk_rectangle_copier constructor

rjodinchr:
row_pitch and slice_pitch can be set to '0', which means that it needs to be computed from region values.


rjodinchr:
This has been found running `test_cl_khr_command_buffer copy_buffer_rect` on ChromeOS.

kpet (approved):

PR #754, opened 2025-01-09
device_properties: make driverVersion and driverId available

rjodinchr:


src/device.cpp
```diff
@@ -1146,10 +1149,10 @@ bool cvk_device::init(VkInstance instance) {
         return false;
     }
 
-    init_clvk_runtime_behaviors();
-
     init_vulkan_properties(instance);
 
+    init_clvk_runtime_behaviors();
+
```
rjodinchr, line 1155: `init_clvk_runtime_behaviors` needs to be called after `init_vulkan_properties` to have `m_driver_properties` set

kpet (approved):

PR #755, opened 2025-01-17
device_properties: nvidia: remove fma

rjodinchr:
Since clspv optimisation to generate fma instead of fmul+fadd (https://github.com/google/clspv/commit/c4f91b91f195cae573aa8bec21af4458b614c9bc) the erf test is failing on nvidia (as it is using fma) in google infra.
Do not use native fma on nvidia to allow that platform to stay compliant.


rjodinchr:
Let's just change the condition for that mad->FMA optimisation: https://github.com/google/clspv/pull/1441

PR #759, opened 2025-01-24
fix init image at creation

rjodinchr:
We need to create the cvk_command_image_init with a valid context. Otherwise we can have a race condition between the destruction of queue and commands being deleted in the executor thread.


rjodinchr:

kpet:
Kicked off https://github.com/kpet/clvk/actions/runs/12951700691

rjodinchr:
Can you re-trigger the failing bot? It's the first time I'm seeing that error. I wonder if it happens just very rarely. I'm not able to reproduce on my machine. That may be something we just want to filter in the test to avoid flakyness I guess

rjodinchr:
In fact, the bug was much simpler than what I was thinking. It's just about not forcing the deletion of the queue.

rjodinchr:
CI should be fixed with https://github.com/kpet/clvk/pull/771

rjodinchr:
Fixing the bug also showed another deadlock between the executor and the main thread, for which I've added the fix.

The deadlock involves the lock of `cvk_executor_thread.m_lock` and `cvk_executor_thread_pool.m_lock`. The main thread has `cvk_executor_thread_pool.m_lock` running in `cvk_executor_thread_pool::get_executor`, and is trying to take `cvk_executor_thread.m_lock` in `cvk_executor_thread::is_idle`, while an executor is holding `cvk_executor_thread.m_lock` from `cvk_executor_thread::executor` and is trying to get `cvk_executor_thread_pool.m_lock` while deleting the queue that noone hold anymore in `cvk_executor_thread_pool::return_executor`.

```
main:
  cvk_executor_thread::is_idle -> tries to take cvk_executor_thread.m_lock
  cvk_executor_thread_pool::get_executor -> has cvk_executor_thread_pool.m_lock
executor:
  cvk_executor_thread_pool::return_executor -> tries to take cvk_executor_thread_pool.m_lock
  cvk_executor_thread::executor -> has cvk_executor_thread.m_lock
```

kpet:
How did you bump into that deadlock? What test or application were you running?

rjodinchr:
> How did you bump into that deadlock? What test or application were you running?

Several tests in `api_tests` started to be flaky (with that deadlock) once I updated ChromeOS with that new version of the fix (with only the release on the command queue).

kpet (approved):

PR #760, opened 2025-01-24
fix clEnqueueNDRangeKernel for NULL work_global_size

rjodinchr:
According to the OpenCL specification, work_dim can be 0 and work_global_size can be NULL.

https://github.com/KhronosGroup/OpenCL-CTS/issues/2240

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -4080,6 +4080,11 @@ cl_int CLVK_API_CALL clEnqueueNDRangeKernel(
         command_queue, kernel, work_dim, num_events_in_wait_list,
         event_wait_list, event);
 
+    static const size_t null_global_work_size[3] = {0};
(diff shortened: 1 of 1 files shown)


rjodinchr:

rjodinchr:
The issue is shown by running: `test_profiling profiling_timebase`. It segfaults with llvmpipe on my end.

It passes with `zero_sized_enqueue` because that test never tries to pass `NULL` as the `global_work_size`, it always gives a valid structure with nothing to compute. While `profiling_timebase` calls `clEnqueueNDRangeKernel` with `NULL` as the `global_work_size`

rjodinchr:
TODO: change the patch to override `global_work_size` when it equals nullptr

kpet (approved):
LGTM, thanks!

PR #761, opened 2025-01-24
fix image/buffer max size

rjodinchr:
OpenCL-CTS is checking that image/buffer max size is below CL_DEVICE_MAX_MEM_ALLOC_SIZE.

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -724,12 +724,15 @@ cl_int CLVK_API_CALL clGetDeviceInfo(cl_device_id dev,
         size_ret = sizeof(val_uint);
         break;
     case CL_DEVICE_IMAGE_MAX_BUFFER_SIZE:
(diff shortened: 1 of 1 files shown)


rjodinchr:

rjodinchr:
In fact, it should even need to be the min with `device->max_mem_alloc_size()/2` because the CTS doing a weird comparison with `CL_DEVICE_MAX_MEM_ALLOC_SIZE/MAX_DEVICE_MEMORY_SIZE_DIVISOR`.
I'll raise a bug in the CTS to ask the group.

rjodinchr:
https://github.com/KhronosGroup/OpenCL-CTS/issues/2245

rjodinchr:
Will be fixed by https://github.com/KhronosGroup/OpenCL-CTS/pull/2342

PR #770, opened 2025-03-05
fix reporting of CL_KERNEL_LOCAL_MEM_SIZE

rjodinchr:
parse clspv NonSemanticClspvReflectionWorkgroupVariableSize and add it to the local memory size needed by a kernel

### DIFF
--- a/src/kernel.cpp
+++ b/src/kernel.cpp
@@ -20,7 +20,7 @@
 #include "memory.hpp"
 
 cl_ulong cvk_kernel::local_mem_size() const {
-    cl_ulong ret = 0; // FIXME take the compile-time allocations into account
+    cl_ulong ret = m_program->workgroup_variables_size();
 
     for (uint32_t i = 0; i < m_args.size(); i++) {
         auto const& arg = m_args[i];
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -316,6 +316,11 @@ spv_result_t parse_reflection(void* user_data,
                 parse_data->binary->add_kernel_argument(kernel, std::move(arg));
                 break;
             }
+            case NonSemanticClspvReflectionWorkgroupVariableSize: {
+                auto size = parse_data->constants[inst->words[6]];
+                parse_data->binary->add_workgroup_variable_size(size);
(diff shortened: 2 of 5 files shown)


rjodinchr:
https://github.com/google/clspv/pull/1465

rjodinchr:
This is missing tests

rjodinchr:
@kpet , because `clvk` returns `SPV_ERROR_INVALID_DATA` when it does not recognise an `OpExtInst SPV_EXT_INST_TYPE_NONSEMANTIC_CLSPVREFLECTION`, we have issues in Google infra while waiting for this patch.
Could you review that CL please?
Maybe we should also change that behavior?

kpet (approved):
LGTM, thanks!

kpet:
>Maybe we should also change that behavior?

Yes, maybe we should relax the error for unrecognized instructions. We'd get compatibility in _some_ cases. We might want to think a bit more about the compatibility story in general, i.e. do we want to support using newer/older versions of clspv with a given version of clvk? What should we test in that case?

rjodinchr:
> LGTM, thanks!

Thank you for the quick review, very much appreciated! 🙏

> > Maybe we should also change that behavior?
> 
> Yes, maybe we should relax the error for unrecognized instructions. We'd get compatibility in _some_ cases. We might want to think a bit more about the compatibility story in general, i.e. do we want to support using newer/older versions of clspv with a given version of clvk? What should we test in that case?

I think `clvk` should not be expecting to know all clspv's non-semantic instructions.
When `clspv` is introducing new non-semantic instructions with breaking ABI, it is protected behind a flag, so any driver can enable it knowingly (like adding an argument to an existing instruction).

For this specific one, it's not breaking anything, it is just making more information available, that's why I did not put it behind a flag.

I don't think we want `clvk` to be able to support older versions of `clspv`, this is quite a lot of work and I don't think anyone has a real need for it.
The only thing I'm thinking about in terms of versioning is that maybe one day we will have clspv version (based on a fixed llvm version). Then maybe we will want to have a clvk version aligning on clspv?

PR #774, opened 2025-03-25
fix clGetProgramInfo for some param_names

rjodinchr:
- CL_PROGRAM_NUM_KERNELS
- CL_PROGRAM_KERNEL_NAMES,
- CL_PROGRAM_SCOPE_GLOBAL_CTORS_PRESENT
- CL_PROGRAM_SCOPE_GLOBAL_DTORS_PRESENT

Those params requires the program to have been built for at least one device according to the specification.
Returning the value when the program has not been built triggers a failure in the CTS (compiler get_program_info_kernel_names).

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -2543,10 +2543,12 @@ cl_int CLVK_API_CALL clGetProgramInfo(cl_program prog,
         return CL_INVALID_PROGRAM;
     }
 
-    // TODO CL_INVALID_PROGRAM_EXECUTABLE if param_name is
-    // CL_PROGRAM_NUM_KERNELS or CL_PROGRAM_KERNEL_NAMES and a successful
-    // program executable has not been built for at least one device in the list
-    // of devices associated with program.
+    bool program_built_once = false;
+    for (auto dev : program->devices()) {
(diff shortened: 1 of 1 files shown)


rjodinchr:
I'm not sure why the [specification](https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/clGetProgramInfo.html) has to be that strict:
```
This information is only available after a successful program executable has been built for at least one device in the list of devices associated with program.
```
Wouldn't a more relaxed approach be enough
```
This information might not be available before a successful program executable has been built for at least one device in the list of devices associated with program.
```
For OpenCL users, it's the same, if they want to be sure to access that information they have to make sure it is built for at least one device. But it leaves the freedom for the implementation to make it accessible before if it's possible.

kpet:
I'd sympathize with this point of view if the information we returned before a build was correct. AFAICT from a quick look we just return 0 kernels and an empty list of names prior to a build. Also, this error code has been in the spec for a long time. Someone somewhere may have made the bad decision of checking whether these queries fail with `CL_INVALID_PROGRAM_EXECUTABLE` to determine whether the program was at least built for one device ... We can't really change the behaviour now.

kpet (approved):

PR #775, opened 2025-03-26
add config to override CL_DEVICE_MAX_COMPUTE_UNITS

rjodinchr:
This is useful especially with application like clpeak which choose the workload to compute using CL_DEVICE_MAX_COMPUTE_UNITS


rjodinchr:
TODO: remove force option

kpet (approved):

PR #777, opened 2025-04-01
disable fp16 support if extension is masked

rjodinchr:


src/device.cpp
```diff
@@ -736,6 +736,10 @@ void cvk_device::build_extension_ils_list() {
         split_string(config.device_extensions_masked(), ',');
     for (auto& config_extension_masked : config_extensions_masked) {
         for (auto it = m_extensions.begin(); it != m_extensions.end(); it++) {
+            if (strcmp(config_extension_masked.c_str(),
+                       CL_KHR_FP16_EXTENSION_NAME) == 0) {
```
kpet, line 740:
```suggestion
            if (config_extension_masked == CL_KHR_FP16_EXTENSION_NAME) {
```

kpet (approved):

Issue #783, opened 2025-05-06
cts test copy_image can fail with CL_PROFILING_COMMAND_SUBMIT > CL_PROFILING_COMMAND_START.


dneto0:
The OpenCL CTS profiling tests can fail in a copy_image subtest.

The clvk commit is a63763d80e07cc5fa1e65b26b1e4c9c959dbc967 

```
Compute Device Name = Tesla V100-SXM2-16GB, Compute Device Vendor = NVIDIA Corporation, Compute Device Version = OpenCL 3.0 CLVK on Vulkan v1.3.242 driver  (.... undisclosed ....)
Supports single precision denormals: NO
sizeof( void*) = 8  (host)
sizeof( void*) = 4  (device)
copy_image...

...

CL_PROFILING_COMMAND_QUEUED: 1539060808899861 CL_PROFILING_COMMAND_SUBMIT: 1539060808910080 CL_PROFILING_COMMAND_START: 1539060809068296 CL_PROFILING_COMMAND_END: 1539060809070408 CL_DEVICE_PROFILING_TIMER_RESOLUTION: 1
Profiling info:
Time from queue to submit : 0.010219ms
Time from submit to start : 0.158216ms
Time from start to end: 0.002112ms
test copy of subimage size 217,72  196,35  33 x 30
Image verified.
CL_PROFILING_COMMAND_QUEUED: 1539060885566660 CL_PROFILING_COMMAND_SUBMIT: 1539060885577440 CL_PROFILING_COMMAND_START: 1539060869925410 CL_PROFILING_COMMAND_END: 1539060869927522 CL_DEVICE_PROFILING_TIMER_RESOLUTION: 1
Profiling info:
Time from queue to submit : 0.010780ms
Time from submit to start : 18446744073693.902344ms
Time from start to end: 0.002112ms
CL_PROFILING_COMMAND_SUBMIT > CL_PROFILING_COMMAND_START.
Time between submit and start is too big: 18446744073.693901s, test limit: 30.000000s.
test copy of subimage size 91,109  37,204  98 x 14
copy_image FAILED
```

I'm only showing the last two subtest reports from the log.  The last one has the SUBMIT time after the START time, and wraparound has not occurred.

This is not urgent. In our system it shows up as a test flake; a second run almost always passes.

dneto0:
My internal reference is crbug.com/416035019

kpet:
[...]

PR #784, opened 2025-05-07
update device_host_timer management

rjodinchr:
- Protect cvk_device internal values with mutex
- Update cvk_device internal values at START time

Ref #783


rjodinchr:
Swiftshader artifacts need to be refreshed.

rjodinchr:
@kpet could you refresh the swiftshader artifacts?

rjodinchr:
I would really like to be able to test that patch in Google infra. Could we merge it and review the state 2 weeks after to see the impact?

PR #795, opened 2025-07-02
fix warnings in api.cpp following OpenCL-Headers update

rjodinchr:
When updating OpenCL-Headers to ToT, `api.cpp` now generates a warning (turning into an error in google infrastructure).


kpet (approved):
Could you rebase on top of or merge the latest main to pick up the Windows CI fix?

PR #801, opened 2025-08-11
add config option to keep allocated memories mapped

rjodinchr:
Some workloads on some devices can get huge performance improvement by keeping allocated memories mapped.


src/config.def
```diff
@@ -66,6 +66,8 @@ OPTION(uint32_t, enqueue_command_retry_sleep_us, UINT32_MAX) // UINT32_MAX meani
 
 OPTION(bool, supports_filter_linear, true)
 
+OPTION(bool, keep_allocated_memories_mapped, false)
```
kpet, line 69:
```suggestion
OPTION(bool, keep_memory_allocations_mapped, false)
```

kpet (approved):

PR #805, opened 2025-08-19
Update NDK

xueliang-zhong-arm:
Update NDK to NDK r27.


xueliang-zhong-arm:
I've updated to patch to use r27d.

kpet (approved):
LGTM, thanks! We'll need to improve how the NDK is managed in the CI at some point...

PR #806, opened 2025-08-20
cmake: fix OpenCL-CTS support

rjodinchr:
OpenCL-CTS depends on SPIRV-Headers since https://github.com/KhronosGroup/OpenCL-CTS/commit/555b7cd38302383efe768074eb6026f4f4af9869


tests/CMakeLists.txt
```diff
@@ -72,5 +72,6 @@ add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/simple-from-binary)
 
 option(CLVK_BUILD_CONFORMANCE_TESTS "Build OpenCL conformance tests")
 if (CLVK_BUILD_CONFORMANCE_TESTS)
+set(SPIRV_INCLUDE_DIR "${SPIRV-Headers_SOURCE_DIR}")
```
kpet, line 75: This should be defined in the `CMakeLists.txt` file that's in the `conformance` folder, shouldn't it?

kpet (changes requested):

tests/CMakeLists.txt
rjodinchr: I'm not sure I understand. The same way we define `SPIRV-Headers_SOURCE_DIR` for `SPIRV-Tools` to help it find its `SPIRV-Headers`, we are defining `SPIRV_INCLUDE_DIR` here for `OpenCL-CTS` to find its `SPIRV-Headers`.

kpet (approved):

PR #807, opened 2025-08-21
cl_khr_command_buffer

rjodinchr:
@kpet draft implementation followed by:

- Compile with CL_ENABLE_BETA_EXTENSIONS
- Complete implementation for new cl*CommandBufferKHR functions
- Complete implementation for new DeviceInfo entries
- Commands can be reused:
  - Reset event
  - Remove command from event
  - Move QueryPool inside event instead of command
- Add optional support to fill image on device (used by default for
  command buffer)
- Refcount command for easier lifetime management
- cvk_command_batch holds both commands and events as getting event
  through command cannot be done once the command has been executed.
- Add 'clone' method for all commands for queue migration


rjodinchr:
I have updated that PR with a new implementation that does not require all images to use read/write flags. It creates a second image (with write_only flag) using the same `VkDeviceMemory` at runtime when it detects a fillimage on a read_only image.

rjodinchr:
Marking as draft while the fill image on device PR is under review.

PR #809, opened 2025-08-21
Add config option to reuse descriptor_set

rjodinchr:
On Mali devices, using the ARM DDK, some application quickly get to a point where the driver consider the memory as fragemented and it not capable of allocated new descriptor_set anymore.

To avoid that, do not free descriptor_set, and when it is not possible to allocate new ones, reuse ones that have been release from the application point of view.


rjodinchr:
@kpet could you please refresh the swiftshader artifact?

kpet (approved):

PR #810, opened 2025-08-27
move CLVK_CONFIG_FILE to config

rjodinchr:
run parse_env before parse_config_file to be able to set
CLVK_CONFIG_FILE with 'log' set inside it and still have it set before
'init_logging' which happens between 'init_early_config' and
'init_config'.

Make sure not to override env value with config file ones

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -316,13 +316,10 @@ void parse_config_file(bool early_option) {
     config_file_paths.push_back("~/.config/clvk.conf");
     config_file_paths.push_back(
         (std::filesystem::current_path() / conf_file).string());
-    // First check if env var has file
-    std::string conv_file_env_var = "CLVK_CONFIG_FILE";
-    const char* conf_file_env_path = getenv(conv_file_env_var.c_str());
-
-    if (conf_file_env_path != nullptr) {
-        config_file_paths.push_back(conf_file_env_path);
+    if (!config.config_file().empty()) {
+        config_file_paths.push_back(config.config_file());
     }
+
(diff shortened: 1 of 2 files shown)


rjodinchr:
In fact, the 4 commits cannot be reviewed in parallel, they all depend on the previous one.
I have created https://github.com/kpet/clvk/pull/815 with the first commit.
I've marked this one as a draft. Until we get to the last one?
The idea being to keep the end goal visible.

src/config.cpp
```diff
@@ -82,7 +82,7 @@ void parse_string_set(void* value_ptr, const char* txt) {
         val = X;                                                               \
         return true;                                                           \
     }
-bool channel_order_from_string(std::string s, cl_channel_order& val) {
+bool channel_order_from_string(std::string& s, cl_channel_order& val) {
```
kpet:
```suggestion
bool channel_order_from_string(const std::string& s, cl_channel_order& val) {
```

src/config.cpp
```diff
@@ -104,7 +104,7 @@ bool channel_order_from_string(std::string s, cl_channel_order& val) {
     CASE(CL_sRGBx)
     return false;
 }
-bool channel_type_from_string(std::string s, cl_channel_type& val) {
+bool channel_type_from_string(std::string& s, cl_channel_type& val) {
```
kpet:
```suggestion
bool channel_type_from_string(const std::string& s, cl_channel_type& val) {
```

kpet:
Also needs conflicts resolved.

kpet (approved):

PR #813, opened 2025-09-05
remove frexp from native builtins for nvidia

rjodinchr:


rjodinchr:
Somehow Google infra regressed running frexp on Nvidia. I don't get exactly why.

The specification seems to match what `clspv` is generating:

> The exponent is returned through the pointer-parameter exp. The exp operand must be a pointer to a scalar or vector with integer component type, with 32-bit component width. The number of components in x and what exp points to must be the same. If x is a zero, the exponent is 0.0. If x is an infinity or a NaN, the exponent is undefined.

https://registry.khronos.org/SPIR-V/specs/unified1/GLSL.std.450.html

But when looking at the validation in [SPIRV-Tools](https://github.com/KhronosGroup/SPIRV-Tools/blob/main/source/val/validate_extensions.cpp#L1399) the story is different:

https://github.com/KhronosGroup/SPIRV-Tools/blob/90156db6aad5bc7eb41ee0406f66d8830eb86314/source/val/validate_extensions.cpp#L1424-L1437

But I don't see any recent changes in that part of the code. Not sure why it regresses recently.

Here is the failure log:
```
error: 128: GLSL.std.450 Frexp: expected operand Exp data type to be a 32-bit int scalar or vector type
  %90 = OpExtInst %float %89 Frexp %42 %88

[CLVK] SPIR-V binary is invalid.
```

rjodinchr:
I think we should land that PR now.

PR #815, opened 2025-09-09
always print warn/error/fatal, whatever the group

rjodinchr:


rjodinchr:
FYI, after that series of patches on logging, I would like to update that whole section in `README.md` to remove individual option documentation. Instead I would propose to document the format in `config.def` and all the individual documentation in `config.def`.
Like that everything will be together (name, type, default value, doc), and it will be easier (I hope) to see that documentation is required when we add a new option.

PR #816, opened 2025-09-09
Add option to fill images on device

rjodinchr:
Introduce a new configuration option `fill_image_on_device` to perform
`clEnqueueFillImage` operations on the device using a runtime-generated
compute kernel.

To support filling `CL_MEM_READ_ONLY` images (which cannot be written to
by kernels), implement a workaround that creates a temporary "write-enabled"
`cvk_image` alias (with `CL_MEM_WRITE_ONLY` flag) sharing the same
`VkDeviceMemory` allocation as the original image.

Implementation details:
- Added `fill_image_on_device` option to `src/config.def`.
- Intercepted `clEnqueueFillImage` in `src/api.cpp` to enqueue the new
  `cvk_command_fill_image_on_device` command when the option is enabled.
- Implemented `cvk_command_fill_image_on_device` in `src/queue.cpp` and
  `src/queue.hpp`. It dynamically generates OpenCL C source for the fill
  operation based on image type, coordinate type, and format, compiles
  it at runtime, and dispatches it.
- Added `cvk_image::create_write_enable_image_from` to support creating
  aliased images sharing the same memory allocation.
- Modified `cvk_image::init_vulkan_image` to accept an optional pre-allocated
  memory block.
- Updated `cvk_device::select_work_group_size` to support a null kernel
  argument, as it is called during command construction before the internal
  kernel is compiled.

### DIFF
--- a/src/memory.hpp
+++ b/src/memory.hpp
@@ -640,6 +640,8 @@ struct cvk_image : public cvk_mem {
                              std::vector<cl_mem_properties>&& properties,
                              cl_int* errcode_ret);
 
+    static cvk_image* create_write_enabled_image_from(cvk_image* image);
+
     bool is_backed_by_buffer_view() const {
         return type() == CL_MEM_OBJECT_IMAGE1D_BUFFER;
     }
@@ -831,9 +833,9 @@ struct cvk_image : public cvk_mem {
                               size_t* size_ret) const;
 
 private:
-    cl_int init_vulkan_image();
+    cl_int init_vulkan_image(std::shared_ptr<cvk_memory_allocation> memory);
     cl_int init_vulkan_texel_buffer();
-    cl_int init();
+    cl_int init(std::shared_ptr<cvk_memory_allocation> memory = nullptr);
 
     size_t num_channels() const {
         switch (m_format.image_channel_order) {
--- a/src/queue.hpp
+++ b/src/queue.hpp
@@ -1130,3 +1130,48 @@ struct cvk_command_image_init final : public cvk_command_batchable {
 private:
     cvk_image_holder m_image;
 };
+
+struct cvk_command_fill_image_on_device final : public cvk_command_batchable {
+    cvk_command_fill_image_on_device(
+        cvk_command_queue* queue, cvk_image* image,
+        const cvk_image::fill_pattern_array& fill_color,
+        const std::array<size_t, 3>& work_offset,
+        const std::array<size_t, 3>& work_size)
+        : cvk_command_batchable(CL_COMMAND_FILL_IMAGE, queue), m_image(image),
+          m_fill_color(fill_color), m_work_offset(work_offset),
+          m_work_size(work_size) {
+
+        if (!m_image->has_any_flag(CL_MEM_WRITE_ONLY | CL_MEM_READ_WRITE)) {
+            auto img = cvk_image::create_write_enabled_image_from(m_image);
+            m_image_write_enabled.reset(img);
+            if (img != nullptr) {
+                img->release();
+            }
+            m_mem = m_image_write_enabled;
+        } else {
+            m_mem = image;
+        }
+    }
+
+    CHECK_RETURN cl_int
+    build_batchable_inner(cvk_command_buffer& cmdbuf) override final;
+
+    const std::vector<cvk_mem*> memory_objects() const override {
+        return {m_image};
+    }
+
+private:
+    cl_uint dimensions() const;
+    const char* get_image_access_qualifier() const;
+    const char* get_image_type() const;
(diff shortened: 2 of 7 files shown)


rjodinchr:
This PR is a subset of https://github.com/kpet/clvk/pull/807

src/memory.hpp
```diff
@@ -700,6 +701,7 @@ struct cvk_image : public cvk_mem {
     cvk_mem* buffer() const { return icd_downcast(m_desc.buffer); }
     cl_uint num_mip_levels() const { return m_desc.num_mip_levels; }
     cl_uint num_samples() const { return m_desc.num_samples; }
+    cl_image_info image_type() const { return m_desc.image_type; }
```
kpet: This seems to duplicate `cvk_mem::type`.

src/queue.hpp
```diff
@@ -1110,3 +1110,58 @@ struct cvk_command_image_init final : public cvk_command_batchable {
 private:
     cvk_image_holder m_image;
 };
+
+struct cvk_command_fill_image_on_device final : public cvk_command_batchable {
+    cvk_command_fill_image_on_device(cvk_command_queue* queue, cl_mem image,
```
kpet: Any reason `image` is not a `cvk_image`?

src/queue.hpp
```diff
@@ -1110,3 +1110,58 @@ struct cvk_command_image_init final : public cvk_command_batchable {
...
+                                     const std::array<uint8_t, 16> fill_color,
```
kpet: `cvk_image::fill_pattern_array`?

src/queue.hpp
```diff
@@ -1110,3 +1110,58 @@ struct cvk_command_image_init final : public cvk_command_batchable {
...
+    }
+
+private:
+    cl_uint dimensions() const;
+    const char* get_image_access_qualifier() const;
+    const char* get_image_type() const;
+    const char* get_image_coord() const;
+    void get_image_color(char* color) const;
+
+    cl_mem m_mem;
```
kpet, line 1160: Do you need another member? Can't you just use the image object with a cast?

src/queue.hpp
```diff
@@ -1110,3 +1110,58 @@ struct cvk_command_image_init final : public cvk_command_batchable {
...
+            m_mem = image;
+        }
+    }
+
+    CHECK_RETURN cl_int
+    build_batchable_inner(cvk_command_buffer& cmdbuf) override final;
+
+    bool can_be_batched() const override final {
+        return cvk_command_batchable::can_be_batched();
+    }
```
kpet: I think you don't need this override.

src/queue.hpp
```diff
@@ -1110,3 +1110,58 @@ struct cvk_command_image_init final : public cvk_command_batchable {
...
+        m_local_size[1] = local_size[1];
+        m_local_size[2] = local_size[2];
+
+        if (!m_image->has_any_flag(CL_MEM_WRITE_ONLY | CL_MEM_READ_WRITE)) {
+            m_image_write_enabled.reset(
+                cvk_image::create_write_enable_image_from(m_image));
+            m_mem = m_image_write_enabled;
+        } else {
+            m_mem = image;
+        }
```
kpet, line 1143: Instead of creating another image and managing all that complexity, maybe we can just always require `VK_IMAGE_USAGE_STORAGE_BIT` when the option is enabled? This would arguably reduce how many formats we support when image fill commands are not used but with your approach we're lying when reporting supported formats anyway, i.e. if we report a format as supported for read-only images, we must support fill image commands on images with this format but we don't if we're not able to create a storage image from it. The alternative would be to alias a buffer for linearly-tiled images but that's more work that we may not want to do without a clear need.

kpet (commented):
As discussed, here are a bunch of older review comments that may or may not be still relevant. Sharing them in case you find them useful.

src/queue.hpp
rjodinchr: When we would switch to `cl_khr_command_buffer`, we would need to have it all the time. Even with the option disabled, as we don't know if an image will be used with a command buffer (and a fillimage), we have to set it up as if it will be used for that.
It would have a big performance impact on all the use case where the image is only read while we set it up with write flags just in case.

src/queue.hpp
rjodinchr: This is just a pointer on either `m_image` or `m_image_write_enabled`. It could be just a bool I guess, it would take less space.

rjodinchr:
I'm wondering if it would be better/easier to make a staging buffer and to copy it at runtime using `vkCmdCopyBufferToImage`. It would require more memory, but less problems with the image format.
What do you think?

PR #818, opened 2025-10-01
recalibrate timestamps when drift is dectected

rjodinchr:
Because of device clock drift, we can end up with uncoherent submit/start/end timestamps.
When such case is dectected, recalibrate start/end time using the device timestamps.

For start time, consider that end time is correct and put start time before end time of the same offset we have seen on the device domain. Make sure that submit is still happening before start time.

For end time, we can only rely on submit time.

### DIFF
--- a/src/event.hpp
+++ b/src/event.hpp
@@ -27,6 +27,34 @@
 struct cvk_command;
 struct cvk_command_queue;
 
+#define CLVK_PROFILING_COMMAND_START_DEVICE (CL_PROFILING_COMMAND_END + 1)
+#define CLVK_PROFILING_COMMAND_END_DEVICE (CL_PROFILING_COMMAND_END + 2)
+
+static inline cl_ulong recalibrate_start_time(cl_ulong start_time,
+                                              cl_ulong end_time,
+                                              cl_ulong submit_time,
+                                              cl_ulong start_dev_time,
+                                              cl_ulong end_dev_time) {
+    cvk_warn_fn(
+        "Host start time %lu is not between submit and end time ([%lu, %lu]), "
+        "most likely this is due to a drift in the device timer. Recalibrating "
+        "the host start time using the device timestamps.",
+        start_time, submit_time, end_time);
+    start_time = end_time - (end_dev_time - start_dev_time);
+    // Make sure submit time happens before start time.
+    return std::max(start_time, submit_time);
+}
+static inline cl_ulong recalibrate_end_time(cl_ulong end_time,
+                                            cl_ulong submit_time,
+                                            cl_ulong start_dev_time,
+                                            cl_ulong end_dev_time) {
+    cvk_warn_fn("Host end time %lu is below submit %lu, most likely this is "
+                "due to a drift in the device timer. Recalibrating the host "
+                "end time using the device timestamps.",
+                end_time, submit_time);
+    return submit_time + (end_dev_time - start_dev_time);
+}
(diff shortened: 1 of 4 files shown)


rjodinchr:
We might want to put that behind an option?

src/event.hpp
```diff
@@ -27,6 +27,34 @@
...
+}
+static inline cl_ulong recalibrate_end_time(cl_ulong end_time,
+                                            cl_ulong submit_time,
+                                            cl_ulong start_dev_time,
+                                            cl_ulong end_dev_time) {
+    cvk_warn_fn("Host end time %lu is below submit %lu, most likely this is "
+                "due to a drift in the device timer. Recalibrating the host "
+                "end time using the device timestamps.",
+                end_time, submit_time);
+    return submit_time + (end_dev_time - start_dev_time);
```
renatopereyra, line 55: This doesn't include the time in between `submit_time` and `start_dev_time`. If the GPU is backed up due to a lot of GPU work, I think that could be a non-neglible amount of time

src/event.hpp
```diff
@@ -27,6 +27,34 @@
...
+                                              cl_ulong start_dev_time,
+                                              cl_ulong end_dev_time) {
+    cvk_warn_fn(
+        "Host start time %lu is not between submit and end time ([%lu, %lu]), "
+        "most likely this is due to a drift in the device timer. Recalibrating "
+        "the host start time using the device timestamps.",
+        start_time, submit_time, end_time);
+    start_time = end_time - (end_dev_time - start_dev_time);
+    // Make sure submit time happens before start time.
+    return std::max(start_time, submit_time);
```
renatopereyra, line 45: I could be wrong but I'm not sure that `start_time` is actually being updated above since `start_time` and `end_time` would probably be based on the same `m_sync_host` and `m_sync_dev`.

If `start_time` is not being updated, then we're fixing the problem by setting `start_time` to `submit_time`. If `start_time` **is** getting updating, at least _some_ of the time we're still doing that. Needing to do it this way makes this approach seem a bit like hiding the problem rather than fixing it.

src/event.hpp
rjodinchr: We re-calibrate `end _time` when it is below `submit_time`. In that case we have no way to know how long time elapsed between the `submit_time` and `start_time`. The only thing available to us is the runtime on the device domain. So we are just using that to put `end_time` no too far in the future (compared to where we wanted to put it with the translation), and then we have no other choice than putting `start_time` equal to `submit_time`.

renatopereyra:
I thought of another idea for recalibrating timestamps.

We could add another sync timestamps (let's call it `m_sync_clvk` for now) to align with `m_sync_dev` and `m_sync_host`. We could set it by averaging the result of `sample_clock` before and after calling `vkGetCalibratedTimestampsEXT`. The idea would be to use this `m_sync_clvk` to convert between the `sample_clock` domain and the domain of the host timestamp returned by the Vulkan driver. Empirically, `sample_clock` differed from the host timestamp returned by `vkGetCalibratedTimestampsEXT` by more than 1 ms in some of my testing.

The idea would be that all the timestamps belonging to an event would key off of a single `vkGetCalibratedTimestampsEXT` sync and they would be recalibrated based on that sync point (either dev->host or clvk->host). The `m_sync_*` could be stored within the individual `cvk_event` as an add on.

Don't have to do it but thought I'd share.

Issue #822, opened 2025-10-23
Compile and link flows lose compile options


alan-baker:
When performing compile and link flows clspv is instructed to compile to IR. This skips the optimizations passes; however currently clspv needs to know several of those options to compile properly. For example, whether generic address space is used (and the requirement to inline). Linking programs are not expected to resupply those options. So those options are sort just lost in the process. This can lead to internal compiler errors in clspv. It probably doesn't make sense to try and optimize before generating IR, but there needs to be a way to include those options later. A worrisome case would be when different programs are linked that have different compile options.

rjodinchr:
I've worked a bit on it, I think it is better to do it in `clvk` in the end. I am starting to implement it.

`clspv` re-parsing looks too complicated, and the worry I had about exporting the binaries in `clvk` looks manageable.

PR #824, opened 2025-10-27
Update build options in cvk_program

rjodinchr:
OpenCL requires certain compiler options and language standards to be
retained across compilation units (compiled objects and libraries) and
carried forward when linking programs with clLinkProgram or saving/loading
binaries with clGetProgramInfo / clCreateProgramWithBinary.

This commit updates the CLVK binary format and build options handling:
* Bump CLVK binary format version to 2 and add `build_options_size` to
  `clvk_binary_header` to serialize preserved build options directly in
  the binary.
* Introduce `cvk_preserved_options` and `cvk_build_options` to parse,
  format, and merge compiler options according to OpenCL linking rules:
  - Preserves options that must be in all inputs ("all", e.g.,
    `-cl-fast-relaxed-math`, `-cl-denorms-are-zero`, etc.).
  - Preserves options that can be in at least one input ("one", e.g.,
    `-g`, `-cl-opt-disable`, `-cl-kernel-arg-info`, etc.).
  - Selects the maximum OpenCL C standard version (`-cl-std=`) across
    all inputs.
* Update `cvk_program` binary serialization (`read`, `write`,
  `binary_size`) to read and write preserved build options.
* Update `cvk_program::build()` to merge preserved options from input
  programs when linking.
* Fix `sizeof(header)` to `sizeof(struct clvk_binary_header)` in
  `read_binary_header()`.
* Update `clGetProgramBuildInfo(CL_PROGRAM_BUILD_OPTIONS)` in `src/api.cpp`
  to safely handle value returns from `program->build_options()`

Ref #822

### DIFF
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -798,11 +798,31 @@ bool validate_binary(spir_binary const& binary,
 
 const uint32_t clvk_binary_magic =
     0x6B766C63; // "clvk" in ASCII in little-endian
-const uint32_t clvk_binary_version = 1;
+const uint32_t clvk_binary_version = 2;
+
+// CLVK Binary Format Layout:
+//
+// +----------------------------------+
+// | clvk_binary_header               |
+// | - magic (0x6B766C63)             |
+// | - version                        |
+// | - binary_type                    |
+// | - build_options_size             |
+// +----------------------------------+
+// | Build Options (variable size)    |
+// | - string of length               |
+// |   build_options_size             |
+// +----------------------------------+
+// | Program Code                     |
+// | - LLVM IR (for library/object)   |
+// |   OR                             |
+// | - SPIR-V (for executable)        |
+// +----------------------------------+
 struct clvk_binary_header {
     uint32_t magic;
     uint32_t version;
     uint32_t binary_type;
+    uint32_t build_options_size;
 };
 #define COPY_WORD(dst, src)                                                    \
     do {                                                                       \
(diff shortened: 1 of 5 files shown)


src/program.hpp
```diff
@@ -422,6 +423,118 @@ enum class build_operation
     link
 };
 
+const std::string m_source_languages[8] = {
+    "Unknown", "CL1.0", "CL1.1", "CL1.2",
+    "CL2.0",   "CL3.0", "CLC++", "CLC++2021",
+};
+class cvk_build_options {
```
kpet, line 448: It's not exactly clear to me why we need this. A first examination suggests we want to transform and/or filter application-provided options based on those stored in binaries. We should make this easier to understand with comments and/or changes to function names.

src/program.hpp
```diff
@@ -422,6 +423,118 @@ enum class build_operation
...
+        for (auto option : parse_binary_header_options_one(options)) {
+            m_binary_header_options_one.insert(option);
+        }
+        m_source_language =
+            std::max(m_source_language, parse_source_language(options));
+    }
+
+    std::string operator()() const {
+        return m_build_options + binary_header_options();
+    }
```
kpet: This is too easy to miss or misread when looking at code using it IMO. I suggest moving the functionality to a named function.

src/program.cpp
```diff
@@ -824,11 +825,21 @@ bool cvk_program::read_llvm_bitcode(const unsigned char* src, size_t size) {
     return false;
 }
 
+size_t cvk_program::binary_header_size() const {
+    return sizeof(struct clvk_binary_header) +
+           m_build_options.binary_header_options().size();
+}
+
```
kpet: Hmm, we should be clear as to what is in the header and what is not. Options are sort of stored in the header and sort of not after this change :).

The current binary layout is pretty simple:
1. Header, fully represented by a `clvk_binary_header` structure
2. Program code, LLVM or SPIR-V

After this change, we have
1. Header, `clvk_binary_header` part
2. Header, build options
3. Program code, LLVM or SPIR-V.

I'm not completely opposed to including build options in the notional header but I think we may want to start documenting the binary format a bit more if we do that (a long comment in `program.[ch]pp` would likely suffice).

I'm also wondering whether we could justify spending the effort to add minimal support for sections with their dedicated header. It would make for a more flexible and extensible format. I'm thinking something along the lines of

Binary: one Header, one or more Sections

Section: SectionHeader, data
SectionHeader: enum or string name, size

kpet (commented):
Took a first look at this.

src/program.hpp
rjodinchr: Changed to `str()`

src/program.hpp
rjodinchr: I've added a comment.
When linking 2 programs together, we have 2 groups of build options. Some need to be kept while others are discarded.

src/program.cpp
rjodinchr: I've added a comment for the binary layout. And I've cleanly separated the header and the build options.

kpet:
It seems tests are crashing with the latest changes.

rjodinchr:
Fixed, this is ready for review

kpet (approved):
Thanks for improving it further! This aspect of the spec is not well tested or specified. I'm sure we'll be finding interesting cases in the future :).

PR #827, opened 2025-10-31
Clean uniform work-group size option

rjodinchr:
Only add '-cl-arm-non-uniform-work-group-size' if options allow to split regions.

options do not allow to split region if '-cl-uniform-work-group-size' is used.

Return CL_INVALID_WORK_GROUP_SIZE if we need to split the region but cannot do it because of options.


ShaobinTao:
https://github.com/google/clspv/pull/1540 is merged now.  Can you merge this PR?  @rjodinchr

kpet (approved):

PR #831, opened 2025-11-04
CI: build libclc manually for android to save disk space

rjodinchr:


rjodinchr:
> Shame we're having to do that! Feels like a step backwards...

I agree, but this time at least it is not a pure build work-around, it is just a case of saving disk space in GitHub. Users can still build with the expected process on Android.

PR #832, opened 2025-11-04
fix default value for compiler_temp_dir

rjodinchr:


rjodinchr:
Require https://github.com/google/clspv/pull/1543 to remove the patch for clspv

kpet (commented):
Looks fine but I think you need to update the documented default in README.md.

rjodinchr:
> Looks fine but I think you need to update the documented default in README.md.

Good point, cannot wait to see https://github.com/kpet/clvk/pull/810 (which depends on https://github.com/kpet/clvk/pull/836) go through. After them, we should be in better shape to have the code and the documentation in one place.

kpet (approved):
Thanks!

PR #836, opened 2025-12-09
refactor device_properties to use config.def as the reference

rjodinchr:
- remove sections in device.{cpp,hpp} that were hard to maintain
- improve printing of relevant device_propeties
- support more data type in config.def for better parsing/printing

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -28,26 +28,142 @@ const config_struct config;
 
 namespace {
 
-template <typename T> constexpr config_option_type option_type() = delete;
-
-#define DEFINE_OPTION_TYPE_GETTER(ctype, type)                                 \
-    template <> constexpr config_option_type option_type<ctype>() {            \
-        return type;                                                           \
-    }
-
-DEFINE_OPTION_TYPE_GETTER(std::string, config_option_type::string)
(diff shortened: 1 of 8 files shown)


src/config.cpp
```diff
@@ -28,27 +28,143 @@ const config_struct config;
...
+
+    cfgval->set = true;
+}
+
+#define CASE(X)                                                                \
+    if (s == std::string(#X)) {                                                \
+        val = X;                                                               \
+        return true;                                                           \
+    }
+bool channel_order_from_string(std::string s, cl_channel_order& val) {
```
kpet, line 85: `s` could be a const reference.

src/config.cpp
```diff
@@ -28,27 +28,143 @@ const config_struct config;
...
+    CASE(CL_BGRA)
+    CASE(CL_ABGR)
+    CASE(CL_RGBx)
+    CASE(CL_sRGB)
+    CASE(CL_sRGBA)
+    CASE(CL_sBGRA)
+    CASE(CL_sRGBx)
+    return false;
+}
+bool channel_type_from_string(std::string s, cl_channel_type& val) {
```
kpet, line 107: same here

kpet (approved):

PR #859, opened 2026-04-08
Use new clspv interface with headers

rjodinchr:
The new clspv interface remove the need to dump header on disk when clvk is compiled with clspv online.

Update the config test to not use compiler_temp_dir as it does not always exist

### DIFF
--- a/src/config.def
+++ b/src/config.def
@@ -16,9 +16,7 @@
 // Compiler
 //
 OPTION(std::string, cache_dir, "")
-OPTION(std::string, compiler_temp_dir, (std::filesystem::temp_directory_path().string().c_str()))
 OPTION(bool, skip_spirv_capability_check, false)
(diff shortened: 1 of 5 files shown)


rjodinchr:
Depends on https://github.com/google/clspv/pull/1584

rjodinchr:
~~@kpet could you update clspv with the latest version?
Thank you 🙏~~

Wait, seem that we need to fix something in the new clspv interface for windows

rjodinchr:
All green ✅

kpet (approved):
Nice!

rjodinchr:
> ~@kpet could you update clspv with the latest version? Thank you 🙏~
> 
> Wait, seem that we need to fix something in the new clspv interface for windows

Note that this PR has updated `clspv`, but the `json` files have not been updated.

PR #860, opened 2026-04-15
Add missing cl_khr_subgroups extension report

rjodinchr:


rjodinchr:
In fact `cl_khr_subgroups` is not supported as long as `CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS` is not supported.

PR #862, opened 2026-04-16
fix out-of-bounds access when set is empty

rjodinchr:


rjodinchr:
This is a major issue as the set is empty by default

kpet (approved):

PR #864, opened 2026-04-23
Add support for VK_KHR_shader_fma

rjodinchr:


rjodinchr:
The vulkan SDK we are using is too old. We need a newer one, but the action we are using does not seem to be able to get `1.4.329.0` which is the first to include `VK_KHR_shader_fma`.

rjodinchr:
Using `1.4.341.1` we can still use humbletim action.

- `macos` are failing due to a validation report. I need to investigate and fix clvk.
- `android` is failing because we need to update the CI to build libclc properly following libclc CMake changes
- `windows` is failing because it cannot find `vulkanrt.zip`

rjodinchr:
It seems that on `windows` the vulkan SDK has been reduced to `bin` only. No more `include`/`lib` that we need. Not sure how to get those properly.

Note that while that PR is trying to add support for `VK_KHR_shader_fma`, the real issue here is about the Vulkan SDK update (to get updated, Vulkan-headers). Maybe we just need to clone the headers? And keep the current SDK (this would work short term, at some point we will need to update the SDK anyway I believe).

rjodinchr:
Addressing independent fixes (needed for VK_KHR_shader_fma) here:
- https://github.com/kpet/clvk/pull/865
- https://github.com/kpet/clvk/pull/866

rjodinchr:
Hum, `android-ndk-r29` won't be enough anyway. Even `r30-beta` only gives vulkan `1.4.317` which is not enough. I guess `VK_KHR_shader_fma` will have to wait a bit before getting to `clvk`.

rjodinchr:
`android-ndk-r30-beta2` is at `1.4.335` which contains `VK_KHR_shader_fma`

PR #866, opened 2026-04-27
Fix push constant range validation with multiple entry points

rjodinchr:
When a program has multiple entry points, and only a subset of them use image/sampler metadata, the SPIR-V shader declares a larger push constant block that accommodates all potential metadata across the program. Previously, the push constant range in the pipeline layout was calculated per entry point in `cvk_entry_point::init()`. This resulted in a mismatch between the declared SPIR-V push constant block size and the pipeline layout's push constant range size on entry points without metadata usage, triggering Vulkan validation layer errors.

This commit resolves the issue by:
* Moving the complete push constant range calculation from `cvk_entry_point` to `cvk_program::prepare_push_constant_range()`.
* Consolidating all potential push constant sources (binary constants, push constant kernel arguments, image metadata, and sampler metadata) upfront across the entire program.
* Ensuring the pipeline layout's push constant range consistently covers the maximum declared SPIR-V block size across all entry points.
* Updating the POD buffer size assertion in `create_pod_buffer()` to precisely match `m_entry_point->pod_buffer_size()`.

### DIFF
--- a/src/kernel.hpp
+++ b/src/kernel.hpp
@@ -416,7 +416,7 @@ struct cvk_kernel_argument_values {
 
 private:
     bool create_pod_buffer() {
-        CVK_ASSERT(m_pod_data->size() >= m_entry_point->pod_buffer_size());
+        CVK_ASSERT(m_pod_data->size() == m_entry_point->pod_buffer_size());
 
         // Create POD buffer and copy data to it
         m_pod_buffer = m_entry_point->allocate_pod_buffer();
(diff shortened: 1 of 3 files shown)


rjodinchr:
Note that we do not see that error in the GitHub CI because the VVL we use is too old.

kpet (approved):
It's a shame clspv only produces a single global struct. Better to align the logic here. This change is a nice cleanup on top of the fixes it brings. Thanks!

PR #869, opened 2026-05-08
Move clspv_options PROPERTY out of COMPILER_AVAILABLE flag

rjodinchr:
As other flags regarding clspv options like
non_uniform_decoration_broken, we can keep them defined even if the
compiler is not available.

It cleans device_properties a bit by removing the preprocessor macros.
I think config PROPERTY should be always defined to avoid having to
deal with macro in device_properties.

Also it makes me realize that since we refactor the config management,
we were adding the clspv_options twice without good reason, thus I
removed once instance of it to avoid duplicate options.

This patch also adds a warning message when a device property is
explicitly overwritten by a configuration value.

The warning is restricted to string, uint32, and boolean types because
these types perform a direct replacement. Other types are excluded from
this warning as they merge the configuration with the device property
rather than overwritting it entirely.

Additionally, `print_option()` now returns a `std::string` instead of a
`char*`. This refactor simplifies string management and improves memory
safety across the configuration logging logic.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -377,7 +377,8 @@ void print_config() {
                            });
     for (const auto& opt : options) {
         void* optval = const_cast<void*>(opt.value);
-        char* txt = print_option(opt.type, optval);
+        std::string print = print_option(opt.type, optval);
+        const char* txt = print.c_str();
         if (opt.set) {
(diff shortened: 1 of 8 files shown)


rjodinchr:
I need to add a warning when the default value from device_properties is overwritten.

rjodinchr:
Warning message added. PR description updated.

kpet (approved):

PR #873, opened 2026-06-29
Validate image dimensions and improve error code propagation

rjodinchr:
Specifically, this introduces dimension validation for Vulkan-backed images by querying the physical device's format properties via `vkGetPhysicalDeviceImageFormatProperties`. If the requested image dimensions exceed the hardware's supported max extents, it now explicitly returns `CL_INVALID_IMAGE_SIZE` rather than failing silently or generically down the line.
Also adds early null checks for `image_format` and `image_desc` descriptors within `cvk_create_image`.

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -4505,14 +4505,6 @@ cl_mem cvk_create_image(cl_context context, cl_mem_flags flags,
         return nullptr;
     }
     // TODO CL_INVALID_VALUE if values specified in flags are not valid.
-    // TODO CL_INVALID_IMAGE_FORMAT_DESCRIPTOR if values specified in
-    // image_format are not valid or if image_format is NULL.
-    // TODO CL_INVALID_IMAGE_DESCRIPTOR if values specified in image_desc are
-    // not valid or if image_desc is NULL.
-    // TODO CL_INVALID_IMAGE_SIZE if image dimensions specified in image_desc
-    // exceed the minimum maximum image dimensions described in the table of
-    // allowed values for param_name for clGetDeviceInfo for all devices in
-    // context.
     // TODO CL_INVALID_HOST_PTR if host_ptr in image_desc is NULL and
     // CL_MEM_USE_HOST_PTR or CL_MEM_COPY_HOST_PTR are set in flags or if
     // host_ptr is not NULL but CL_MEM_COPY_HOST_PTR or CL_MEM_USE_HOST_PTR are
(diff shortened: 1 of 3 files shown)


src/api.cpp
```diff
@@ -4499,20 +4499,12 @@ cl_mem cvk_create_image(cl_context context, cl_mem_flags flags,
                         const cl_image_format* image_format,
                         const cl_image_desc* image_desc, void* host_ptr,
                         std::vector<cl_mem_properties>&& properties,
-                        cl_int* errcode_ret) {
+                        cl_int& errcode_ret) {
```
kpet: It's a bit weird to change just that one function from a pointer to a reference.

src/memory.cpp
```diff
@@ -415,20 +415,43 @@ bool cvk_image::init_vulkan_image() {
...
+            cvk_error_fn("At least one image dimension is bigger than its "
+                         "limit: x(%u > %u) || y(%u > %u) || z(%u > %u)",
+                         extent.width, imageProperties.maxExtent.width,
+                         extent.height, imageProperties.maxExtent.height,
+                         extent.depth, imageProperties.maxExtent.depth);
+            return CL_INVALID_IMAGE_SIZE;
+        }
+    } else {
+        cvk_warn_fn(
+            "Could not get vulkan image properties before creating image");
```
kpet, line 448: Shouldn't this be an error?

rjodinchr: The idea is that, if that call fails, we will still continue with the image creation, just we have less checks (as it is today), so it is possible that the image can still be generated. Thus a warning in case, but not an error per say.

src/api.cpp
rjodinchr: I understand. I can revert that part if you prefer. I think it is a better pattern, but it is certainly not the common one we have in `clvk` nowadays.

If I am to revert, should I also use a pointer and not a ref in `cvk_image::create`?

kpet: Pros and cons. The pointer makes it clearer to the caller that the parameter is an output parameter but a reference improves readability a bit in the callee and cannot be null (which we want here). I'd rather we make this change wholesale everywhere if we decide on it rather than for just a couple of functions here.

src/memory.cpp
kpet: Fine for now.

kpet (approved):

PR #875, opened 2026-07-03
Support denormal modes based on device capabilities

rjodinchr:
Update device configuration and compiler option preparation to support preserving or flushing denormals based on Vulkan driver support.

- Update `cvk_device::fp_config` to report `CL_FP_DENORM` for a given precision only if the Vulkan driver supports preserving denormals for that width.
- Add `supports_denorm_preserve_*` and `supports_denorm_ftz_*` helpers to `cvk_device`.
- Update `cvk_device::supports_capability` to handle `spv::CapabilityDenormPreserve` and `spv::CapabilityDenormFlushToZero`.
- In `cvk_program::prepare_build_options` translate OpenCL options (`-cl-denorms-are-zero`, `-cl-unsafe-math-optimizations`, and `-cl-fast-relaxed-math`) and device capabilities into clspv options.

### DIFF
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -47,6 +47,7 @@
 #include "log.hpp"
 #include "program.hpp"
 #include "tracing.hpp"
+#include "utils.hpp"
 
 struct membuf : public std::streambuf {
     membuf(const unsigned char* begin, const unsigned char* end) {
@@ -927,6 +928,10 @@ std::string cvk_program::prepare_build_options(const cvk_device* device) const {
         options += " ";
         options += m_build_options;
     }
+    const bool denorms_are_zero =
+        options.find("-cl-denorms-are-zero") != std::string::npos ||
+        options.find("-cl-unsafe-math-optimizations") != std::string::npos ||
+        options.find("-cl-fast-relaxed-math") != std::string::npos;
     std::vector<std::pair<std::string, std::string>> option_substitutions = {
         // FIXME The 1.2 conformance tests shouldn't pass this option.
         //       It doesn't exist after OpenCL 1.0.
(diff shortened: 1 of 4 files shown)


rjodinchr:
@kpet Can you update clspv to https://github.com/google/clspv/commit/0f162d74eda0267f9d571a1fd640c6b89f22d7a7 ?

rjodinchr:
`Presubmit / Build macos-latest compiler true online true (pull_request)` is not a real failure I believe. We should restart it.

src/program.cpp
```diff
@@ -927,6 +927,10 @@ std::string cvk_program::prepare_build_options(const cvk_device* device) const {
         options += " ";
         options += m_build_options;
     }
+    bool denorms_are_zero =
```
kpet:
```suggestion
    const bool denorms_are_zero =
```

src/program.cpp
```diff
@@ -959,6 +963,50 @@ std::string cvk_program::prepare_build_options(const cvk_device* device) const {
...
+    // configuration
+    if ((options.find("-denorm-preserve=") == std::string::npos) &&
+        (options.find("-denorm-flush-to-zero=") == std::string::npos)) {
+        if (!denormPreserve.empty()) {
+            options += " -denorm-preserve=";
+            for (unsigned i = 0; i < denormPreserve.size(); i++) {
+                options += denormPreserve[i];
+                options += ",";
+            }
+            options[options.size() - 1] = ' ';
```
kpet: Looks like you want a Python-like `",".join(denormPreserve)`. Maybe that's worth a utility function for clarity?

kpet (commented):
Looks fine. A couple of suggestions for clarity.

kpet (approved):
Nice!

PR #878, opened 2026-07-17
Fix lifetime issues and thread safety in resource management

rjodinchr:
- Make cvk_device refcounted and hold a reference to it in cvk_context using refcounted_holder. This ensures cvk_device is kept alive as long as cvk_context (and transitively, all api_objects associated with it) remains alive.
- Release input programs in cvk_program destructor to fix memory leaks. Changed m_input_programs to store non-const pointers to allow calling release().
- Lock end_current_command_batch and flush_no_lock calls in cvk_command_queue::enqueue_command_with_retry to prevent race conditions on queue state (m_groups and m_command_batch) when enqueuing commands.
- Release remaining event dependencies in cvk_command destructor to prevent leaks if commands are destroyed before they are executed. Clear m_event_deps in cvk_command::execute after they are released.
- Properly clean command batch when first call to add_command fails.
  It prevents having a batch (hold by a queue) keeping and handle on
  the queue while having no command. Leading to a leak that can lead
  to crashes on some platform.

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -1049,6 +1049,7 @@ cvk_context* cvk_create_context(
 
     *errcode_ret = context->init();
     if (*errcode_ret != CL_SUCCESS) {
+        context->release();
         return nullptr;
     }
 
--- a/src/context.hpp
+++ b/src/context.hpp
@@ -139,7 +139,7 @@ struct cvk_context : public _cl_context,
     void free_image_init_command_queue();
 
 private:
-    cvk_device* m_device;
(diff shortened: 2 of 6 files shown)


rjodinchr:
Divided into individual PR:
- https://github.com/kpet/clvk/pull/880
- https://github.com/kpet/clvk/pull/881
- https://github.com/kpet/clvk/pull/882
- https://github.com/kpet/clvk/pull/883
- https://github.com/kpet/clvk/pull/884
- https://github.com/kpet/clvk/pull/885

PR #879, opened 2026-07-19
Update deps and CTS

kpet:
Change-Id: I8320f42cc657a8e3c4715573c164f93b4c1821b2


kpet:
Done with https://github.com/kpet/clvk/pull/876

PR #882, opened 2026-07-22
fix cvk_command_queue to avoid memory leak

rjodinchr:
Without this fix, the following command returns some leaks:
```
$ valgrind --leak-check=full -- api_tests --gtest_filter=WithCommandQueue.VkBeginCommandBufferError

...

==785609== 2,615 (464 direct, 2,151 indirect) bytes in 1 blocks are definitely lost in loss record 2,514 of 2,522
==785609==    at 0x4B41F93: operator new(unsigned long) (vg_replace_malloc.c:487)
==785609==    by 0x503DA70: std::__detail::_MakeUniq<cvk_command_queue>::__single_object std::make_unique<cvk_command_queue, cvk_context*, cvk_device*, unsigned long&, std::vector<unsigned long, std::allocator<unsigned long> > >(cvk_context*&&, cvk_device*&&, unsigned long&, std::vector<unsigned long, std::allocator<unsigned long> >&&) (lib/gcc/x86_64-linux-gnu/15/../../../../include/c++/15/bits/unique_ptr.h:1084)
==785609==    by 0x4FFB181: cvk_create_command_queue(_cl_context*, _cl_device_id*, unsigned long, std::vector<unsigned long, std::allocator<unsigned long> >&&, int*) (src/api.cpp:1554)
==785609==    by 0x4FFB33D: clCreateCommandQueue (src/api.cpp:1580)
==785609==    by 0x41115EE: WithContext::CreateCommandQueue(_cl_device_id*, unsigned long) (tests/api/testcl.hpp:409)
==785609==    by 0x4111564: WithCommandQueue::SetUpQueue(unsigned long) (tests/api/testcl.hpp:535)
==785609==    by 0x4111529: WithCommandQueue::SetUpWithProperties(long const*, unsigned long, void*) (tests/api/testcl.hpp:543)
==785609==    by 0x410D34E: WithCommandQueue::SetUp() (tests/api/testcl.hpp:551)
==785609==    by 0x4185F73: void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (external/clspv/third_party/llvm/third-party/unittest/googletest/src/gtest.cc:2613)
==785609==    by 0x416B15F: void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (external/clspv/third_party/llvm/third-party/unittest/googletest/src/gtest.cc:2668)
==785609==    by 0x414FDD7: testing::Test::Run() (external/clspv/third_party/llvm/third-party/unittest/googletest/src/gtest.cc:2683)
==785609==    by 0x415065B: testing::TestInfo::Run() (external/clspv/third_party/llvm/third-party/unittest/googletest/src/gtest.cc:2837)
```

### DIFF
--- a/src/queue.cpp
+++ b/src/queue.cpp
@@ -187,17 +187,13 @@ cl_int cvk_command_queue::enqueue_command(cvk_command* cmd, _cl_event** event) {
     if (cmd->can_be_batched()) {
         if (!m_command_batch) {
             // Create a new command batch
-            m_command_batch = new cvk_command_batch(this);
+            m_command_batch = std::make_unique<cvk_command_batch>(this);
         }
 
         // Add command to current batch
         err = m_command_batch->add_command(
             static_cast<cvk_command_batchable*>(cmd));
         if (err != CL_SUCCESS) {
-            if (m_command_batch->batch_size() == 0) {
-                delete m_command_batch;
-                m_command_batch = nullptr;
-            }
             return err;
         }
 
@@ -295,13 +291,11 @@ cl_int cvk_command_queue::end_current_command_batch(bool from_flush) {
         if (!m_command_batch->end()) {
             return CL_OUT_OF_RESOURCES;
         }
-        enqueue_command(m_command_batch);
+        enqueue_command(m_command_batch.release());
 
         for (auto& controller : m_controllers) {
             controller->update_after_end_current_command_batch(from_flush);
         }
-
-        m_command_batch = nullptr;
 
         batch_enqueued();
     }
--- a/src/queue.cpp
+++ b/src/queue.cpp
@@ -194,6 +194,10 @@ cl_int cvk_command_queue::enqueue_command(cvk_command* cmd, _cl_event** event) {
         err = m_command_batch->add_command(
             static_cast<cvk_command_batchable*>(cmd));
         if (err != CL_SUCCESS) {
+            if (m_command_batch->batch_size() == 0) {
+                delete m_command_batch;
+                m_command_batch = nullptr;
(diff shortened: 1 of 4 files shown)


kpet:
I pushed a proposed alternative that uses `std::unique_ptr`. Let me know if that works for you.

rjodinchr:
> I pushed a proposed alternative that uses `std::unique_ptr`. Let me know if that works for you.

It does not work. But we can do something like that. We need to release it when `add_command` fails if it is empty.

The issue is about the first call to `add_command` in a command_batch failing, leaving us with an empty command batch that holds a ref on the queue. But the command batch is held by the queue, which makes a circular dependency preventing the memory from being released.

Another way to look at it would be to make sure that when the queue is flush because released, the command batch is properly deleted even if empty.

src/queue.cpp
```diff
@@ -290,14 +290,12 @@ cl_int cvk_command_queue::end_current_command_batch(bool from_flush) {
         if (!m_command_batch->end()) {
             return CL_OUT_OF_RESOURCES;
         }
-        enqueue_command(m_command_batch);
+        enqueue_command(m_command_batch.release());
```
rjodinchr: We need to release it even if `batch_size() == 0`.

rjodinchr:
Making `m_command_batch` a `unique_ptr` might not be strictly needed for that specific bug, but I think it is better overall to keep it that way.

kpet (approved):

PR #886, opened 2026-07-26
ci: update Vulkan SDK

kpet:
Change-Id: Id6c9e35ad6c577a65fdbc2d84492975518f6bff7

PR #889, opened 2026-07-27
Fix Windows build for Vulkan SDK 1.4.350 update

rjodinchr:
This PR fixes the Windows build issue by switching to an installer action that correctly extracts the Vulkan SDK.

Clearly stating: This is an AI agent (Jetski) doing this work.


rjodinchr:
@kpet I'll let you pick the change if you agree with them in https://github.com/kpet/clvk/pull/886

PR #890, opened 2026-07-28
Fix subgroup support

rjodinchr:
For shaders using subgroups, dynamically choose a subgroup size that does not exceed the device's limit on the number of subgroups per workgroup. This is done by doubling the default subgroup size until the workgroup fits within the subgroup limit. Also, clamp the reported maximum workgroup size to respect these subgroup limits.

For shaders that do not use subgroups, stop requesting a specific subgroup size. This allows the driver to choose the best size and avoids forcing a default size that could limit the maximum workgroup size.

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -3270,23 +3270,45 @@ cl_int CLVK_API_CALL clGetKernelSubGroupInfo(
         return CL_INVALID_DEVICE;
     }
 
-    switch (param_name) {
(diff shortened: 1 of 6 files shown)


rjodinchr:
Some context/testing: https://gitlab.freedesktop.org/mesa/mesa/-/work_items/15175#note_3575275

kpet (approved):

PR #892, opened 2026-08-18
Add config options to force or mask denorm support

rjodinchr:
This can be used to workaround driver bugs.

It can also be useful for testing and experimenting with applications and CTS without having to modify application code or rebuild clvk.

### DIFF
--- a/src/config.def
+++ b/src/config.def
@@ -111,6 +111,15 @@ OPTION(std::set<std::string>, device_extensions, std::set<std::string>())
 // a comma separated list of extensions.
 OPTION(std::set<std::string>, device_extensions_masked, std::set<std::string>())
 
+// Specify floating point types for which CL_FP_DENORM support should be
+// forcibly reported. Expects a comma separated list among "16", "32", "64".
+OPTION(std::set<std::string>, denorm_support, std::set<std::string>())
+
+// Specify floating point types for which CL_FP_DENORM support should be
+// masked (prevented from being reported). Expects a comma separated list among
+// "16", "32", "64".
+OPTION(std::set<std::string>, denorm_support_masked, std::set<std::string>())
+
 // Specify whether using samplers with `CL_FILTER_LINEAR` is supported.
 OPTION(bool, supports_filter_linear, true)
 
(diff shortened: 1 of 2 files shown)


rjodinchr:
@kpet , this is blocking some tests in Google infra.

Could you please have a look?

src/config.def
```diff
@@ -111,6 +111,15 @@ OPTION(std::set<std::string>, device_extensions, std::set<std::string>())
 // a comma separated list of extensions.
 OPTION(std::set<std::string>, device_extensions_masked, std::set<std::string>())
 
+// Specify floating point types for which CL_FP_DENORM support should be
+// forcibly reported. Expects a comma separated list among "16", "32", "64".
+OPTION(std::set<std::string>, denorm_support, std::set<std::string>())
```
kpet, line 116:
```suggestion
OPTION(std::set<std::string>, denorm_support_force, std::set<std::string>())
```

Maybe?

src/config.def
```diff
@@ -111,6 +111,15 @@ OPTION(std::set<std::string>, device_extensions, std::set<std::string>())
...
+
+// Specify floating point types for which CL_FP_DENORM support should be
+// masked (prevented from being reported). Expects a comma separated list among
+// "16", "32", "64".
+OPTION(std::set<std::string>, denorm_support_masked, std::set<std::string>())
```
kpet, line 121:
```suggestion
OPTION(std::set<std::string>, denorm_support_disable, std::set<std::string>())
```

Maybe?

kpet (commented):
I've made a couple of naming suggestions that IMO help clarity. Why are you using strings instead of integers for the bitwidths?

rjodinchr:
> I've made a couple of naming suggestions that IMO help clarity. 

I was using the same "convention" we have for `device_extensions`/`device_extensions_masked`.

> Why are you using strings instead of integers for the bitwidths?

Because that's a type available for the config. `std::vector<uint32_t>` would need to be implemented to be supported.
Then I felt taking the time to convert it would not bring much so I just kept it like that.

kpet:
Right, better to keep consistency for now then. Let's get this in as is as a first step.

rjodinchr:
thank you

PR #899, opened 2026-09-10
Do not require linear filtering format feature for integer images

rjodinchr:
Integer image formats (CL_SIGNED_INT*, CL_UNSIGNED_INT*) in Vulkan never support linear filtering (VK_FORMAT_FEATURE_SAMPLED_IMAGE_FILTER_LINEAR_BIT), nor does OpenCL require or permit linear filtering for integer images.

Previously, cvk_image::required_format_feature_flags_for unconditionally included VK_FORMAT_FEATURE_SAMPLED_IMAGE_FILTER_LINEAR_BIT for all read-only and read-write images when supports_filter_linear was true. This caused clGetSupportedImageFormats to erroneously filter out all mandatory integer image formats on Vulkan drivers like Lavapipe that do not support linear filtering on integer formats.

This change passes the format channel data type to required_format_feature_flags_for and only requests VK_FORMAT_FEATURE_SAMPLED_IMAGE_FILTER_LINEAR_BIT for non-integer image data types.


rjodinchr:
Needs https://github.com/google/clspv/pull/1650

rjodinchr:
@kpet , can you update clspv with the latest revision please?

